Issue #103, opened 2014-10-28
Support attribute lifecycle callback for catching unspecified changes.


treshugart:
You should be able to have a fallback handler when listening to attributes:

```js
skate('my-el', {
  attributes: {
    'my-attr': {
      // Called instead of `catchall()`.
      created: function (element, change) {}

      // Called for `updated()` and `removed()`, but not `created()`.
      catchall: function (element, change) {}
    }
  }
});
```

Issue #102, opened 2014-10-28
Support setting default attribute values.


treshugart:
There should be a way to set default attribute values if none are specified:

```js
skate('my-el', {
  attributes: {
    // value instead of object
    'my-attr': 'value'

    // object instead of value
    'my-attr': {
      default: 'value';
      
      // Or should this be used to be inline with Object.defineProperties nomenclature?
      // Using `default` is arguably more self-documenting.
      value: 'value'
    }
  }
});
```

The setting of the default value should trigger the `created()` attribute lifecycle callback, so it should be set after the lifecycle is set up.

Issue #101, opened 2014-10-28
Support linking attributes and properties.


treshugart:
Specifying an attribute:

```js
skate('my-el', {
  attributes: {
    'my-attr': { ... }
  }
});
```

Should link that attribute to a property:

```js
myEl.myAttr.should.equal(myEl.getAttribute('my-attr'));
```

PR #104, opened 2014-10-28
Attribute properties

treshugart:
Implements #101, #102 and #103.


docs/src/index.md
```diff
@@ -118,16 +118,30 @@ skate('my-component', {
...
 
+      // Called when the attribute is removed.
       removed: function (element, change) {
 
+      },
+
+      // Called when created, updated or remove is not specified. This is the
+      // same as specifying a function instead of an options object for a
+      // given attribute.
+      catchall: function (element, change) {
```
nicolasv: >  I don't like the name 'catchall', maybe something like 'always'?

The idea of this callback and it's current name are a bit confusing.

I think it'd be simpler to have an `always` callback ([like in jQuery](http://api.jquery.com/deferred.always/)) and just always call it, even if you already have a callback for that specific lifecycle event. You'd need to add an argument that specifies the type, though.

src/lifecycle.js
```diff
@@ -37,12 +38,12 @@ function ensureLifecycleFlag (target, component, name) {
...
- * @param {String} e The event to parse.
- *
- * @returns {Object]}
- */
+* Parses an event definition and returns information about it.
+*
+* @param {String} e The event to parse.
+*
+* @returns {Object]}
+*/
```
nicolasv: Wait, is this style of doc comment `*` intentional?

It looks like this:

```text
/**
*
*/
```

when every bit of code that does the `/**` thing looks like this:

```text
/**
 *
 */
```

docs/src/index.md
nicolasv: > You'd need to add an argument that specifies the type, though.

Oh, that's already provided in `change.type`.

src/lifecycle.js
```diff
@@ -51,90 +52,141 @@ function parseEvent (e) {
...
-    type = 'removed';
-  }
+/**
+* Camel-cases the specified string.
+*
+* @param {String} str The string to camel-case.
+*
+* @returns {String}
+*/
+function camelCase (str) {
```
nicolasv: Move `camelCase` to `utils.js`?

docs/src/index.md
```diff
@@ -121,16 +121,30 @@ skate('my-component', {
   // - oldValue: The old value. If type === 'created', this will be undefined.
   attributes: {
     'my-attribute': {
+      // The element's default value if not already specified on the element.
+      // Can also be a function that returns a value.
```
nicolasv: Maybe document what argument that function receives (the `target` from looking at code below)?

src/lifecycle.js
```diff
@@ -51,90 +52,141 @@ function parseEvent (e) {
...
-  // lifecycle callbacks. Skate will initialise each attribute by calling the
-  // created callback for the attributes that already exist on the element.
-  for (a = 0; a < attrsLen; a++) {
-    var attr = attrsCopy[a];
-    triggerAttributeChanged(target, component, {
-      name: attr.nodeName,
-      newValue: attr.value || attr.nodeValue
-    });
+  for (var attribute in componentAttributes) {
+    if (hasOwn(componentAttributes, attribute) && !hasOwn(target, attribute)) {
```
nicolasv: If `hasOwn(target, attribute)` is `true` does that indicate a programming error from the caller to `skate()`, i.e., they've defined an attribute that would override one on the element?

test/unit/attributes.js
```diff
@@ -4,127 +4,306 @@ import helpers from '../lib/helpers';
 import skate from '../../src/skate';
 
 describe('Attribute listeners', function () {
-  it('should listen to changes in specified attributes', function (done) {
+  'use strict';
+
+  describe('default values', function () {
+    it('should set a default value using the "default" option', function () {
+      var MyEl = skate('my-el', {
```
nicolasv: You'll need to update these tests to use `safeTagName('my-el')` to avoid re-registering the same tag name.

test/unit/attributes.js
```diff
@@ -4,127 +4,306 @@ import helpers from '../lib/helpers';
...
+            default: 'true'
+          }
+        }
+      });
+
+      skate.init(helpers.fixture('<my-el test="false"></my-el>'))
+      .firstChild
+      .test
+      .should
+      .equal('false');
```
nicolasv: No two space continuation indent? Looks like http://editorconfig.org/#file-format-details doesn't have that option.

test/unit/attributes.js
```diff
@@ -4,127 +4,306 @@ import helpers from '../lib/helpers';
...
     });
 
-    helpers.fixture('<my-el open="created"></my-el>', tagName);
+    it('should respect attributes that are already camel-cased', function () {
+      myEl.camelCased.should.equal('true');
+    });
+
+    it('should camel-case the property name and leave the attribute name as is', function () {
+      myEl.notCamelCased.should.equal('true');
+    });
```
nicolasv: Should there be an assert to see if it leaves the attribute name as is?

docs/src/index.md
nicolasv: >  I think it'd be simpler to have an always callback (like in jQuery) and just always call it.

Ah, I see that's what the function value form of an attribute property already does.

test/unit/attributes.js
```diff
@@ -4,127 +4,306 @@ import helpers from '../lib/helpers';
...
-            element.setAttribute('open', 'updated');
-          });
-        } else if (data.type === 'updated') {
-          setTimeout(function () {
-            element.removeAttribute('open');
-          });
-        } else if (data.type === 'removed') {
-          assert(true);
-          done();
+    it('should allow a catchall callback to be specified that catches all changes (same as passing a function instead of an object)', function (done) {
```
nicolasv: There should probably be a test that makes sure `catchall` is called only for the undefined lifecycle events.

src/lifecycle.js
treshugart: That was probably because Atom fucked up the formatting when pasting the doc block.

src/lifecycle.js
treshugart: Possibly. It's only used here, was just keeping it private until it needed to be public? Will think about it.

docs/src/index.md
treshugart: > In lieu of that, if the catchall option is to remain, I think default (like in a switch statement) or fallback are good alternative names.

I like `fallback` as well. That was my other option.

docs/src/index.md
treshugart: > Maybe document what argument that function receives (the target from looking at code below)?

Good idea.

src/lifecycle.js
treshugart: It just means that it doesn't need to set up the link for attributes that already have it such as `id`. If we didn't do this check, that would unnecessarily be overridden. If we threw an error, then that would mean they couldn't define any lifecycle behaviour for any existing attribute that has a corresponding property.

The original intent of this feature is to make attributes that don't have a corresponding property behave like those that do without the user needing to have any knowledge about which one's do and don't.

test/unit/attributes.js
treshugart: > No two space continuation indent? Looks like http://editorconfig.org/#file-format-details doesn't have that option.

It had one. It was messed up by Atom.

test/unit/attributes.js
treshugart: > Should there be an assert to see if it leaves the attribute name as is?

Do you think that's necessary? There aren't any code paths that change the attribute name.

nicolasv: If there are no code paths, then it's fine as it is.

PR #116, opened 2014-12-30
Wait for DOM ready before initialising pre existing elements

cyberdash:
There was inconsistent behaviour when this occurred in the <head> script tag, due to the debounce providing a timeout.

This caused problems in IE (sometimes execution was slow enough that by the time the timeout was complete, the DOM was still not ready.)

### DIFF
--- a/src/skate.js
+++ b/src/skate.js
@@ -19,9 +19,19 @@ import version from './version';
  * @returns {undefined}
  */
 var initDocument = debounce(function () {
-  document.addEventListener('DOMContentLoaded', function initialiseSkateElementsOnDomLoad() {
+  if (hasDomContentLoadedFired) {
     initElements(document.getElementsByTagName('html'));
-  });
+  } else {
+    document.addEventListener('DOMContentLoaded', function initialiseSkateElementsOnDomLoad() {
+      initElements(document.getElementsByTagName('html'));
+    });
+  }
+});
+
+var hasDomContentLoadedFired = false;
+
+document.addEventListener('DOMContentLoaded', function () {
+  hasDomContentLoadedFired = true;
 });
 
 /**
--- a/src/skate.js
+++ b/src/skate.js
@@ -14,24 +14,18 @@ import version from './version';
 
 /**
  * Initialises all valid elements in the document. Ensures that it does not
- * happen more than once in the same execution.
+ * happen more than once in the same execution, and that it happens after the DOM is ready.
  *
  * @returns {undefined}
  */
 var initDocument = debounce(function () {
-  if (hasDomContentLoadedFired) {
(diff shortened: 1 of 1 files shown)


treshugart:
It would be a good idea to use `DOMContentLoaded`. Things to keep in mind that this PR should take into account:

1. Every call to `skate()` should still debounce DOM initialisation so consecutive calls should only trigger a single DOM traversal after the last call is made.
2. When using `DOMContentLoaded`, it should traverse the DOM once. Calling `skate()` before it is fired should not traverse the DOM because when it is fired, components will be initialised. Calling `skate()` after should traverse the DOM because it would have already been fired and would not trigger again.

For number 2, that probably means there will need to be some sort of flag that it can check before it attempts to traverse the DOM.

src/skate.js
```diff
@@ -19,7 +19,19 @@ import version from './version';
  * @returns {undefined}
  */
 var initDocument = debounce(function () {
-  initElements(document.getElementsByTagName('html'));
+  if (hasDomContentLoadedFired) {
+    initElements(document.getElementsByTagName('html'));
+  } else {
+    document.addEventListener('DOMContentLoaded', function initialiseSkateElementsOnDomLoad() {
+      initElements(document.getElementsByTagName('html'));
```
treshugart: Can this go in the listener below which sets the flag?

src/skate.js
```diff
@@ -19,7 +19,19 @@ import version from './version';
...
+    });
+  }
+});
+
+var hasDomContentLoadedFired = false;
+
+document.addEventListener('DOMContentLoaded', function () {
+  hasDomContentLoadedFired = true;
```
treshugart: Can probably just `initElements...` after the flag is set here, yeah?

src/skate.js
```diff
@@ -19,7 +19,19 @@ import version from './version';
  * @returns {undefined}
  */
 var initDocument = debounce(function () {
-  initElements(document.getElementsByTagName('html'));
+  if (hasDomContentLoadedFired) {
+    initElements(document.getElementsByTagName('html'));
```
treshugart: Should probably consider using `document.documentElement` instead of using `getElementsByTagName`. I know this existed before, but if you could make that change as well it'd be nice.

src/skate.js
cyberdash: I think that would mean elements are initialised regardless of whether `initDocument` is called though; so it'd always initialise the document on DOM ready. That would be unnecessary if no skate skate handlers are called.

It'd make the code simpler though, so maybe it's not worth it. I'd be happy to change it if you think the performance in that case isn't an issue.

src/skate.js
cyberdash: Sounds good!

PR #123, opened 2015-01-16
Don't share the same object literal for prototypes when unspecified (#122)

### LINKED ISSUE #122: Default prototype property ({}) is reusued across different elements
The value of `skate.defaults.prototype` is a object literal (`{}`) which means that if you don't provide a `prototype` property as part of your *`definition`* argument to `skate()`, you'll end up sharing the same prototype objects with other instances that also don't define a `prototype`.

This causes an error like this in Chrome:

```text
Error: Failed to execute 'registerElement' on 'Document': Registration failed for type 'my-element'. The prototype is already in-use as an interface prototype object.
```

nicolasv:
Fixes #122.

### DIFF
--- a/src/skate.js
+++ b/src/skate.js
@@ -83,6 +83,9 @@ function makeElementConstructor (definition) {
 function skate (id, definition) {
   definition = inherit(definition || {}, skate.defaults);
   definition.id = id;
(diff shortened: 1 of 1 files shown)


nicolasv:
Whoops, I labeled the branch and commit message with #121 when I meant #122.

PR #124, opened 2015-01-16
Unique tag names in tests to avoid re-registration errors (#120)

### LINKED ISSUE #120: Some tests fail in Chrome with "a type with that name is already registered"
This is due to the fact that the tests run in a single browser session and thus attempt to re-register element tag names, causing the error:

```text
Error: Failed to execute 'registerElement' on 'Document': Registration failed for type 'my-element'. A type with that name is already registered.
```

nicolasv:
Fixes #120.

### DIFF
--- a/test/lib/helpers.js
+++ b/test/lib/helpers.js
@@ -1,8 +1,29 @@
 'use strict';
 
+var tagNameCounter = 1;
+
 export default {
-  add: function (name) {
-    return this.fixture('<' + name + '></' + name + '>').querySelector(name);
+  /**
+   * Generate a unique tag name to avoid errors due to re-registering a tag
+   * name.
+   *
+   * Returns an object that can be looked up using the provided id to obtain
+   * a unique tag name to use with skate() and fixture().
+   *
+   * @param id the identifier for the tag
+   * @returns {Object.<string, string>} an object with the key 'id' mapped to
+   *   the value of the id and the value of the id mapped to a unique tag name
+   */
+  uniqueTagName: function (id) {
+    var safeId = id + '-' + (tagNameCounter++).toString();
+    var tagName = {id: id};
+    tagName[id] = safeId;
+    return tagName;
+  },
+
+  add: function (name, tagName) {
+    var safeTagName = typeof tagName === 'undefined' ? name : tagName[name];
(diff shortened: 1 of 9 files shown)


test/unit/registration.js
```diff
@@ -112,9 +114,11 @@ describe('Returning a constructor', function () {
   describe('when an extends option is specified', function () {
     var Div;
     var div;
+    var tagName;
 
     beforeEach(function () {
-      Div = skate('my-element', {
+      tagName = helpers.uniqueTagName('my-element');
```
nicolasv: It would be nice to use destructuring assignment here, i.e.,

```js
{'my-element': tagName} = helpers.uniqueTagName('my-element');
```
but it seems for destructuring assigment doesn't work for existing `var`s (unless I'm mistaken).

test/lib/helpers.js
```diff
@@ -1,8 +1,29 @@
...
+   *
+   * @param id the identifier for the tag
+   * @returns {Object.<string, string>} an object with the key 'id' mapped to
+   *   the value of the id and the value of the id mapped to a unique tag name
+   */
+  uniqueTagName: function (id) {
+    var safeId = id + '-' + (tagNameCounter++).toString();
+    var tagName = {id: id};
+    tagName[id] = safeId;
+    return tagName;
```
nicolasv: Maybe returning an object like

```js
return {unsafe: id, safe: safeId};
```
provides a better API. Usages would end up looking like:

```js
var tagName = helpers.uniqueTagName('my-element');
skate(tagName.safe, /* ... */)
var {safe: tagName, unsafe: unsafeTagName} = helpers.uniqueTagName('my-element');
```

test/lib/helpers.js
```diff
@@ -10,9 +31,18 @@ export default {
     return element;
   },
 
-  fixture: function (html) {
-    var fixture = document.getElementById('fixture');
+  fixture: function (html, tagName) {
+    if (typeof html === 'string' && typeof tagName !== 'undefined') {
```
nicolasv: This entire block should be moved into the `if (typeof html === 'string')` block below.

nicolasv: Done in de06358.

test/lib/helpers.js
nicolasv: Done in 0f93a09.

PR #131, opened 2015-01-20
Native registerElement prototype must inherit from HTMLElement

nicolasv:
See the commit message for 192a033 more details.

### DIFF
--- a/src/skate.js
+++ b/src/skate.js
@@ -83,9 +83,6 @@ function makeElementConstructor (definition) {
 function skate (id, definition) {
   definition = inherit(definition || {}, skate.defaults);
   definition.id = id;
-  if (!HTMLElement.prototype.isPrototypeOf(definition.prototype)) {
-    definition.prototype = inherit(Object.create(HTMLElement.prototype), definition.prototype, true);
-  }
 
   registry.set(id, definition);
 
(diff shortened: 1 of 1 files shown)


nicolasv:
Fixes this failing test under Chrome on master:

```text
Chrome 39.0.2171 (Mac OS X 10.10.1) Lifecycle scenarios use the constructor then add it to the DOM "before each" hook FAILED
	Error: Uncaught TypeError: undefined is not a function (/Users/nvenegas/Code/skatejs/.tmp/run-unit-tests.js:372)
	    at Object.module.exports.fixture (/Users/nvenegas/Code/skatejs/.tmp/run-unit-tests.js:1312:17)
	    at Context.<anonymous> (/Users/nvenegas/Code/skatejs/.tmp/run-unit-tests.js:2285:15)
```

src/skate.js
```diff
@@ -84,7 +84,7 @@ function skate (id, definition) {
   definition = inherit(definition || {}, skate.defaults);
   definition.id = id;
   if (definition.prototype === skate.defaults.prototype) {
-    definition.prototype = {};
+    definition.prototype = Object.create(HTMLElement.prototype);
```
nicolasv: Rather than only setting this if the caller didn't provide a `prototype` in the `definition`, perhaps we should insert `HTMLElement.prototype` into `definition.prototype`'s prototype chain.

Doing so may make less existing code break.

nicolasv: Or, perhaps more straightforward than monkeying around with prototype chains: if `definition.prototype` is provided and it is not an instance of `HTMLElement`, we copy over its own properties onto the result of `Object.create(HTMLElement.prototype)`.

treshugart: I like the latter solution. Simple and provides a lot of convenience and prevents potential gotchas with how the spec and native implementations handle things. It's unexpected that a custom element wouldn't implicitly inherit from HTMLElement. It is, after all, an element. Even `HTMLUnknownElement` derives from `HTMLElement`.

Issue #133, opened 2015-01-22
Is it possible to prerender everything on serverside?


Align-Deployment:
Subj. 
Thanks.

treshugart:
Actually, yes. It depends on how you're templating your custom elements,
though. On its own, all Skate does is bind behavior. Whether or not you
render your template client or server side doesn't really matter.

If you want to be able to do both, you can set a flag attribute in the
server markup and check for it in your template code. If the attribute
exists on the element, don't render client side, otherwise go for it. I've
been thinking about building that into Skate as a feature, but it'd only
save you a couple lines of code anyways.

I hope that helps.

treshugart:
Just FYI, a recent commit will make it so you don't have to worry about checking manually.

https://github.com/skatejs/skatejs/commit/0d5ed709b9da54958690ac2e16414707a9aa2a06

I just came across an issue where jQuery was cloning and reinserting the element after it was templated causing it to happen twice. This only seemed to happen if browsers supported custom elements natively and when skate was using those means to register the component.

treshugart:
Closing this now. Please re-open if you still have any questions.

Cheers.

Align-Deployment:
Thanks a bunch

treshugart:
@Align-Deployment, just FYI, this technique for server-side rendering isn't public API yet. I don't want to solidify anything until we feel it's ok to start using as public API. I'm putting it on the road map, but until then use with caution.

treshugart:
https://github.com/skatejs/skatejs/issues/136

PR #135, opened 2015-01-27
Ensure attach lifecycle callback is not triggered if it is not in the DOM.

treshugart:


src/lifecycle.js
```diff
@@ -269,21 +253,22 @@ function addEventListeners (target, component) {
...
+  targetData.set('created', true);
+
   // TODO: This doesn't need to happen if using native.
   inherit(target, component.prototype, true);
 
   // We use the unresolved / resolved attributes to flag whether or not the
-  // element has been templated or not. The resolved attribute is also how
-  // you can ensure there isn't a FOUC. If no templating is happening, then
-  // there is no reason to add / remove these attributes.
-  if (component.template && !target.hasAttribute(component.resolvedAttribute)) {
```
nicolasv: Why'd you move away from using `resolvedAttribute`? Does this mean that #133 (prerendering things server-side) should be reopened?

test/unit/init.js
```diff
@@ -153,16 +153,12 @@ describe('Instantiation', function () {
...
+          idsToCheck.push(id);
         }
       });
     });
 
     skate.init(div);
-
-    idsToSkate.forEach(function (id) {
-      idsToCheck[id].should.equal(true);
-    });
```
nicolasv: What assertion is made in this test now (if any)?

test/unit/lifecycle.js
```diff
@@ -4,38 +4,66 @@ import helpers from '../lib/helpers';
 import skate from '../../src/skate';
 
 describe('Lifecycle Callbacks', function () {
-  it('should call the created() callback when the element is attached', function (done) {
-    var tagName = helpers.safeTagName('my-element');
-    skate(tagName.safe, {
+  var MyEl;
+  var myEl;
```
nicolasv: Hmm, I'm generally not a fan of var names that are this similar. Would you consider `myEl` and `myElConstructor` (or `myElCtor`)?

treshugart: Sure.

src/lifecycle.js
treshugart: I'm working on a separate commit to re-address this.

test/unit/init.js
treshugart: I should probably actually remove this test. It's hard to grok and it's not testing anything useful. The original problem was specific to firefox and existed when you had a component called `watch`. This isn't a problem if using the recommended spec'd naming convention which we're recommending moving forward.

PR #158, opened 2015-03-15
Fix ie test timeouts

nicolasv:


src/mutation-observer.js
```diff
@@ -100,13 +100,13 @@ MutationObserver.fixIe = function () {
   }
 
   // We have to call the old innerHTML getter and setter.
-  var oldInnerHtml = Object.getOwnPropertyDescriptor(elementPrototype, 'innerHTML');
```
treshugart: Nitpick but I do see your reasoning. All uppercase is a shitty convention that should be done away with. You can't tell where one word ends and another begins as quickly: SomeHTMLString vs SomeHtmlString. No need to change, but for future reference that's a convention we follow.

treshugart: It also seems that there isn't a standard naming convention: document.getElementById.

nicolasv: Yeah, I'm breaking the style guide in this instance because I think consistency with `innerHTML` is more important: I think it makes it easier to make the mental connection of what the variable is.

Issue #140, opened 2015-01-30
Remove registry.clear().


treshugart:
Being able to unregister element's doesn't follow the custom element spec and it's only for testing anyways; it's not public API. Best just to remove it because we should be writing our tests as close to the spec as possible especially since we now support native.

PR #160, opened 2015-03-18
Remove registry.clear() method. Update tests to use the safe-tag-name he...

treshugart:
...lper.

### DIFF
--- a/test/unit/registration.js
+++ b/test/unit/registration.js
@@ -6,44 +6,46 @@ import skate from '../../src/skate';
 describe('Registration', function () {
   it('should not allow you to register the same component more than once.', function () {
     var multiple = false;
-
-    skate('div', {});
+    var tag = helpers.safeTagName('my-el');
+    var Element = skate(tag.safe, {});
 
     try {
-      skate('div', {});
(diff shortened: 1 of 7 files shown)


treshugart:
Fixes #140.

test/unit/registration.js
```diff
@@ -51,34 +53,24 @@ describe('Returning a constructor', function () {
...
 
-    Div.prototype.func2 = function () {};
-
-    var div = new Div();
-
-    expect(div.func1).to.be.a('function');
-    expect(div.func2).to.be.a('function');
-  });
-
-  it('should allow the overwriting of the prototype', function () {
```
nicolasv: What is the new behaviour of attempting what this test did? Perhaps it's worth keeping the test around to ensure the new behaviour remains.

nicolasv:
:+1:

test/unit/registration.js
treshugart: We could write a test to make sure the prototype cannot be overwritten, but I'm not sure if we need to go that far. It's just not a feature anymore.

Issue #141, opened 2015-02-02
Initialise each component's lifecycle callbacks separately.


treshugart:
Currently, when initialising an element using Mutation Observers,`initElements` will call `triggerLifecycle` which causes the `created` and `attached` callbacks for a given component to be executed before another component's. It would better mimic the lifecycle of native if we ensure all `created` callbacks are executed before all `attached` callbacks.

PR #161, opened 2015-03-19
Call created and attached callbacks in their own loops when initialising...

treshugart:
... components to be more inline with how native custom elements behave.

Resolves: #141

### DIFF
--- a/build/task/test.js
+++ b/build/task/test.js
@@ -16,7 +16,12 @@ galvatron.transformer
   .post('babel')
   .post('globalize');
 
+var browsers = commander.browsers || 'Firefox';
 var clientArgs = [];
+
+if (browsers === 'all') {
+  browsers = 'Chrome,Firefox,Opera,Safari';
+}
 
 if (commander.grep) {
   clientArgs.push('--grep');
@@ -30,7 +35,7 @@ function run () {
     hostname: commander.host || '0.0.0.0',
     port: commander.port || 9876,
     frameworks: ['mocha', 'sinon-chai'],
-    browsers: (commander.browsers || 'Firefox').split(','),
+    browsers: browsers.split(','),
     client: { args: clientArgs },
     files: [ '.tmp/unit.js' ]
   });
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -365,32 +365,21 @@ function createElementTreeWalker (element) {
 }
 
 /**
- * Initialises a set of elements.
+ * Returns a flattened array of elements and their component definitions from
+ * the passed array of elements.
  *
- * @param {DOMNodeList | Array} elements A traversable set of elements.
+ * @param {Array} elements The array of elements to flatten.
  *
- * @returns {undefined}
+ * @returns {Array}
  */
-function initElements (elements) {
-  var definitions;
-  var definitionsLength;
-  var element;
-  var elementAttrs;
-  var elementsLength;
-  var elementsList;
-  var elementsListLength;
-  var elementsListDefinitions;
-  var elementWalker;
-  var elementWalkerCurrentNode;
-
-  elementsList = [];
-  elementsListDefinitions = [];
-  elementsLength = elements.length;
+function flattenElements (elements) {
+  var flattened = [];
+  var elementsLength = elements.length;
 
   // Build a list of nodes that we will initialise.
   for (var a = 0; a < elementsLength; a++) {
-    element = elements[a];
-    elementAttrs = element.attributes;
+    var element = elements[a];
+    var elementAttrs = element.attributes;
 
     // We screen the root node only. The rest of the nodes are screened in the
     // tree walker.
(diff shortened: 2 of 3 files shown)


treshugart:
All of the `for` loops are micro-optimisations for IE just like the rest of the code. They do add up and make a difference in large DOMs.

I have a couple of worries about this PR, mostly having to do with performance in IE:
- Using a `TreeWalker` may in fact degrade performance even though it's better on memory than using a recursive function.
- Iterating over the elements separately in order to separate the `created` and `attached` callbacks may also degrade performance in IE.

src/lifecycle.js
```diff
@@ -346,16 +343,36 @@ function triggerDetached (target, component) {
...
- * @param {Object} component The component data.
+ * @param {DOMElement} element The element to walk.
  *
- * @returns {undefined}
+ * @returns {TreeWalker}
  */
-function triggerLifecycle (target, component) {
-  triggerCreated(target, component);
-  triggerAttached(target, component);
+function createTreeWalker (element) {
```
nicolasv: `createTreeWalker` → `createElementTreeWalker` maybe? Since you're using just `NodeFilter.SHOW_ELEMENT`.

build/task/test.js
```diff
@@ -16,8 +16,13 @@ galvatron.transformer
   .post('babel')
   .post('globalize');
 
+var browsers = commander.browsers || 'Firefox';
```
nicolasv: If line 9's use of `[Chrome,Firefox]` in 
```text
-b, --browsers [Chrome,Firefox]', 'The browsers to run the tests in.
```
is meant to signify the defaults, then these two lines are out of sync.

src/lifecycle.js
```diff
@@ -366,27 +383,67 @@ function triggerLifecycle (target, component) {
  * @returns {undefined}
  */
 function initElements (elements) {
-  var elementsLen = elements.length;
-
-  for (var a = 0; a < elementsLen; a++) {
-    var element = elements[a];
-
+  var a, b, c, d, e;
```
nicolasv: I'd probably just reuse `i` and `j` across the different for loops.

I think declaring all these as the first line is a little jarring if you haven't seen the rest of the function before. So I'd also say declare & init them in the `for` statements.

build/task/test.js
treshugart: No, it's just an example of how you can pass one or more browsers.

src/lifecycle.js
treshugart: Ended up declaring in the loops. That's how I normally do it anyways. I've been declaring vars at the top most of the time and was just seeing how it panned out here. The end goal is to refactor this to be less complex.

src/lifecycle.js
```diff
@@ -366,27 +372,67 @@ function triggerLifecycle (target, component) {
...
+
+  // Build a list of nodes that we will initialise.
+  for (a = 0; a < elementsLength; a++) {
+    element = elements[a];
+
+    // We screen the root node only. The rest of the nodes are screened in the
+    // tree walker.
     if (element.nodeType !== 1 || element.attributes[ATTR_IGNORE]) {
       continue;
     }
```
nicolasv: Do you need to screen the root node here, or will the tree walker do that for you?

src/lifecycle.js
```diff
@@ -366,27 +372,67 @@ function triggerLifecycle (target, component) {
...
 
-    var currentNodeDefinitions = registry.getForElement(element);
-    var currentNodeDefinitionsLength = currentNodeDefinitions.length;
+    elementWalker = createTreeWalker(element);
+    elementsList.push(element);
+    elementsListDefinitions.push(registry.getForElement(element));
 
-    for (var b = 0; b < currentNodeDefinitionsLength; b++) {
-      triggerLifecycle(element, currentNodeDefinitions[b]);
+    while (elementWalker.nextNode()) {
```
nicolasv: I would probably push this looping over the tree walker into the the `createTreeWalker()` function (or introduce a layer) — you give it the element and it returns a list of elements — would make this function a tad simpler.

src/lifecycle.js
treshugart: I think line 374 does what you're talking about, right? I should change that slightly actually to guard against no attributes.

test/unit/ignoring.js
```diff
@@ -1,32 +1,90 @@
...
+  it('should ignore a flagged element if defined after it is inserted', function () {
     var called = 0;
+    var tag = helpers.safeTagName();
 
-    // Test attaching before.
-    document.body.innerHTML = '<div></div><div id="container-1" data-skate-ignore><div><div></div></div></div><div></div>';
-    document.getElementById('container-1').innerHTML = '<div><div></div></div>';
+    helpers.fixture(`<${tag.safe} data-skate-ignore></${tag.safe}>`);
+    skate(tag.safe, {
+      created: function () {
```
nicolasv: Is it worth also defining `attached` handlers and testing that their call counts are 0, too?

nicolasv:
Nice update to `initElements()`!

test/unit/ignoring.js
treshugart: Yeah, good idea.

Issue #162, opened 2015-03-24
Ensure Skate components can be extended via ES6 classes.


treshugart:
References #159.

alexreardon:
Can this issue be closed?

treshugart:
Closing now. There was one last bug that I noticed and just pushed a fix for. I'll reference in a separate issue.

PR #167, opened 2015-04-15
Ensure Skate components can be extended using ES6 classes.

treshugart:
References #162.


treshugart:
This includes better separation between modules and a lot of cleanup work.

PR #176, opened 2015-04-22
#175 backport noConflict() to 0.12.x

zephor:


src/skate.js
```diff
@@ -180,7 +180,11 @@ skate.defaults = {
 // Always export the global. We don't know how consumers are using it and what
 // their environments are like. Doing this affords them the flexibility of
 // using it in an environment where module and non-module code may co-exist.
-window.skate = skate;
+var previousSkate = window.skate;
+export default function () {
```
treshugart: Doing `export default` here will export the `noConflict()` function instead of the `skate()` function to ES6. This should probably read `skate.noConflict = function () {...` instead.

PR #189, opened 2015-05-26
Move triggering of attribute created callbacks to after the created lifecycle callback is triggered.

### LINKED ISSUE #179: suspend execution of attribute.created?
For all of the attributes I implement when using skate, I have to do checks like:

```
...
attributes: { foo: { created(el, diff) {
    if(!("shadowRoot" in el)) return;
    ...
}, }, },

created(el) {
    ... // (initialise shadow root here, etc.)
    if(el.foo) el.setFoo(el.foo);
}
```

This is because my attributes interact with the element's shadow DOM, and so `created` can't function properly if called before the element has finished being created.

Perhaps it would be better to change skate's behaviour so that `attribute.*` doesn't get run before the element's `created` function has finished, as I think there are few use cases for using these callbacks in the absence of the DOM. Is that feasible?

treshugart:
Fixes #179.

Also fixes tests in browsers that support native custom elements by checking if we should apply the prototype at runtime in a different way.

### DIFF
--- a/src/lifecycle/created.js
+++ b/src/lifecycle/created.js
@@ -185,10 +185,11 @@ export default function (options) {
     patchAttributeMethods(element, options);
     addAttributeToPropertyLinks(element, options);
     initAttributes(element, options);
-    triggerAttributesCreated(element, options);
 
     if (options.created) {
       options.created(element);
     }
+
+    triggerAttributesCreated(element, options);
   };
 }
--- a/src/lifecycle/created.js
+++ b/src/lifecycle/created.js
@@ -162,10 +162,10 @@ export default function (options) {
 
     // Native custom elements automatically inherit the prototype. We apply
     // the user defined prototype directly to the element instance if not.
-    // Note that in order to catch modified prototype chains - such as when
-    // setPrototypeOf() or ES6 classes are used - we must walk each prototype
-    // and apply each member directly.
-    if (!options.isNative) {
+    // Skate will always add lifecycle callbacks to the definition. If native
+    // custom elements are being used, one of these will already be on the
+    // element. If not, then we are initialising via non-native means.
+    if (!element.attributeChangedCallback) {
       getPrototypes(options.prototype).forEach(function (proto) {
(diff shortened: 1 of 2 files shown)


src/lifecycle/created.js
```diff
@@ -162,10 +162,10 @@ export default function (options) {
 
     // Native custom elements automatically inherit the prototype. We apply
     // the user defined prototype directly to the element instance if not.
-    // Note that in order to catch modified prototype chains - such as when
-    // setPrototypeOf() or ES6 classes are used - we must walk each prototype
-    // and apply each member directly.
-    if (!options.isNative) {
+    // Skate will always add lifecycle callbacks to the definition. If native
```
treshugart: This bit was causing tests to fail in browsers that support native custom elements. The use case was:

```js
var el = document.createElement('div');
el.setAttribute('is', 'custom-element');
skate.init(el);
```

This isn't natively supported by web components, but is functionality that used to work, and be supported by, Skate. This actually doesn't add any overhead, we're just checking to see if the element has been natively set up in a different, more predicable, manner.

src/lifecycle/created.js
```diff
@@ -185,10 +185,11 @@ export default function (options) {
     patchAttributeMethods(element, options);
     addAttributeToPropertyLinks(element, options);
     initAttributes(element, options);
-    triggerAttributesCreated(element, options);
 
     if (options.created) {
       options.created(element);
     }
+
+    triggerAttributesCreated(element, options);
```
treshugart: This is actually the only change *required* by the PR.

test/unit/init.js
```diff
@@ -189,7 +189,8 @@ describe('skate.init()', function () {
       it(`type "${type}", extending "${tagToExtend}"`, function () {
         var calls = 0;
         var callsPerCreationType = {};
-        var {safe: tagName} = helpers.safeTagName('my-element');
+        var { safe: tagName } = helpers.safeTagName();
```
treshugart: Conventional whitespace.

Issue #216, opened 2015-06-19
Completely remove Shadow DOM support in events.


treshugart:
References https://github.com/skatejs/skatejs/issues/203#issuecomment-113186520.

PR #217, opened 2015-06-19
Remove Shadow DOM support from event delegation.

treshugart:
See #216.

Shadow selectors are being removed from the spec therefore we're removing it from Skate. We can still add this at a later point if it ever gets re-introduced, even after a 1.0.

Also moved the event tests to the public API testing area and fixed them in IE.


nicolasv:
LGTM other than I think you need to delete `test/unit/events.js` since its contents are now in `test/unit/api/event.js`.

PR #218, opened 2015-06-19
Fix extending tests since IE <= 10 can't do some class things via Babel

nicolasv:


test/unit/extending.js
```diff
@@ -46,17 +54,22 @@ describe('extending', function () {
...
 
   it('should allow overriding of callbacks', function () {
-    var ExtendedCtor = skate(tag, class extends Ctor {
-      static created () {
-        super.created();
-        this.textContent += 'ing';
-      }
-    });
-    expect(new ExtendedCtor().textContent).to.equal('testing');
+    if (canResolveSuper) {
```
treshugart: I've been experimenting with putting checks like this before the call to `it()` or `describe()`. For example:

```js
canResolveSuper && it('should allow overriding of callbacks', ...);
```

I just find I like the way it reads with the descriptions of the tests and reduces the need for indentation in the test itself. No need to change, just wonder what you think.

nicolasv: I like it.

PR #220, opened 2015-06-19
Fix readyState handling in IE (since it was firing too early)

### LINKED ISSUE #219: readyState handler can fire too early in Internet Explorer
Same deal as jquery/jquery#901 (“sometimes IE10 and lower have document.readyState == "interactive" way too early”).

nicolasv:
Fixes #219.


nicolasv:
We'll need to backport this to 0.12.0 and 0.13.4 for AUI.

PR #226, opened 2015-06-23
Modularise binding types

treshugart:
The idea behind this is that it gives us the ability to pull out unnecessary binding types at a later point. I think it's really worth considering only having the w3c compat spec as the only binding type by default but have the ability to add in custom ones. This way we could remove legacy support for older components as we see fit without affecting backward compatibility in Skate after 1.0.

Some things addressed as part of this:

- Added perf tests for different number of mutation observers on the page and their impact on performance.
- Removed constants file as it's not necessary.

### DIFF
--- a/src/api/type.js
+++ b/src/api/type.js
@@ -1,11 +1,98 @@
-import {
-  TYPE_ATTRIBUTE,
-  TYPE_CLASSNAME,
-  TYPE_ELEMENT
-} from '../constants';
+function getClassList (element) {
+  var classList = element.classList;
+
+  if (classList) {
+    return classList;
+  }
+
+  var attrs = element.attributes;
+
+  return (attrs['class'] && attrs['class'].nodeValue.split(/\s+/)) || [];
+}
 
 export default {
-  ATTRIBUTE: TYPE_ATTRIBUTE,
-  CLASSNAME: TYPE_CLASSNAME,
-  ELEMENT: TYPE_ELEMENT
+  ATTRIBUTE: {
+    create (opts) {
+      var elem = document.createElement(opts.extends || 'div');
+      elem.setAttribute(opts.id, '');
+      return elem;
+    },
+    find (elem, defs) {
+      var attrs = elem.attributes;
+      var attrsLen = attrs.length;
+      var definitions = [];
+      var tag = elem.tagName.toLowerCase();
+
+      for (let a = 0; a < attrsLen; a++) {
+        let attr = attrs[a].nodeName;
+        let definition = defs[attr];
+
+        if (definition && definition.type === this) {
+          let tagToExtend = definition.extends;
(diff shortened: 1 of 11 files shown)


src/api/type.js
```diff
@@ -1,11 +1,71 @@
...
+function getClassList (element) {
+  var classList = element.classList;
+
+  if (classList) {
+    return classList;
+  }
+
+  var attrs = element.attributes;
+
+  return (attrs['class'] && attrs['class'].nodeValue.split(/\s+/)) || [];
```
nicolasv: Just use element.className rather than checking for the class attribute?

treshugart: Yep. That's residual from older code.

src/defaults.js
```diff
@@ -1,4 +1,4 @@
-import { TYPE_ELEMENT } from './constants';
+import binding from './polyfill/binding';
```
nicolasv: From the naming of this file and its export, it's not obvious to me that it is only for elements — so either I'm not familiar with the code/w3c-spec enough (which is true) or it could use a more telling name.

treshugart: I was thinking about that this morning actually. Probably better in something like `src/binding/w3c`. #230.

src/polyfill/binding.js
```diff
@@ -0,0 +1,30 @@
...
+    if (definition && definition.type === this) {
+      let tagToExtend = definition.extends;
+
+      if (isAttrValue) {
+        if (tag === tagToExtend) {
+          definitions.push(definition);
+        }
+      } else if (!tagToExtend) {
+        definitions.push(definition);
+      }
```
nicolasv: This layout would be a bit clearer to me:
```js
if (tagToExtend && isAttrValue) {
  …
} else {
  …
}
```

src/polyfill/binding.js
```diff
@@ -0,0 +1,30 @@
+export default {
+  /* jshint expr: true */
+  create (opts) {
+    var elem = document.createElement(opts.extends || opts.id);
+    opts.extends && elem.setAttribute('is', opts.id);
+    return elem;
```
nicolasv: Say we had

```js
skate('x-foo', …);
skate('bar', { extends: 'x-foo', … });
```

does this creation order (1. `createElement`, 2. add the `is` attr) risk running the `x-foo` `created` callback when if you had done `….innerHTML = '<x-foo is="bar"></x-foo>'` you wouldn't have?

treshugart: Adding the `is` attribute post creation doesn't do anything in native land. This factory never gets run in native, though, it runs the native constructor instead.

src/polyfill/binding.js
treshugart: Yup, that'd work. Feel free to refactor some code if it needs it.

Issue #118, opened 2015-01-12
If a document fragment is initialised, getClosestIgnoredElement fails.


treshugart:
If you try and initialise an element inside a document fragment, when it checks to see if the element is ignored, it calls getClosestIgnoredElement. This goes up the DOM tree until it hits the document fragment and since hasAttribute doesn't exist on the fragment, it fails.

zephor:
related to: https://ecosystem.atlassian.net/browse/AUI-3239

treshugart:
This bug only exists when you add an element to the DOM that has a mutation observer registered.

1. Add a element to the DOM.
2. Before the mutation observer can fire (so basically in the same synchronous block) remove the element and insert it into a document fragment.

This generally isn't a use case you will see, however, jQuery uses this technique internally and causes problems. 

The solution is to check if the `parent` has the `hasAttribute()` method before calling it inside the `getClosestIgnoredElement()` function.

PR #232, opened 2015-06-26
Backport document fragment fix from #118 to 0.9.x.

nicolasv:


nicolasv:
The other test that 548ecee adds (`should not call attached when initialised inside a document fragment`) doesn't pass in 0.9.x so I've skipped it 'cause I assume it was fixed before 118 was fixed.

PR #233, opened 2015-06-26
Backport document fragment fix & tests from #118 to 0.12.x

nicolasv:


test/unit/dom.js
```diff
@@ -95,4 +95,46 @@ describe('DOM', function () {
...
+  describe('Document Fragments', function () {
+    var frag;
+    var created;
+    var attached;
+    var MyEl;
+
+    beforeEach(function () {
+      created = false;
+      attached = false;
+      MyEl = skate('my-element', {
```
nicolasv: `0.12.x` doesn't have the safe tag name stuff, so just reusing `'my-element'` — I guess the tests here don't have a problem with registering the same skate `id` more than once?

test/unit/dom.js
```diff
@@ -95,4 +95,46 @@ describe('DOM', function () {
...
+    });
+
+    it('GH-118 - should not trigger the element is inserted, removed and then put into a fragment', function (done) {
+      var div = document.createElement('div');
+
+      document.body.appendChild(div);
+      frag.appendChild(div);
+
+      helpers.afterMutations(function () {
+        done();
```
nicolasv: Oh, I want to add the comment from #232 here.

test/unit/dom.js
treshugart: No they don't. That was prior to native support where we were able to unregister components.

PR #244, opened 2015-07-05
check for interactive document.readyState in skate.ready (fixes #243)

### LINKED ISSUE #243: skate.ready: callback not executed if document.readyState === "interactive"
This is because `DOMContentLoaded` is fired when `document.readyState` becomes `"interactive"` (finished parsing), while the `load` event is fired when it becomes `"complete"` (finished loading sub-resources).

If, then, the current `document.readyState` is `"interactive"` and `skate.ready` is called, the `DOMContentLoaded` event has already been fired and so the callback will be waiting for an event that will never be fired again.

marcoms:


treshugart:
@nicolasv What are your thoughts here? Might be good to merge to get other browsers working again and then apply a fix specific to the IE versions that are affected by the bug we experienced?

marcoms:
superseded by #249

PR #248, opened 2015-07-10
Add skate.fragment() method that returns a fragment that inits elements

nicolasv:
This method is designed to allow us to remove `skate.init()`: where previously you would do something like

```js
customEl.innerHTML = html;
skate.init(customEl);
```
you can now do

```js
customEl.appendChild(skate.fragment(html));
```

It differs from `skate.create()` in that it allows you to pass in HTML that would result in multiple sibling elements; for that same HTML string, `skate.create()` creates a single element corresponding to the first element in the string.


src/api/fragment.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function createInitingFragment() {
+  var frag = document.createDocumentFragment();
+  decorateFragmentMethods(frag);
+  return frag;
+}
+
+export default function (html) {
+  var frag = createInitingFragment();
```
nicolasv: Might be worth just inlining `createInitingFragment()`.

nicolasv:
We could potentially merge `skate.create()` and `skate.fragment()` such that `create()` always returns a fragment, but `create()`'s `props` argument is problematic (it wouldn't be clear what element(s) should have the `props` added to them).

nicolasv:
Looks like there's some work to do to fix the tests in IE:

```text
IE 9.0.0 (Windows 7) api/fragment should return an empty fragment if no argument specified FAILED
	expected undefined to equal 0
IE 9.0.0 (Windows 7) api/fragment overriden methods "before each" hook FAILED
	Object doesn't support this action
IE 9.0.0 (Windows 7) api/fragment html should not init an element without a definition FAILED
	Unable to get value of the property 'hasAttribute': object is null or undefined
IE 9.0.0 (Windows 7) api/fragment html should init a single element FAILED
	Object doesn't support this action
IE 9.0.0 (Windows 7) api/fragment html should init multiple elements FAILED
	Object doesn't support this action
IE 9.0.0 (Windows 7) api/fragment html should init an element and its descendents FAILED
	Object doesn't support this action
IE 9.0.0 (Windows 7) api/fragment html should work with special tags FAILED
	expected undefined to equal 2
```

The `Object doesn't support this action` ones seem quite problematic :(

treshugart:
This is awesome. I think keeping it separate from `skate.create()` is a good idea if it makes the second argument redundant. Either that or we consider removing the second argument. It's quite convenient, though to be able to set a bunch of props right away. Skate constructors also allow a props argument (native does not, but we patch it).

Thoughts?

nicolasv:
> Either that or we consider removing the second argument. It's quite convenient, though to be able to set a bunch of props right away.

I'd leave `skate.create()` as a separate method with the `props` argument since it provides a nicer API for what might be the majority use case where you're creating a single element or a container element.

treshugart:
:+1:

nicolasv:
This has 1 test failure in IE11 that also exists on `master`:

```text
IE 11.0.0 (Windows 7) dom General DOM node interaction. should pick up descendants that are detached if an ancestor's innerHTML is set. FAILED
	AssertionError: expected false to equal true (/Users/nvenegas/Code/skatejs/skatejs/node_modules/karma-sinon-chai/node_modules/chai/chai.js:869)
```

PR #251, opened 2015-07-13
initDocument only inits after the document is ready (0.12.x)

### LINKED ISSUE #250: 0.12.x inits the document's elements prior to readyState complete/interactive
This can result in elements not getting init'd.

nicolasv:
Fixes #250.


nicolasv:
0.12.x and 0.13.x should have the fix from #249 backported, too.

Issue #255, opened 2015-07-15
Mutation observer not present in tests in IE.


treshugart:
This is because it was moved out of the Skate repo into a separate one. Just need to make sure it's loaded and that Skate's document observer is correctly accessing it.

PR #256, opened 2015-07-15
Fix ie tests

### LINKED ISSUE #254: IE doesn't allow innerHTML to be set on table elements so skate.create() fails.
See http://stackoverflow.com/questions/12613617/innerhtml-not-working-on-table-row-element-on-ie7.

treshugart:
Fixes #254 and #255 because they were discovered during this blitz. Also refactored tests to not use the all-in helpers file as it was becoming confusing and I had to add a helper anyways.


treshugart:
Closing and reopening another one because GitHub. See: https://github.com/isaacs/github/issues/18.

PR #257, opened 2015-07-15
Fix ie tests

treshugart:


treshugart:
Created in lieu of #256 because the other one accidentally pointed to 0.13.x instead of master, but we can't change destination.

treshugart:
One test is expected to fail because we haven't normalised that behaviour yet.

PR #258, opened 2015-07-20
Lifecycle exploration

treshugart:
Ok this PR should be GTG for review. What's in here:

- #276 - `skate.emit()` now simulates bubbling for browsers that don't support it on detached nodes. This is useful in situations where you have a detached host that has descendant components that trigger events to communicate with ancestors. Skate will ensure they bubble even if the element is detached from the DOM which ensures your component never becomes out of sync.
- #277 - The `created()` callback is now fired *after* descendants have been upgraded. This is so that you can do whatever you need in it without having to worry about descendants and if they've been upgraded. It also makes it so that you don't have to worry about any other lifecycle callbacks; everything just works in `created()`.
- #279 - Property value initialisation in custom properties are initialised *after* `created()` has been called, thus after descendants have been upgraded. They do not trump any values that have been manually set in `created()` and do not trump values if they came from attributes already defined on the element at the time of initialisation. This is so that any property setters that require descendants to be upgraded can be sure that they will be upgraded by the time the property itself is initialised.
- #278 - Event handlers are queued up and executed *after* `created()` has been called, thus after descendants have been upgraded. This is so that any event handlers that require descendants to be upgraded can be sure that they will be upgraded by the time they are invoked. This is executed after property initialisation so that if events require access to the initial value of a property that they can access it. Even if events are fired from the attribute setter, they will still be queued up and handled.

This behaviour assumes that no matter where you've loaded your `skate()` definitions that they're loaded in the proper order. This means that if `host-b` is used in `host-a` that `host-b` is loaded before `host-a`. This is easily achieved by splitting up your definitions into modules and properly declaring your dependencies for that module however your module format dictates.


src/lifecycle/attached.js
```diff
@@ -16,13 +16,11 @@ export default function (opts) {
     var info = data(this, opts.id);
     var isNative = this.attachedCallback;
 
-    if (info.attached) {
-      return;
-    }
-
+    if (info.attached) return;
     info.attached = true;
+    info.detached = false;
```
zephor: why was this re-ordered?

treshugart: Because it doesn't really matter if it's before or after the call and it reads better to be set all in one spot.

src/lifecycle/created.js
```diff
@@ -46,46 +49,79 @@ function callCreatedOnDescendants (elem, id) {
...
+    // Call created() on the host. This may be called at any time, so no one
+    // should not rely on descendants being initialised yet. This conforms with
+    // native behaviour.
     created.call(this);
-    triggerAttributesCreated(this);
+
+    // We trigger all property handlers for properties that exist once all
+    // descendants are ready for any incoming state updates. This will also
+    // trigger changes for any attributes that are properties.
+    initProps();
```
treshugart: I wonder if this should go between `callCreatedOnDescendants()` and `created.call()`.

src/lifecycle/created.js
```diff
@@ -46,46 +49,79 @@ function callCreatedOnDescendants (elem, id) {
...
+
+    // We trigger all event handlers that have queued up so that nothing has
+    // been lost since they were bound. This is done after initialising
+    // properties because events may depend on initialised values. Any events
+    // triggered from property initialisation will be queued up and initialised
+    // here.
+    initEvents();
```
treshugart: This still won't catch events triggered from descendants that have been initialised natively before the host. For example, if `host-component` requires `descendant-component` and they're loaded at the end of the `body`, `descendant-component` would be initialised before `host-component`. Any events triggered in the `created` callback of `descendant-component` will be lost.

treshugart:
When reviewing this, ignore `dist/*` and `lib/*`. Those were updating so that I could dogfood the branch before issuing the PR. If anyone things it's better I reset those folders then I can do that.

treshugart:
Anyone wanting to review this without the `dist` and `lib` files can just run:

```js
$('.js-details-container').has('[data-path^="dist/"], [data-path^="lib/"]').remove();
```

In your console.

test/unit/api/emit.js
```diff
@@ -1,5 +1,4 @@
 import emit from '../../../src/api/emit';
```
treshugart: This file was refactored and tests reordered to test simpler and dependent functionality first before it's used later in the tests.

test/unit/api/emit.js
```diff
@@ -7,10 +6,9 @@ describe('api/emit', function () {
   var triggered;
 
   beforeEach(function () {
-    fixture('<div><span></span></div>');
-
-    child = fixture().querySelector('span');
-    parent = fixture().querySelector('div');
+    child = document.createElement('child');
```
treshugart: We don't have to append the test element to a fixture anymore because we are simulating bubbling for trees not in the document.

test/unit/api/events.js
```diff
@@ -138,4 +138,33 @@ describe('api/event', function () {
     inst.focus();
     expect(focus).to.equal(true);
   });
+
+  describe('queue', function () {
```
treshugart: The new tests for event queueing before `created` is called.

src/api/emit.js
```diff
@@ -20,14 +23,54 @@ function createCustomEvent (name, opts) {
...
+  cevent = createCustomEvent(name, opts);
+  isBubbling = opts.bubbles;
+  status = elem.dispatchEvent(cevent);
+  elem = elem.parentNode;
+
+  // Simulate bubbling if the browser doesn't support it on detached elements.
+  if (isBubbling && !utilElementContains(document, elem) && !hasBubbleOnDetachedElements) {
+    let oldStopPropagation = cevent.stopPropagation;
+
+    // Patch stopPropagation() to set isPropagationStopped because there's no
```
nicolasv: I think this comment is out of date? (`isPropagationStopped` → `isBubbling`)

src/api/events.js
```diff
@@ -41,18 +40,24 @@ function bindEvent (elem, event, handler) {
...
+  var ready = false;
+
   Object.keys(events).forEach(function (name) {
-    bindEvent(elem, name, events[name]);
+    var handler = events[name].bind(elem);
+    bindEvent(elem, name, function (e) {
+      if (ready) {
+        handler(e);
+      } else {
+        queue.push(handler.bind(elem, e));
```
nicolasv: Is there any problem with calling `bind` twice on the original `events[name]` here? It seems OK to me (it's just that the original `this` can't be overridden it is set to the same object in both calls).

src/api/emit.js
treshugart: It is!

src/api/events.js
```diff
@@ -1,5 +1,4 @@
 import matches from '../util/matches-selector';
```
treshugart: Do we like the rename from singular to plural? I find that it better describes the intent of the function that it supports many events rather than one event.

src/api/events.js
treshugart: Will refactor that to be a little more clear.

treshugart: @nicolasv I ended up keeping the two `bind` calls here. The first one prevents the function from having to do `handler.call(elem, e)` within the wrapper handler every single time the event is handled. The second is so that the queue doesn't have to know the element and event objects to pass to the handlers. I removed one above in the `parseEvent` function that was unnecessary, though.

Happy to discuss changing the two binds above further if you still have any objections.

src/api/events.js
```diff
@@ -36,23 +35,28 @@ function bindEvent (elem, event, handler) {
   var parsed = parseEvent(event);
   var { name, selector } = parsed;
   var capture = selector && (name === 'blur' || name === 'focus');
-  handler = handler.bind(elem);
```
treshugart: @nicolasv This is the bind that I removed as stated in my comment below.

treshugart:
@nicolasv @zephor I'm not sure #278 is actually necessary because there are instances where it won't be able to catch an event if one is triggered in the `created` lifecycle (in native where definitions are at the bottom of the page). Maybe it's better to just bind events after and discourage triggering events in the `created` callback to communicate with the parent since the parent may not even be ready to catch events.

Thoughts?

CHANGELOG.md
```diff
@@ -33,95 +41,55 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
-Current:
-
-```js
-attribute (name, oldValue, newValue) {
-
-}
-```
-
-To translate the above variations to the new `properties` option, you would do the following:
+This is a little backward, though, since not all API points need, or should, have linkage between properties and attributes. Now, everything gets defined as properties and has an option to be linked to an attribute. To translate the above variations to the new `properties` option, you would do the following:
```
nicolasv: The opening sentence (“This is a little backward…”) seems to not have any context before it is introduced, i.e., the code above doesn't seem to (obviously) try to link the attribute to a property.

src/defaults.js
```diff
@@ -5,9 +5,10 @@ export default {
   attached: function () {},
 
   // Attribute lifecycle callback or callbacks.
-  attributes: undefined,
+  attributes: function () {},
```
nicolasv: Hasn't this been renamed to `attribute` (singular)?

src/lifecycle/created.js
nicolasv: That seems reasonable to me.

src/lifecycle/detached.js
```diff
@@ -14,15 +14,12 @@ export default function (opts) {
...
-      return;
-    }
 
+    if (info.detached) return;
     info.detached = true;
-    opts.detached.call(this);
-    isNative || callDetachedOnDescendants(this, opts.id);
     info.attached = false;
+
+    callDetachedOnDescendants(this, opts.id);
```
nicolasv: I wonder if the descendents should have `detached` called on them before the host?

src/api/emit.js
```diff
@@ -20,19 +31,48 @@ function createCustomEvent (name, opts) {
...
-export default maybeThis(function (elem, name, opts = {}) {
+export default function (elem, name, opts = {}) {
   var names = (typeof name === 'string' ? name.split(' ') : name);
   return names.reduce(function (prev, curr) {
-    if (!emitOne(elem, curr, opts)) {
+    if (emitOne(elem, curr, opts) === false) {
       prev.push(curr);
     }
     return prev;
   }, []);
```
nicolasv: The intention of this reduction is to return the events whose default was not prevented?

src/api/emit.js
```diff
@@ -20,19 +31,48 @@ function createCustomEvent (name, opts) {
...
+  };
+}
+
+function simulateBubbling (elem, cEvent) {
+  var status;
+  cEvent.stopPropagation = createReadableStopPropagation(cEvent.stopPropagation);
+  while (elem && !cEvent.isPropagationStopped) {
+    cEvent.currentTarget = elem;
+    if (elem.dispatchEvent(cEvent) === false) {
+      status = false;
```
nicolasv: I suggest renaming `status` to `didPreventDefault` (or similar) to make this a little easier to understand.

src/api/emit.js
treshugart: To return the names of the events that were cancelled (events whose default *was* prevented).

src/defaults.js
treshugart: Yeah. Addressed.

src/api/emit.js
treshugart: Good idea. Done.

src/lifecycle/created.js
treshugart: Done.

src/lifecycle/created.js
treshugart: @nicolasv would it make sense to do the same thing we did with `initProps()` above to this call, too?

treshugart: Also, @nicolasv, in light of recent discussion and how this might behave inconsistently is it worth event doing this? To give context, in native, if you put your definitions at the bottom of the page, then descendants will be queued for upgrade before the parent so the parent won't be able to catch any events that are triggered during the `created()` phase of a descendant.

src/lifecycle/detached.js
treshugart: I don't know of any use-cases where the ordering of this would actually matter, so I just decided to make it consistent with the others. Maybe it should be inverted, though. Ideas?

test/unit/lifecycle/created.js
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    // host requires child, so this is second
+    skate(`x-child-${tag}`, {
+      created () {
+        child = ++num;
+      }
+    });
+
+    // host has no dependants so its first
```
nicolasv: `its first` → `it's last`?

treshugart: Yeah, good ol out of sync comments.

treshugart: 'ol ;)

CHANGELOG.md
```diff
@@ -9,17 +10,24 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
 - #173 - The presence of the `resolved` attribute skips the `template` callback.
-- #177 - `skate.create()` as an alternative to `document.createElement()`.
+- #177 - `skate.create()` as an alternative to `document.createElement()` and `skate.init()`.
 - #187 - `attribute` handlers are now synchronous.
 - #200 - Support for a `properties` definition that defines instance properties and their behaviour.
 - #206 - Element constructors can be called like a function, or instantiated like a constructor.
+- #275 - Allow an object of properties to be passed to element constructors / functions and `skate.create()`.
+- #276 - `skate.emit()` triggers events and simulates bubbling for detached elements in browsers that don't support it.
 
 ### Changed
```
zephor: I assume 'changed' refers to breaking changes? Should there be links to the different sections of the upgrade guide here?

CHANGELOG.md
```diff
@@ -9,17 +10,24 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
```
zephor: I assume 'changed' refers to breaking changes? Should there be links to the different sections of the upgrade guide here?

CHANGELOG.md
```diff
@@ -33,95 +41,55 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
-    set (newValue, oldValue) {
+    set: function (newValue, oldValue) {
 
     }
   },
   myAttribute2: {
     attr: true,
-    value: 'initial value',
-    set (newValue, oldValue) {
+    init: 'initial value',
```
zephor: init vs default? (if that was an earlier discussion you can ignore this comment)

CHANGELOG.md
treshugart: That wouldn't be a bad idea at all. Wishing we had BitBucket tasks here (and approvals for that matter).

CHANGELOG.md
treshugart: Naming is hard. Happy to discuss alternatives:

- default
- init
- value

Though, `default` is a keyword and it's use is overloaded.

CHANGELOG.md
treshugart: @zephor https://github.com/skatejs/skatejs/issues/280

src/api/emit.js
treshugart: @zephor https://github.com/skatejs/skatejs/issues/281

PR #259, opened 2015-07-21
Bugfix for <object> tag edge case usage

a-ignatov-parc:
skate.js will fail to resolve `element.tagName` and `element.childNodes` in rare cases when processing `object` tag that is used to expose NPAPI plugin inteface to js.

screen shot 2015-07-21 at 12 08 56

### DIFF
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -382,7 +382,9 @@ function initElements (elements) {
       triggerLifecycle(element, currentNodeDefinitions[b]);
     }
 
-    var elementChildNodes = element.childNodes;
+    // When <object> tag is used to expose NPAPI api to js may have different behaviour then other
+    // tags. One of those differences is that it's childNodes can be undefined.
+    var elementChildNodes = element.childNodes || [];
     var elementChildNodesLen = elementChildNodes.length;
 
     if (elementChildNodesLen) {
--- a/src/registry.js
+++ b/src/registry.js
@@ -47,7 +47,10 @@ export default {
     var definitions = [];
     var isAttr = attrs.is;
     var isAttrValue = isAttr && (isAttr.value || isAttr.nodeValue);
-    var tag = element.tagName.toLowerCase();
+
+    // Using localName as fallback for edge cases when processing <object> tag that is used
+    // as inteface to NPAPI plugin.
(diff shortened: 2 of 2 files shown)


treshugart:
Sorry it's taken me so long to get around to reviewing this. Looks good, but would you mind writing a couple tests that ensure that `<object>` tags are handled correctly?

a-ignatov-parc:
No problems. Had a busy week too. Will try to recreate this use case with tests but not sure if it possible to create such `object` tag without NPAPI plugin. Need to investigate.

treshugart:
It looks like the element interface inherits from HTMLElement so you might have to mock it.

treshugart:
@a-ignatov-parc I was interested in seeing if this could be easily mocked. It seems that the tests I wrote in #272 break 0.13.x and would probably be fixed by this PR? Can you give those a quick review? Then we can merge those into this maybe and then get this into 0.13.x.

treshugart:
Sorry, the actual PR is #273. #272 is the issue it references.

treshugart:
Hey @a-ignatov-parc I ended up merging this and my test branch and releasing 0.13.9 for you guys. Hopefully I didn't step on your toes at all.

a-ignatov-parc:
@treshugart thanks a lot. Will check it in near future. Sorry for being so silent. Was hard working on big releasing for our project based on web-components built on top of skatejs.

treshugart:
Cool. What is it?

a-ignatov-parc:
Can't tell much right now but we are moving to web-component in our project with focus on creating reusable components without knowing much about how do they work. We need this because we have a lot of freelancers and a lot of small task like creating promo popups. And we don't want our core team to be evolved in this project except creating tools to make this real.

For example we need to show promo popups so we create base component that expose some props that can be changed. Next we give this component to freelancer who describe content of specific popup he need to create in html-like dsl overriding props in base popup component. In result we will have custom element that will show that promo popup.

treshugart:
Awesome! Thanks for sharing. Would love to see how it's being used someday
if you can ever show it.

a-ignatov-parc:
Sure! It is already in production but only for limited amount of users. Will release in public later in september-october ;)

PR #266, opened 2015-08-03
Check for camel cased version of attr before making property link

### LINKED ISSUE #265: Attribute-to-property links overwrite prototype properties that have the same camel cased named as the attribute
In 0.13.x, if you define a property (method, value, defineProperty) in the skate definition `prototype` that has the same name (after camel casing) as an attribute in the `attributes` property of the skate definition, it (the property as defined in `prototype`) will be overwritten by an attribute-to-property link.

For example, for this definition

```js
skate('x-foo', {
  prototype: {
    fooBar: function () {
      console.log('Call to fooBar().')
    }
  },
  attributes: {
    'foo-bar': {
      created: function () {
        // …
      }
    }
});
```
`document.createElement('x-foo').fooBar()` will blow up since `fooBar` is a property (in the `Object.defineProperty` sense) rather than a method.

nicolasv:
Fixes #265.

### DIFF
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -80,8 +80,8 @@ function initAttributes (target, component) {
  *
  * @returns {undefined}
  */
-function defineAttributeProperty (target, attribute) {
-  Object.defineProperty(target, camelCase(attribute), {
+function defineAttributeProperty (target, attribute, property) {
+  Object.defineProperty(target, property, {
     get: function () {
       return this.getAttribute(attribute);
     },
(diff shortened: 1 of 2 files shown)


test/unit/attributes.js
```diff
@@ -158,6 +162,14 @@ describe('Attribute listeners', function () {
         done();
       });
     });
+
+    it('should not override an existing camel cased property', function (done) {
+      myEl.setAttribute('overrideCamelCased', 'false');
```
nicolasv: This should actually be setting `override-camel-cased`.

src/lifecycle.js
```diff
@@ -111,7 +111,7 @@ function addAttributeToPropertyLinks (target, component) {
   }
 
   for (var attribute in componentAttributes) {
-    if (hasOwn(componentAttributes, attribute) && !hasOwn(target, attribute)) {
+    if (hasOwn(componentAttributes, attribute) && !hasOwn(target, camelCase(attribute))) {
```
treshugart: It'd be an unnecessary micro-optimisation at this point probably, but my only concern here is that we're doubling up on the `camelCase()` calls. Thoughts? I'm happy to just let it go through; I doubt there'll be any issues with it. Easy fix if there is.

treshugart:
:+1:

src/lifecycle.js
nicolasv: I agree in that we've opened up the of a bug since the knowledge that the property is camel cased is now spread in two functions.

Issue #260, opened 2015-07-27
Clarifying purpose of Skate on the website


TravisTheTechie:
Per a conversation with @treshugart...

> Skate provides an API to bind behaviours to DOM elements

And the benefits, or why one would use Skate to do this...

* Allows easy transition from selector-based behaviour binding to element binding
* Provides lifecycle hooks for elements
* Provides a superset of the custom elements spec
* Small, gzipped at 4kb (or whatever size it is)

So we can do some word smithing on this to make it a bit better, then implement it on the website. Once that's done, update the tag line on repo to match.

antitoxic:
I would vote for this. It's hard to explain what Skate is. Especially with the name, people are asking me what is this to do with web components.

Also asking me why should they use it over `<insert name of framework here>` - like angular, polymer and so on.

Simplicity, Performance?, Size... so on.

treshugart:
Since we should have web components in the title so that it shows up in relevant searches I was thinking about:

> Skate is a web component library that provides an API to bind behaviour to DOM elements. It's based on the W3C specification for Custom Elements.

treshugart:
Implemented in master and 0.13.x.

PR #283, opened 2015-08-17
#260 - changing wording of tag line and sync benefits between docs and frontpage

zephor:
I synced up the benefits on the front page with the benefits listed on the issue and on the docs subpage. In doing so I've had to remove some of the other benefits that were on the front page  like browser compatibility. Let me know if you think these are still important to put on the front page or we can find somewhere else to put them too.

Main thinking behind this is to make sure the docs and front page are in sync, it's better to have a consistent message across pages IMO.


treshugart:
:+1:

Issue #289, opened 2015-08-24
Only export the global in the dist


cyberdash:
This would remove the need for calling `skate.noConflict()`

treshugart:
@cyberdash can I close this?

treshugart:
Or can you ;)

PR #290, opened 2015-08-24
#289 remove window.skate when importing through src/index.js

cyberdash:
Since we change the entry point for the dist task, I've had to duplicate the build. This should be temporary until we're using a version of chippy that supports configuration.

If we don't want to duplicate the build like this, we could change the module system entry point, and leave the build as-is.

Also worth noting that `skate.noConflict()` is now only present when consuming skate through the dist (and it's on the window). It doesn't really make sense to define it if there's nothing it could conflict with.


test/unit/api/no-conflict.js
```diff
@@ -1,9 +0,0 @@
-import skate from '../../../src/index';
```
treshugart: I'm thinking we should still keep the test but just alter it to use the global module.

treshugart: Oh wait, looking at file above. Disregard. (I was looking from bottom up because lib files are at top)

test/unit/api/global.js
```diff
@@ -0,0 +1,15 @@
+import skate from '../../../src/global';
```
treshugart: :+1:

treshugart:
:+1:

PR #292, opened 2015-08-24
Fixed rare issue with components initialization

a-ignatov-parc:
Found today problem with rendering simple DOM tree in browsers that doesn't support custom-elements natively (all except Google Chrome).

DOM tree example
```xml
<component1>
	<section class="some-class"></section>
	<section class="some-class2">
		<component2 id="id1">
			<content>
				Some content
			</content>
		</component2>
		<span></span>
		<component2 id="id2">
			<content>
				Some content
			</content>
		</component2>
		<span></span>
		<component2 id="id3">
			<content>
				Some content
			</content>
		</component2>
	</section>
</component1>
```

Last `<component2>` node that was not initialized after appending to document. Problem was with `MutationObserver` shim that was not firing event on appending child nodes to `section.some-class2`.

I was trying to recreate issue in tests with simple DOM manipulations but with no luck. `MutationObserver` shim was working properly.

So here is a list of circumstances that could lead to this problem:

1. `<component2>` on `attached` callback moves some child nodes to `component2.nextSibling`. This behaviour is similar to `<label>` element that move inappropriate elements away.
1. Around 100+ components and subcomponents are rendering on the page at the same time.
1. Components' content renders with [virtual-dom](https://github.com/Matt-Esch/virtual-dom) lib.
1. Changes to document are made in batches with using [fastdom](https://github.com/wilsonpage/fastdom) lib.

P.S. I will continue my attempts to reproduce this problem in tests. But for now i wasn't be able to do this

---

I guess my problem is related with [this](https://github.com/skatejs/skatejs/blob/0.13.x/src/mutation-observer.js#L183) line of code that can lead to increasing [elements length](https://github.com/skatejs/skatejs/blob/0.13.x/src/lifecycle.js#L370). Need to investigate... maybe there should be another fix...

---

If i am right the problem is in [this](https://github.com/skatejs/skatejs/blob/0.13.x/src/mutation-observer.js#L217) line of code. Because `MutationObserver` callback is synchronously fired before cleaning `lastBatchedRecord` this could lead to adding other nodes exactly like in case i faced today. Native MutationObserver callbacks are fired at the "end of micro-task" (there is a good explanation in [this](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) article) and i am afraid can't be achieved properly with `setTimeout`. Only promises behave same way but that is not an option for browsers that does not have support for them.

Will make another fix tomorrow.

### DIFF
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -14,6 +14,7 @@ import {
   objEach
 } from './utils';
 
+var {Node} = window;
 var elProto = window.HTMLElement.prototype;
 var nativeMatchesSelector = (
   elProto.matches ||
@@ -367,12 +368,14 @@ function triggerLifecycle (target, component) {
  * @returns {undefined}
  */
 function initElements (elements) {
-  var elementsLen = elements.length;
-
-  for (var a = 0; a < elementsLen; a++) {
+  // [CATION] Browsers that don't support custom-elements natively in very rare cases won't
+  // trigger mutation observer event on subtree changes when component add siblings on `attached`
+  // lifecycle hook. This could lead to issue with components initialization placed at the end
+  // of childNodes list because they will be out of range of cached length value.
+  for (var a = 0; a < elements.length; a++) {
     var element = elements[a];
 
-    if (element.nodeType !== 1 || element.attributes[ATTR_IGNORE]) {
+    if (element.nodeType !== Node.ELEMENT_NODE || element.attributes[ATTR_IGNORE]) {
       continue;
     }
 
@@ -403,12 +406,11 @@ function initElements (elements) {
  * @returns {undefined}
  */
 function removeElements (elements) {
-  var len = elements.length;
-
-  for (var a = 0; a < len; a++) {
+  // Don't cache `childNodes` length. For more info see description in `initElements` function.
+  for (var a = 0; a < elements.length; a++) {
     var element = elements[a];
 
-    if (element.nodeType !== 1) {
+    if (element.nodeType !== Node.ELEMENT_NODE) {
       continue;
     }
 
--- a/src/lifecycle.js
+++ b/src/lifecycle.js
@@ -368,10 +368,10 @@ function triggerLifecycle (target, component) {
  * @returns {undefined}
  */
 function initElements (elements) {
-  // [CATION] Browsers that don't support custom-elements natively in very rare cases won't
-  // trigger mutation observer event on subtree changes when component add siblings on `attached`
-  // lifecycle hook. This could lead to issue with components initialization placed at the end
-  // of childNodes list because they will be out of range of cached length value.
+  // [CATION] Don't cache elements length! Components initialization could append nodes
+  // as siblings (see label's element behaviour for example) and this could lead to problems with
+  // components placed at the end of processing childNodes because they will change they index
+  // position and get out of cached value range.
   for (var a = 0; a < elements.length; a++) {
     var element = elements[a];
 
--- a/src/mutation-observer.js
+++ b/src/mutation-observer.js
@@ -7,7 +7,7 @@ import {
   objEach
 } from './utils';
 
-var Attr = window.Attr;
+var {Node, Attr} = window;
 var NativeMutationObserver = window.MutationObserver || window.WebkitMutationObserver || window.MozMutationObserver;
 var isFixingIe = false;
 var isIe = window.navigator.userAgent.indexOf('Trident') > -1;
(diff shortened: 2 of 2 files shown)


treshugart:
Yep, fully aware of microtasks etc. The only thing I could recommend trying is removing batching and setTimeout, but that'd have adverse affects on peformance most probably because batching helps to eliminate unnecessary opts iirc.

src/lifecycle.js
```diff
@@ -367,12 +368,14 @@ function triggerLifecycle (target, component) {
...
  */
 function initElements (elements) {
-  var elementsLen = elements.length;
-
-  for (var a = 0; a < elementsLen; a++) {
+  // [CATION] Browsers that don't support custom-elements natively in very rare cases won't
+  // trigger mutation observer event on subtree changes when component add siblings on `attached`
+  // lifecycle hook. This could lead to issue with components initialization placed at the end
+  // of childNodes list because they will be out of range of cached length value.
+  for (var a = 0; a < elements.length; a++) {
```
treshugart: My only gripe here is that it's cheaper in IE9 to save `elements.length` as `elementsLen` in super massive doms.

treshugart: But I guess we can't do that :(. I wonder what the performance would be like now.

treshugart:
Do you think we'll need to do anything in `master` to fix this? I'm not really sure how to reproduce it so I can't really help you there. Also, the mutation observer polyfill has been split from the core in `master`. It's now over at https://github.com/skatejs/polyfill-mutation-observer.

treshugart:
Couple more questions:

1. Does this happen if you remove the virtual-dom and fastdom layers?
2. It's perfectly acceptable to use those and this should be fixed for that reason, but it sounded like you wanted to land another fix? Do you want me to hold off on merging this until then?
3. Since you said "will make another fix tomorrow", does that mean that this doesn't fix the issue?

I'm not overly concerned about it, but it'd be interesting to see if this affects IE performance.

treshugart:
I think what you're experiencing in not being able to reproduce it might be because of fastdom. I'm not certain, but in order to batch dom mutations it'd have to override internal methods. It's probably delaying those which would mean you're right int that it becomes out of sync. You might be able to reproduce it by delaying an appendChild.

a-ignatov-parc:
Good morning. Yes i want to put this PR on hold until i finish my investigations. I hope i will complete them in 2-3 hours

a-ignatov-parc:
I think the problem was not with `fastdom` but with rendering pipeline. I'll try to explain:

1. DOM tree in my example renders
1. Subcomponentes are created and they moved child nodes to siblings.
1. MutationObserver fires handler and shim starts to collect changes.
1. Because `eTargetParent` is similar to `lastBatchedRecord.target` ([this](https://github.com/skatejs/skatejs/blob/0.13.x/src/mutation-observer.js#L174)) shim appends new nodes to already sent `childNodes` collection

I should notice that this problem take place because there were no other changes in DOM tree

a-ignatov-parc:
> Since you said "will make another fix tomorrow", does that mean that this doesn't fix the issue?

This PR fix my issue but it is more hack then proper fix. That is why I want to look into MutationObserver shim

a-ignatov-parc:
Well... the problem is not with `MutationObserver` but with traversing through `childNodes` [here](https://github.com/skatejs/skatejs/blob/0.13.x/src/lifecycle.js#L392). `childNodes` is live collection and i see only two way to handle this:

1. Freeze `childNodes`
1. Don't cache `length`

Second way is more performant then freezing.

src/lifecycle.js
a-ignatov-parc: Is super massive DOM trees is real world scenario or just synthetic test case?

a-ignatov-parc:
Updated comment and i think this fix is the best from performance point of view. But always open to discuss other ways to fix this problem.

treshugart:
:+1: let me know when you're ok for this to be merged. Are you able to check to see if this needs fixing on master? I'm happy if you just spot the part in master that needs changing rather than trying to integrate master to where you're reproducing the issue.

a-ignatov-parc:
Our QA team is still testing this fix. I'll post results here when they are done. Later will check master branch if it is need to be fixed ;)

treshugart:
Let me know when I can merge this :)

a-ignatov-parc:
Our QA team gives green light to this fix. You can merge it ;)

treshugart:
Created issues for master and the mutation observer polyfill repo.

a-ignatov-parc:
Checking master for possible changes right now. Going to create PR soon

Issue #270, opened 2015-08-08
Properties should take existing value when overwriting properties that already exist.


treshugart:
For example, if you specify a `textContent` property, it overwrites the property but does not take the existing value.

~~The `properties` `init` option should take precedence over existing values.~~

Properties should actually override `init` because `init` should be default if nothing is provided. Existing values on elements then take precedence.

treshugart:
PR'd in #303.

alexreardon:
I was thinking that `init` should be renamed to `fallback` (`default` would be best but it is a reserved word) to more clearly indicate it's behaviour

treshugart:
I'm going to keep this closed and raise the naming issue in #281.

PR #303, opened 2015-09-01
#270 - Add support for retaining existing property values if overriding.

treshugart:


src/lifecycle/properties.js
```diff
@@ -122,6 +123,8 @@ function defineProperty (elem, name, prop) {
       elem.attributeChangedCallback(prop.attr, null, elem.getAttribute(prop.attr));
     } else if (prop.init) {
       elem[name] = prop.init.call(this);
+    } else if (existingValue !== undefined) {
```
alexreardon: :+1:

test/unit/lifecycle/properties.js
```diff
@@ -218,4 +218,36 @@ describe('lifecycle/properties', function () {
     el.prop1 = '';
     expect(el.getAttribute('prop1')).to.equal('');
   });
+
+  describe('270', function () {
```
alexreardon: i think this should have a more descriptive name of the issue under test. Perhaps:

`Existing properties`

src/lifecycle/properties.js
```diff
@@ -122,6 +123,8 @@ function defineProperty (elem, name, prop) {
       elem.attributeChangedCallback(prop.attr, null, elem.getAttribute(prop.attr));
     } else if (prop.init) {
       elem[name] = prop.init.call(this);
```
alexreardon: Should init take precedence over what is on the element?

src/lifecycle/properties.js
```diff
@@ -120,6 +121,8 @@ function defineProperty (elem, name, prop) {
   return function () {
     if (prop.attr && elem.hasAttribute(prop.attr)) {
       elem.attributeChangedCallback(prop.attr, null, elem.getAttribute(prop.attr));
+    } else if (existingValue !== undefined) {
```
alexreardon: :+1:

alexreardon:
PR looks good to go @treshugart

test/unit/lifecycle/properties.js
treshugart: The sub tests will be appended to that, that's just to classify it. Could do that tho.

alexreardon:
I was thinking that perhaps instead of calling it `init` it should be called `fallback` (ideally `default` but that is reserved word) to be clearer about it's behaviour

Issue #281, opened 2015-08-14
Finalise property options


treshugart:
Currently we have:

- `attr`
- `init` see https://github.com/skatejs/skatejs/issues/270#issuecomment-136902990
- `type`
- `get`
- `set`

Do we want to change these?

### `attr`

Maybe:

- `attribute`
- `attrLink`
- `attributeLink`
- `linkAttribute`

Remember:

- If `true` it `dash-cases` the property name and uses that as the linked attribute.
- If `falsy` it doesn't link it at all.
- If a `string` it uses the provided string as the attribute name to link.

### `init`

Maybe:

- `default` (but that's a keyword)
- `defaultValue`
- `value` (clashes with normal `defineProperty` option, but we can get around that)
- `initialValue`

Remember:

- If it's a `function` then the return value of the function becomes the initial value.
- Any other value is the exact value of the property.

### `type`

Maybe:

- `coerce`

Remember:

- It should coerce the value, but you can also do type checking here or throw errors if it's empty (required). Type checking et al can also go in `set` if you really want.

### `get`

I think this should be removed altogether. It kind of defies the rest of the options. If you want to make a readonly property, them you can probably just put a getter on the `prototype`.

### `set`

Maybe:

- `change`
- `update`
- `set` is good, but it kind of implies there's also a corresponding `get`

Remember:

- It has the same semantics as a normal property setter, but it receives two parameters:
  - `newValue`
  - `oldValue`

### One last question...

Do we want to support normal `defineProperty` options?

treshugart:
@marcoms @alexreardon you guys probably have some opinions on this.

marcoms:
Since ES5 reserved words are OK to use as object keys, so "default" should be fine to use.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Reserved_word_usage

>Reserved words actually only apply to Identifiers (vs. IdentifierNames) . As described in es5.github.com/#A.1, these are all IdentifierNames which do not exclude ReservedWords.

>```js
a.import
a["import"]
a = { import: "test" }.
```

Probably best to use "default" as a quoted key if it's used, though.

Also:
https://developer.mozilla.org/en-US/docs/Talk%3AJavaScript/Reference/Reserved_Words
http://stackoverflow.com/questions/7022397/using-reserved-words-as-property-names-revisited

marcoms:
Also one possible reason to keep `get` is that it could be used to make computed properties.

```js
get() {
[...]

PR #304, opened 2015-09-01
Fix issue where properties weren't getting removed if a type was specified.

### LINKED ISSUE #293: RFC: Don't pass undefined and null values to property.type
The function passed to the `type` property of a property object is typically used to coerce values to the desired type before giving it to the property's setter. In doing this, though, the recently added ability to pass `undefined` or `null` to a component's property to remove the corresponding attribute is lost, and JS types don't always return `undefined` when passed `undefined` (e.g. `Number(undefined) -> NaN`, `String(undefined) -> "undefined"`).

It is possible to get around this by wrapping each JS type in your own function:

```js
function type(fn) {
  return (value) => {
    if(value === undefined || value === null) return value;
    return fn(value);
  };
}

...

myProperty: {
  type: type(Number),
},
```

...However it is a bit of a hassle :)

Is there a use case for the user handling `undefined` or `null` specially (where you could otherwise use `false`)?

treshugart:
If you specify a type and pass `null` or `undefined`, properties weren't being unset.

Fixes #293.


src/lifecycle/properties.js
```diff
@@ -77,7 +77,7 @@ function property (elem, name, prop) {
 
       if (isBoolean && internalValue) {
         this.setAttribute(prop.attr, '');
-      } else if (internalValue == null || isBoolean && !internalValue) {
+      } else if (value == null || isBoolean && !internalValue) {
```
treshugart: I'm not sure if we should be doing this as far down as we are here. What this means is that the value is still being coerced and passed to the setter. This could cause the attribute to be removed even if the coercing function made it a valid value.

marcoms: In my opinion if the value is `undefined` then I think that we should skip calling `prop.type()` on it altogether. I can't imagine anyone would want to handle `undefined` as a valid value for a component's property (perhaps `null` would be appropriate as it often intentionally denotes "not applicable", "empty", etc. for example in `Element#querySelector(".no-elements-with-this-class") === null`, `Element#getAttribute("no-such-attribute") === null`).

treshugart: Good call.

treshugart: @marcoms updated in 3b6dcaf.

treshugart:
> I'd say that the type coercion be skipped, but the setter still be called later on

That's probably a good idea. After finalising https://github.com/skatejs/skatejs/pull/307 I'd like to focus on the property lifecycle. See https://github.com/skatejs/skatejs/issues/281 for details.

marcoms:
>I'd say that the type coercion be skipped, but the setter still be called later on

@treshugart is there any scope for discussion at this time or do you want to wait until #281 is settled?

treshugart:
@marcoms sure, could be implemented alongside #281.

PR #306, opened 2015-09-02
Make getClosestIgnoredElement iFrame-safe

TheConstructor:
When skate is included in an iframe it still may receive events from the parent-document if they share domain. Then ```document``` will refer to the iframe's document while climbing up parentNodes will lead us to the parent-document's document

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -84,7 +84,8 @@ export function debounce (fn) {
 export function getClosestIgnoredElement (element) {
   var parent = element;
 
-  while (parent && parent !== document && !(parent instanceof DocumentFragment)) {
+  // e.g. document doesn't have a function hasAttribute; no need to go further up
+  while (parent && typeof parent.hasAttribute === 'function' && !(parent instanceof DocumentFragment)) {
     if (parent.hasAttribute(ATTR_IGNORE)) {
       return parent;
     }


treshugart:
Are you able to write a test for this fix?

src/utils.js
```diff
@@ -84,7 +84,8 @@ export function debounce (fn) {
 export function getClosestIgnoredElement (element) {
   var parent = element;
 
-  while (parent && parent !== document && !(parent instanceof DocumentFragment)) {
+  // e.g. document doesn't have a function hasAttribute; no need to go further up
+  while (parent && typeof parent.hasAttribute === 'function' && !(parent instanceof DocumentFragment)) {
```
a-ignatov-parc: It is better to check for `document` constructor:

```javascript
while (parent && !(parent instanceof HTMLDocument) && !(parent instanceof DocumentFragment)) {
    // ...
}
```

treshugart: True, but since we're only calling `hasAttribute` here, we could probably just simply check:

```js
while (parent && parent.hasAttribute) {
```

TheConstructor:
@treshugart sorry, but I am not that firm into JS-testing. I just happen do stumble about this bug when writing a Jira-Gadget.

src/utils.js
a-ignatov-parc: Not only `document` element doesn't have `hasAttribute` method. constructor check is way more bulletproof and explicit.

TheConstructor: Your future-safe when you check for function-existence. If you want to ensure that you reached top-level, you could also check validity of `parent.parentNode` instead of using `instanceof`.

a-ignatov-parc: Also no one stops me from doing this ;)
screen shot 2015-09-02 at 13 41 11

treshugart: > Not only document element doesn't have hasAttribute method

That's why I suggested it.

> constructor check is way more bulletproof and explicit.

That's true. So maybe we should re-evaluate exactly what we want to get from this check. In my mind we want to ensure that we can call `hasAttribute` on it, right? There's a few options here:

1. Check `instanceof` on both `HTMLDocument` and `DocumentFragment`
2. Check `instanceof` on just `Element` (since that's the interface that implements `hasAttribute`)
3. Simply check if the `parent` has `hasAttribute`

I don't mind option 3. I don't see any real benefits to being super explicit here. If it has `hasAttribute` then we should be able to call it, right?

Option 2 is probably a good balance between the two. Concise and we have two checks instead of one.

treshugart:
@TheConstructor would you be able to provide steps to reproduce the issue then?

src/utils.js
treshugart: You'll likely break a lot more than Skate if you do something like `document.hasAttribute = true`, but why would one do that in the first place?

treshugart: I'm happy to go with option 2, I just have a preference for duck typing in JavaScript. Re "future safe", this is also true. They've definitely changed inheritance chains between DOM Level versions so checking for exactly what you want to call would have been better here.

a-ignatov-parc: :+1: for option 2

TheConstructor:
This is vaguely what happens:
# load iframe with skatejs
# add element to outer document
# `getClosestIgnoredElement` is called by event handler

It seems that at least Firefox 38.0.5 handles element-addition-events "globally" scoped.

src/utils.js
treshugart: > Check instanceof on just Element (since that's the interface that implements hasAttribute)

@TheConstructor are you able to implement option 2, please?

treshugart:
Hey @TheConstructor and @a-ignatov-parc, I pulled this into #314, wrote a test and updated the `while` condition.

treshugart:
Declining in favour of that one. Thanks again @TheConstructor, this doesn't discount the work that you've done.

TheConstructor:
Thank you @treshugart! Been on vacation and just returned today. Did hope to be able to check back on this before, but wasn't.

treshugart:
No worries @TheConstructor hope you had a good time and this is now working for you. Let us know if you have any issues.

Issue #285, opened 2015-08-19
created function called after property initialisation


marcoms:
It seems that skated elements' created callbacks are now executed *after* properties are initialised. Not sure if this is intended behaviour or not..?

treshugart:
Yeah, it was intentional, but I'm remembering you raised #179 awhile back. Is this biting you in the ass? It'd be good to discuss the benefits and caveats of its placement.

Pros:

1. Everything is set up for you by the time created is invoked.
2. If you use `template` and `properties` together, `created` may rarely be needed.

Cons:

?

Not sure yet because any property behaviour I put in the property setter. I rarely have to use the `created` callback, but I see it useful for edge cases.

Ideas?

treshugart:
We used it during ShipIt (Atlassian's hackathon) and came to a fork in the road on this. Thought this is different from what you've described, I feel that it relates to it. What we decide here will ultimately decide the fate of this issue.

### Before `template`

Your template can be rendered dynamically, however, you cannot make alterations to the template on property initialisation as the templated nodes wouldn't exist yet. Modifications after `template` has been called will work.

This means your DOM alterations all go in `template`. If you need to re-render your template when a property changes you must call `template`; you should not use property setters to do this. This isn't all bad, but if you're doing bare-bones templating, you're smashing the DOM pretty hard. It does make for less side-effects and you can use a DOM diffing library if you don't want to thrash it. For most use-cases, this is probably a fairly practical approach.

### After `template

DOM modifications can be done piecemeal from within your property setters without worry that the template won't be there when they're initialised. You may use them in your templates, but their initial values won't be ready when `template` is called. If you want them to affect your template, then you will have to ensure that the renderer is called when properties are updated.

### My feelings

Up until now, I felt that properties should be initialised after `template` because I like putting my DOM alterations in the setters. However, this creates a lot of side-effects. It does initially help save you from yourself so you don't have to re-render an entire template for a small change, but you can ensure the DOM isn't thrashed from within whatever rendering engine you use within the `template` callback. 

My vote: before `template`. This means that properties will also be initialised before `created` which is what you originally had concerns about. What are your thoughts.

cc @alexreardon.

marcoms:
Well I actually worked around properties being initialised before created by moving `created` code into `template`.

At the end of the day, I don't mind too much whether property init goes before or after `created`, however having properties initialised before template poses some problems for me (having always had DOM mutation in property setters):

[...]

Issue #269, opened 2015-08-03
RFC: a way of specifying properties that cause a property to be re-evaluated


marcoms:
For example:

```js
properties: {
  foo: {
    observe: [
      "bar",
    ],

    set(val) {
      console.log("foo set() called");
    }
  },

  bar: {},
}

// ...

let myEl = MyEl();
myEl.bar = 99;  // "foo set() called"
```

I was previously using `this.propertyName = this.propertyName` to trigger this manually, but it has stopped working in one of skate's latest releases (`master`/`lifecycle-exploration`), probably as a measure to improve performance, however now it is impossible to manually trigger the property's setter unless you move the setter's code into the `prototype` and to call that method whenever you want to re-evaluate the property.

treshugart:
Yeah, those are still in the experimental stage. FWIW I do find them extremely useful but when I brought it to the table the feeling I got was that they should be a separate extension to Skate as it didn't feel like they should belong to the base library. I was of two minds. On one side it was very useful, but on the other it wasn't a complete implementation.

I'm happy to try and get it going again but I think they need to be thought of in terms of a simple, but fairly complete solution.

- Observable properties on host.
- Host able to observe properties on descendants.
- Host able to observe properties on future descendants.
- Dependencies placed on all types of observable properties from within a single property definition.
- Should we observe

marcoms:
The "observe" name was probably poor nomenclature on my part for what I wanted to achieve at a basic level, which is to just have a way to manually trigger a property's setter. Having this done declaratively using `observe` would be ideal, however any other solution that would enable me to trigger a properties setter would resolve this for me, perhaps even in the skate API.

treshugart:
I think a certain level of this would be useful. In your example above, what would `observe` mean? Would it be the property that declared it listening to the properties in the `observe` array and triggering its `set` when they fire? Or would it be the other way around?

Another question is would this API cover only properties on the component in which they're declared, or would it listen for properties on descendant components?

I'm interested in fleshing out this idea more as it would be a useful feature, even if it was a plugin.

marcoms:
>the property that declared it listening to the properties in the observe array and triggering its set when they fire

That's right

>would this API cover only properties on the component in which they're declared, or would it listen for properties on descendant components?

I would expect that by default it would only cover the single component, whereas something could optionally be passed to the observe property to specify the latter (like `HTMLElement#clone(deep=false)`)

Perhaps like:

```js
[...]

PR #307, opened 2015-09-02
#285 - Rename template to ready and reorder lifecycle.

treshugart:
References discussion in #285. More specifically, it implements https://github.com/skatejs/skatejs/issues/285#issuecomment-133372496.

## Before

1. bind `events`
2. define `properties`
3. call `template`
4. invoke `created` lifecycle on descendants
5. set initial property values
6. call `created` on host
7. execute queued event handlers

## After

1. define `properties` and only initialise internal value to the init option, or the overridden property's value
2. bind `event` listeners
3. call `created` on host
4. invoke entire `created` lifecycle on all descendants
5. call `ready` on host

## Notes

- one step closer to #269: added experimental `skate.property` event that is emitted when a property changes
- `template` was renamed to `ready`
- `ready` is basically the new version of `created`
- `created` is now invoked *before* descendant initialisation
- both `created` and `ready` provide before and after hooks around descendant initialisation
- the property `get` callback was removed
- the property `set` callback was renamed to `update`
- `properties` don't trigger `update` on initialisation
- overridden property values now correctly set the initial property value for the new property
- tests were added to cover common scenarios when using properties and templating


test/unit/lifecycle/properties.js
```diff
@@ -219,35 +219,98 @@ describe('lifecycle/properties', function () {
     expect(el.getAttribute('prop1')).to.equal('');
   });
 
-  describe('270', function () {
-    function setup (initialValue) {
-      var el = elem.create();
-      el.textContent = 'existing content';
+  it('should not be triggered on initialisation', function () {
```
treshugart: @nicolasv, @cyberdash this refers to where we were saying that the `update()` callback should not be called for the property initialisation.

test/unit/lifecycle/properties.js
```diff
@@ -219,35 +219,98 @@ describe('lifecycle/properties', function () {
...
+    expect(newValue).to.equal('one');
+    expect(oldValue).to.equal(undefined);
+
+    el.prop = 'two';
+    expect(triggered).to.equal(true);
+    expect(newValue).to.equal('two');
+    expect(oldValue).to.equal('one');
+  });
+
+  it('should override existing properties', function () {
```
treshugart: Basic overriding.

test/unit/lifecycle/properties.js
```diff
@@ -219,35 +219,98 @@ describe('lifecycle/properties', function () {
...
+    });
+
+    let el = skate.create(`<${elem.safe}>initial content</${elem.safe}>`);
+    expect(el.textContent).to.equal('initial content');
+    el.textContent = 'updated content';
+    expect(el.textContent).to.equal('updated content');
+    expect(el.innerHTML).to.equal('initial content');
+  });
+
+  describe('templating integration', function () {
```
treshugart: @nicolasv, @cyberdash these are the common scenarios we discussed. We didn't discuss what happens if you override `innerHTML`, though. The skinny of it is that if you override `innerHTML` then you can't use it for templating in scenario 2, however, it will still work in scenario 1. In scenario 2, though, we'd probably use a dom differ of some sort anyways which wouldn't be using innerHTML.

src/lifecycle/properties.js
```diff
@@ -90,50 +58,60 @@ function property (elem, name, prop) {
...
     // is useful information for the setter.
-    if (internalSetter) {
-      internalSetter.call(this, newValue, oldValue);
+    if (prop.update) {
+      prop.update.call(this, newValue, oldValue);
+    }
+
+    // If we are emitting notify the element of the change.
+    if (prop.emit) {
+      emit(this, prop.emit, {
```
treshugart: This is the addition of the `skate.property` event. @marcoms will probably be happy about this. I think it's not disputed that we do need this, but I would like to discuss whether or not we should be emitting:

1. `skate.property`
2. `skate.property.[name]`
3. one or both of the above

A case for 1 is a template that wants to listen to all property updates.

A case for 2 is possibly performance so handlers aren't invoked for every property change. The downside is that you *must* be specific about which properties you want to listen to.

A case for both is flexibility, but performance might suffer when a lot of  state changing is happening.

marcoms:
Oops mistapped a button...

Sorry @treshugart was this not previously exposed through `prop.notify`? Is this now `prop.emit`?

marcoms:
Just had a look at the source code and apparently `notify` was not implemented, sorry for the confusion.

treshugart:
It was being experimented with at some point, but it was never stable.

PR #308, opened 2015-09-03
Fix missing ignore section in bower.json

joscha:
This fixes:
`bower invalid-meta  skatejs is missing "ignore" entry in bower.json`

and also prevents downloading unnecessary files & folders


joscha:
We could also think about excluding `docs` and `README.md` if a bower install is considered being a production install.

treshugart:
@joscha thanks for the PR! Yeah, I think all we really need to expose in a bower install is `dist/`, `lib/` and `src/`.

joscha:
@treshugart cool, I'll add `docs` and `README.md` to the commit. One rebase coming up.

Issue #312, opened 2015-09-07
Update readme to reflect revised lifecycle order/naming


marcoms:
Relevant section: https://github.com/skatejs/skatejs/tree/master#component-lifecycle

treshugart:
Done in #313. Can you please review when you have a chance?

PR #313, opened 2015-09-07
#312 - Update README for changes to component lifecycle.

treshugart:


README.md
```diff
@@ -377,10 +374,10 @@ skate('my-element', {
 The component lifecycle consists several callbacks:
 
 - `created`
```
bqx: It might be good to also indicate that the prototype/properties are set up here

bqx:
:+1: lgtm

marcoms:
By the way the table of contents still references the old template function

alexreardon:
Was a decision made about going down the function composition path?

treshugart:
@alexreardon as discussed IRL, yeah. Staying declarative unless it's absolutely necessary. We're pretty confident that this is the final big lifecycle change and should facilitate any templating / component route people would want to take.

treshugart:
Done, done, done and done :)

treshugart:
@bqx @marcoms I've also updated this removing a lot of the wordy stuff around the lifecycle in hopes that the ordered list is good enough along with the following description and `attribute` description.

I've also removed the wordy templating stuff. Even if this gets added back in (see #321) then the specifics around integration should probably be left out of the readme and put in the website or in the form of a blog / tutorial or something.

PR #314, opened 2015-09-08
#306 - Make getClosestIgnoredElement iFrame-safe

treshugart:
PR that pulls in fix from #306 and updates the check in the `while` loop. Added test to ensure `getClosestIngoreElement()` works for elements in an iframe.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -85,7 +85,7 @@ export function getClosestIgnoredElement (element) {
   var parent = element;
 
   // e.g. document doesn't have a function hasAttribute; no need to go further up
-  while (parent && typeof parent.hasAttribute === 'function' && !(parent instanceof DocumentFragment)) {
+  while (parent instanceof Element) {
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -84,7 +84,8 @@ export function debounce (fn) {
 export function getClosestIgnoredElement (element) {
   var parent = element;
 
-  while (parent && parent !== document && !(parent instanceof DocumentFragment)) {
+  // e.g. document doesn't have a function hasAttribute; no need to go further up
+  while (parent instanceof Element) {
```
joscha: :+1:

joscha:
LGTM

treshugart:
@TheConstructor I've released this in 0.13.12.

src/utils.js
a-ignatov-parc: :+1:

TheConstructor:
That's great! Thank you!

PR #315, opened 2015-09-08
Fix overriding prototypes

bqx:
This fixes an issue where skating an element more than once would fail to update the prototype after the first time. This happened because the check for `isNative` just checked for the presence of `createdCallback` on the element, which would be set the first time the element was skated.

The solution is to store the result of the first check for `isNative` and use that value instead.

### DIFF
--- a/src/lifecycle/created.js
+++ b/src/lifecycle/created.js
@@ -55,7 +55,7 @@ export default function (opts) {
 
   return function () {
     let info = data(this, opts.id);
-    let isNative = this.createdCallback;
+    let isNative = !this.hasOwnProperty('_isPolyfill') && this.createdCallback || !(this._isPolyfill = true);
     let isResolved = this.hasAttribute(opts.resolvedAttribute);
 
     if (info.created) return;
--- a/src/lifecycle/created.js
+++ b/src/lifecycle/created.js
@@ -55,7 +55,7 @@ export default function (opts) {
 
   return function () {
     let info = data(this, opts.id);
-    let isNative = this.createdCallback;
+    let isNative = opts.isNative;
     let isResolved = this.hasAttribute(opts.resolvedAttribute);
 
     if (info.created) return;
--- a/src/lifecycle/created.js
+++ b/src/lifecycle/created.js
@@ -55,7 +55,7 @@ export default function (opts) {
 
   return function () {
     let info = data(this, opts.id);
-    let isNative = !this.hasOwnProperty('_isPolyfill') && this.createdCallback || !(this._isPolyfill = true);
+    let isNative = this.createdCallback;
     let isResolved = this.hasAttribute(opts.resolvedAttribute);
 
     if (info.created) return;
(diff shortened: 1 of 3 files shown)


src/lifecycle/created.js
```diff
@@ -55,7 +55,7 @@ export default function (opts) {
 
   return function () {
     let info = data(this, opts.id);
-    let isNative = this.createdCallback;
+    let isNative = !this.hasOwnProperty('_isPolyfill') && this.createdCallback || !(this._isPolyfill = true);
```
bqx: This is equivalent to:

```
let isNative = false;
if (!this.hasOwnProperty('_isPolyfill')) {
  isNative = !!this.createdCallback;
  if (!isNative) {
    this._isPolyfill = true;
  }
}
```

treshugart: Would it be better to store the result of this on the definition (opts): https://github.com/skatejs/skatejs/blob/master/src/index.js#L69?

test/unit/registration.js
```diff
@@ -103,6 +103,34 @@ describe('Returning a constructor', function () {
...
+      prototype: {
+        func1:  function () {
+          return true;
+        }
+      }
+    });
+    skate.init(element);
+
+    element.setAttribute('test-attr', '');
+    skate('test-attr', {
```
treshugart: Should probably use the safe tag helper here, too.

test/unit/registration.js
```diff
@@ -103,6 +103,34 @@ describe('Returning a constructor', function () {
...
+    var element = document.createElement(tag.safe);
+
+    skate(tag.safe, {
+      prototype: {
+        func1:  function () {
+          return true;
+        }
+      }
+    });
+    skate.init(element);
```
treshugart: You can remove this call and rely on the one before the assertions.

test/unit/registration.js
```diff
@@ -103,6 +103,34 @@ describe('Returning a constructor', function () {
     expect(called).to.equal(true);
   });
 
+  it('should merge prototype functions if an element is skated more than once', function () {
+    var tag = helperElement('my-el');
+    var element = document.createElement(tag.safe);
+
+    skate(tag.safe, {
```
treshugart: Instead of having to do the run around of passing `tag.safe` to `skate()` I created another method in the `element` helper:

https://github.com/skatejs/skatejs/blob/master/test/lib/element.js#L14

src/lifecycle/created.js
treshugart: > This is equivalent to:

Yep, but if we can pre-calculate then we don't have to do it multiple times.

Also, if you're going to store information on an element, probably better to use `util/data`.

treshugart: > Yep, but if we can pre-calculate then we don't have to do it multiple times.

What I mean by that is that we also check for whether or not a definition is native in the `attached` and `detached` callbacks. If we can pre-calc this, then we can also reuse that in those callbacks.

treshugart:
:+1:

PR #317, opened 2015-09-10
Fix updating the property when setting a linked attribute

### LINKED ISSUE #316: Setting a linked attribute doesn't change the value of the property
Setting the attribute directly doesn't change the property.

```
    skate('my-el', {
      properties: {
        test: {
          attr: 'test2'
        }
      }
    });

    ...

    el.test = 'value1';
    el.setAttribute('test2', 'value2');
    console.log(el.test); // will print 'value1' instead of 'value2'
```

bqx:
PR to fix #316

### DIFF
--- a/test/unit/lifecycle/properties.js
+++ b/test/unit/lifecycle/properties.js
@@ -76,6 +76,41 @@ describe('lifecycle/properties', function () {
     expect(el.getAttribute('my-attr')).to.equal('test2', 'attribute updated to new value');
   });
 
+  describe('attribute - setting', function() {
+    it('with `attr` set to true', function() {
+      skate(elem.safe, {
+        properties: {
+          test: {
+            attr: true
+          }
+        }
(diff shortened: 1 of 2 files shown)


test/unit/lifecycle/properties.js
```diff
@@ -76,6 +76,41 @@ describe('lifecycle/properties', function () {
...
+        }
+      });
+
+      var el = fixture(`<${elem.safe}></${elem.safe}>`).querySelector(elem.safe);
+      skate.init(el);
+
+      el.test = 'value1';
+      el.setAttribute('test', 'value2');
+      expect(el.test).to.equal('value2');
+    });
```
bqx: This test passed without any of the changes made, but is added here for better coverage.

treshugart: +1. Thanks!

Issue #319, opened 2015-09-12
Server-side rendering.


treshugart:
PR: #320.

In that PR, it invokes the entire lifecycle and it would be up to the developer to not template again on the client side. For this reason it would probably be a good idea to have a `template` callback that gets fired after `created`, but before `callCreatedOnDescendants()` (thus before `ready`) that will be invoked if the element doesn't have the `resolved` attribute.

treshugart:
Closing because this is all facilitated now. The rest of the work involves writing a server-side script that will call the created() and render() callbacks on the custom elements and ensuring Skate can be loaded sever-side.

PR #320, opened 2015-09-12
#319 - Server-side rendering.

treshugart:
References #319.


treshugart:
Declining this since the API has changed. We should still implement server-side rendering, but now it won't need to have as large of a footprint in terms of DOM dependencies (if hardly any at all).

Issue #321, opened 2015-09-12
Re-implement template callback for server-side rendering.


treshugart:
`template` callback implementation for #319. It should:

- be called after `created`
- be called before descendants are initialised
- therefore be called before `ready`
- not be called if the `resolved` attribute is on the element by the time the lifecycle gets to it

This is different from the previous `template` lifecycle in that it is called after created and before descendants are initialised. This will work for most any templating scenario.

treshugart:
This has been implemented as `render`. Server-side rendering isn't solved, and may still involve the entire lifecycle, but the `render` callback needs that meaning because it will be skipped if the element is already resolved. In the case where it is skipped, Skate still binds behaviour to the element.

PR #322, opened 2015-09-12
#321 - (Re-)implement template lifecycle callback as "render".

treshugart:
Implements #321. Will make it so that #319 will work and so that #320 can be updated to do the bare minimum of descending through a list of elements invoking the `template` callback on them if we find that's the best solution. At the very least, we have an opinionated mechanism for server-side rendering in that the `template` callback will not be called if the element is already resolved (already templated).


alexreardon:
I am confused as to how this plays with the new `ready` callback

treshugart:
The order with this will *always* be:

- `events`
- `properties`
- `created`
- `template`
- `callCreatedOnDescendants`
- `ready`

alexreardon:
I am confused about the need for a 'template' function when we have `created` and `ready`. Also, naming it `template` might be misleading as we are now leaving it up to the developer when to template.

treshugart:
Going to decline this for now. I don't think it's necessary at this point.

treshugart:
Renamed `template` to `render` as it more clearly describes an action.

treshugart:
I think this should be a thing because of the fact that it will only render if the element hasn't been `resolved` yet. Making users worry about this concern sucks if we can provide a specific hook for it that takes care if it.

PR #323, opened 2015-09-12
Use native callback names.

treshugart:
There's several reasons behind me spiking this:

1. Make the transition from the spec'd callbacks to Skate as simple as possible.
2. Using `this` inside the callback feels more natural.
3. Simplification and better mind mapping of internals.
4. Skate is now inherently compatible with vanilla `document.registerElement()`.
5. Extending base class definitions is simpler internally.

Changes made are:

- `created` -> `prototype.createdCallback`
- `ready` -> `prototype.readyCallback`
- `attached` -> `prototype.attachedCallback`
- `detached` -> `prototype.detachedCallback`
- `attribute` -> `prototype.attributeChangedCallback`

All callback ordering stays the same.

Obviously, if you pass in a vanilla definition, it will insert them into the Skate lifecycle. However, this shouldn't introduce any breakages into native custom elements since Skate doesn't change the order of the native callbacks. Even the `createdCallback` will behave like native because we don't actually force init descendants until after it's called.

I realise this introduces breaking changes in the names and where you define them, however, I think offering two ways of doing something, even if deprecated, in development versions is not ideal. However, if you guys think we should offer deprecations then maybe we should.

### DIFF
--- a/src/api/init.js
+++ b/src/api/init.js
@@ -1,5 +1,3 @@
-import attached from '../lifecycle/attached';
-import created from '../lifecycle/created';
 import elementContains from '../util/element-contains';
 import registry from '../global/registry';
 import walkTree from '../util/walk-tree';
@@ -12,12 +10,12 @@ export default function (element) {
     var componentsLength = components.length;
 
     for (let a = 0; a < componentsLength; a++) {
-      created(components[a]).call(descendant);
+      components[a].prototype.createdCallback.call(descendant);
     }
 
     for (let a = 0; a < componentsLength; a++) {
       if (isInDom) {
-        attached(components[a]).call(descendant);
+        components[a].prototype.attachedCallback.call(descendant);
       }
     }
   });
--- a/src/index.js
+++ b/src/index.js
@@ -52,11 +52,11 @@ var initDocument = debounce(function () {
     var componentsLength = components.length;
 
     for (let a = 0; a < componentsLength; a++) {
-      created(components[a]).call(element);
+      components[a].prototype.createdCallback.call(element);
     }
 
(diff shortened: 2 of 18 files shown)


marcoms:
Honestly I prefer the current API but the justifications you made are valid. Why not have `*Callback`s as working identifiers for compatibility with the spec and to ease migration, but with the current `created` `ready`, ... as the preferred names?

treshugart:
The alternative here would be to make the callbacks not use `this` and instead take an argument that is the element like in 0.13. There's a couple benefits to this:

- You can now access the definition via `this` in the lifecycle callbacks. This allows for easier extending of base definitions or existing definitions.
- It's more natural to the language because there's no context rewriting going on.

The only possible caveat would be that using `this` feels cleaner and more intuitive (depending on how you look at it).

alexreardon:
I am a big fan of `this` being bound in skate called events. This was the state of affairs in an older spike that I am currently using. It makes life a lot nicer.

treshugart:
+1 to niceness. After our discussion, though, I'm still leaning toward passing an argument for the aforementioned reasons. I want to use `this`, but I'm not sure if it's worth sacrificing the natural behaviour of objects and flexibility that you get by being able to access the component definition. 

I had issues when using `this` in methods not on the `prototype` when using ES6 classes, too. When rewriting the context, you lose the inheritance that you get in staic methods.

bqx:
Another +1 to `this` being nice, but I think the benefits of passing the element as an argument might outweigh the niceness, such as being able to access the skate definition and also potentially inheritance.

treshugart:
I'm going to decline this PR. I'll raise one proposing the changing of `this` -> first argument.

PR #324, opened 2015-09-15
#281 - Implement change where update() is invoked before ready() but after created().

treshugart:


src/lifecycle/created.js
```diff
@@ -43,6 +43,18 @@ function callCreatedOnDescendants (elem, id) {
...
 }
 
+function createCallUpdateOnProperties (opts) {
+  let props = opts.properties || {};
+  let names = Object.keys(props);
+  return function (elem) {
+    names.forEach(function (name) {
+      let prop = props[name];
+      let update = prop && prop.update;
+      update && update.call(elem, elem[name]);
```
bqx: could you combine these two lines?
`prop && prop.update && prop.update.call(elem, elem[name]);`

src/lifecycle/created.js
```diff
@@ -51,6 +63,7 @@ function markAsResolved (elem, resolvedAttribute, unresolvedAttribute) {
 export default function (opts) {
   let created = opts.created;
   let isNative = opts.isNative;
+  let callUpdateOnProperties = createCallUpdateOnProperties(opts);
   let prototype = applyPrototype(opts.prototype);
```
bqx: it would be good for these function names to also be more descriptive, maybe something like
`let applyPrototype = createApplyPrototype(opts.prototype);` ?

src/lifecycle/created.js
treshugart: You could. I guess I'm just used to IE being slow on property access.

bqx: Either way is fine :)

bqx:
lgtm :+1:

src/lifecycle/created.js
treshugart: Fixed.

PR #334, opened 2015-09-19
#281 - Finalise property options.

treshugart:
### Property Changes

The property API has been reworked so that it can eventually be applied to the `prototype` rather than the instance which isn't possible now because of a Webkit bug as explained later in the PR. The API has also been reworked to be more descriptive and allow greater flexibility when creating re-usable properties. It supports a superset of options on top of the native options so that it works into the Skate lifecycle:

- `created` called after the main `created` lifecycle to set up the property. Good for setting initial values if overriding an existing property without having to invoke the setter which may cause mutations to happen before the descendant tree is initialised.

Having a hook to initialise the property for the element is necessary because:

- overriding of existing properties
- setting up the property based on information from the element
- the ability to delay the triggering of a callback when the property is set in case it has mutations and the descendant DOM must be initialised

This wouldn't be necessary if we were to define the property at initialisation time while passing in the element but this has performance and long-term API implications:

- you end up running the property setup function for every single property of every single instance
- it at least can run and return the definition that will be applied to every single property of every single instance, so it's saving a little bit of memory and cpu cycles
- it sets us up to be able to define properties on the prototype once, instead of for every single instance once [this 5 year-old bug](https://bugs.webkit.org/show_bug.cgi?id=49739) is fixed and we support a version of Webkit that it lands in. Every other browser supports accessing of native property getters / setters
- even if we decided to set aside the performance concerns, it would lock us into a property API that would prevent us from later using the prototype

### Added Properties

- `boolean` - `true` adds a void attribute, `false` removes the attribute
- `enum` (wip from @nicolasv)
- `float` - ensures the attribute value is a float
- `number` - ensures the value is a number
- `string` - ensures the value is a string

Currently these simply coerce values from the user supplied value to the property value and automatically `serialize` and `deserialize` values to and from attributes. In these definitions, `attribute` defaults to `true` but can be overridden to supply a custom attribute name or to turn off linking. It's on by default because the most common use-case when creating your public API is to link attributes so that you can declaratively render your web components rather than having to imperatively update properties.

### Render Changes

The following has been done to help accommodate server-side rendering. It's recommended this be done with property finalisation because how we finalise properties will affect server-side rendering and vice-versa.

- Moving responsibility of `render` to `renderer` that is responsible for rendering the "shadow" DOM.
- Re-purpose `render` so that it's sole responsibility is to return a rendered "shadow" template.
- `renderer` defaults to setting `innerHTML` but does not respond to changes on the element. This helps to strike a balance between convenience and opinion in that it will stamp out your template, but doesn't make assumptions as to how you want to respond to changes. There will be further work on render implementations for different types of templating engines and methods at a later point as separate plugins.


src/api/property.js
```diff
@@ -0,0 +1,133 @@
...
+    info.removeAttribute = elem.removeAttribute;
+    info.setAttribute = elem.setAttribute;
+    info.updatingProperty = false;
+
+    // TODO Refactor
+    if (info.linkedAttribute) {
+      if (!info.attributeMap) {
+        info.attributeMap = {};
+        info.removeAttribute = elem.removeAttribute;
+        info.setAttribute = elem.setAttribute;
```
nicolasv: These two (`info.removeAttribute` and `info.setAttribute`) are already defined above.

treshugart: It's a bit unclear here; this is kinda why it needs a refactor. The redefinitions here happen only if we are linking an attribute and no map exists yet (which we check to ensure we only run that code once). If we are linking, we must override the default attribute methods to update the property, which is why they get redefined here.

nicolasv: Well in this method (`prop.created`) one pair of these definitions is redundant.

treshugart: I see what you mean now. I didn't read the second one, I was just going by memory and I remember doing that for a reason that is probably no longer valid. I'll remove and run the tests.

src/lifecycle/created.js
```diff
@@ -1,93 +1,59 @@
...
+function ensurePropertyFunctions (opts) {
+  let props = opts.properties;
+  let names = Object.keys(props || {});
+  return names.reduce(function (prev, curr) {
+    prev[curr] = opts.properties[curr];
+    if (typeof prev[curr] !== 'function') {
+      prev[curr] = apiProperty(prev[curr]);
+    }
+    return prev;
+  }, {});
```
nicolasv: `prev` and `curr` as parameter names are confusing here: why not just `obj`/`descriptor` and `name`?

treshugart: That's a good idea. When I mock out a reduce I use those to start with. Forgot to refactor here.

alexreardon:
is `attr` still a thing?

treshugart:
@alexreardon yeah, currently. The proposal is to rename it to something a little bit more descriptive.

treshugart:
So this PR now brings things all into alignment. We have better names and the ability to serialise data between attributes as well as having a hook for checking property types.

src/lifecycle/property.js
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    let changeData = {
+      name: name,
+      newValue: newValue,
+      oldValue: oldValue
+    };
+
+    if (opts.set) {
+      opts.set(this, changeData);
```
nicolasv: Does having quite a different signature for the descriptor vs. options `set` mean it'll be harder to move to a world where the skate API would allow passing the descriptors directly rather than passing in the options?

treshugart: It shouldn't. Internally, Skate will just move the `Object.defineProperty()` call to the `prototype` rather than doing it on the element instance itself. The public definition remains the same.

treshugart: This is a good point of discussion, though. Are we being too radical here in passing the `element` and `changeData` as arguments instead of rewriting the context? I've a big fan of not rewriting contexts if we can help it. This also keeps it in convention with the old `attributes` API and with the current convention of passing the element as the first argument to non-prototyped methods. Maybe there's a good argument against this?

nicolasv: I think it's fine for the element to be the first argument and to not set `this`.

In my initial question in this thread I overlooked the fact that if you want to pass descriptors rather than opt-in to the functionality of this file you just need to provide a function for the property definition in skate (rather than an object with `default`, `emit`, etc.), so it really isn't a problem that `set` in those two approaches has a different signature.

test/unit/api/property.js
```diff
@@ -0,0 +1,93 @@
...
+    expect(skate.property.float).to.be.a('function');
+    expect(skate.property.number).to.be.a('function');
+    expect(skate.property.string).to.be.a('function');
+  });
+
+  describe('boolean', function () {
+    it('serialize and deserialize', function () {
+      let elem = create(property.boolean());
+
+      // Initial values.
```
nicolasv: I think these comments would be better served as the assertion message to `to.equal(…)`.

treshugart: Very good point.

README.md
```diff
@@ -263,47 +261,45 @@ skate('my-element', {
...
+
+      // This will be used as the default value for the property. If you specify
+      // a function then it will be invoked and the return value will be used.
+      // This option will also be used in place of the value returned from the
+      // `get()` option if it returns `undefined`.
+      default: 'initial value'
 
       // Whether or not to trigger events when the property changes. Defaults to
-      // `true`. If you do not want events triggered, set this to a falsy value.
+      // `false`. If you do not want events triggered, set this to a falsy value.
```
nicolasv: > If you do not want events triggered, set this to a falsy value.

I guess should talk about if you want to enable it, since the default has been inverted.

README.md
```diff
@@ -263,47 +261,45 @@ skate('my-element', {
...
       //
       // If this is truthy, when the property is changed it will trigger an
-      // event called `skate.property` if `notify` is `true`, or an event with
-      // the same name as `notify` if specified as a `String`.
+      // event called `skate.property` if it is `true`, or an event with
```
nicolasv: This isn't saying the literal string `skate.property` but `skate.your-property-name-here`, right? Might be worth clarifying.

README.md
```diff
@@ -311,14 +307,11 @@ skate('my-element', {
       // a custom value, specify a getter.
       set: function (newValue, oldValue) {},
```
nicolasv: This doesn't match the signature of the call to `opts.set(this, changeData)` in `src/lifecycle/property.js`.

README.md
```diff
@@ -311,14 +307,11 @@ skate('my-element', {
...
-      // and passed as `newValue` in the setter.
-      type: Boolean,
-
-      // This will be used as the initial value for the property. If you specify
-      // a function then it will be invoked and the return value will be used.
-      value: 'initial value'
+      // A function that gets called before `set()`. It's up to you what you do
+      // here. You can log, warn, or throw an exception if an unacceptable value
+      // is detected. By default this does nothing and the return value is not
+      // used.
```
nicolasv: > // … By default this does nothing and the return value is not
> // used.

This is no longer accurate, right? `src/lifecycle/property.js` sets `newValue` to the result of this if you've provided it.

README.md
```diff
@@ -263,47 +261,45 @@ skate('my-element', {
...
+      // Deserializes the attribute value when it is set so that it can be
+      // set as the property value. This only is called if this property is
+      // linked to an attribute.
+      deserializeFromAttribute: function () {},
+
+      // This will be used as the default value for the property. If you specify
+      // a function then it will be invoked and the return value will be used.
+      // This option will also be used in place of the value returned from the
+      // `get()` option if it returns `undefined`.
+      default: 'initial value'
```
nicolasv: I'm not sure that `'initial value'` is a good example to use now that the option is named `default`.

README.md
```diff
@@ -263,47 +261,45 @@ skate('my-element', {
...
-        //
-        // At the time the dependency event bubbles up from the descendant, it
-        // must be accessible via `this.some.nested.child`. If it is not, then
-        // the dependency is ignored.
-        'some.nested.child.deepDependencyProperty'
-      ],
+      // Serializes the property value when it is set so that it can be stored
+      // as an attribute value. This only is called if this property is linked
+      // to an attribute.
+      serializeToAttribute: function () {},
```
nicolasv: This option is named `serialize` in the implementation below (and ditto for `deserializeFromAttribute`).

src/lifecycle/property.js
```diff
@@ -0,0 +1,149 @@
+import dashCase from '../util/dash-case';
+import data from '../util/data';
+import emit from '../api/emit';
+
+// TODO Decouple boolean attributes from the Boolean function.
```
nicolasv: This TODO is complete, right?

src/api/property/float.js
```diff
@@ -0,0 +1,6 @@
+export default {
+  default: 0,
+  serialize: String,
+  deserialize: parseFloat,
```
nicolasv: For these (float, number, string) `deserialize` methods, should they behave the same as for boolean by returning `undefined` when they equal the default value so that the attribute is removed?

src/lifecycle/attribute.js
```diff
@@ -1,25 +1,10 @@
...
   return function (name, oldValue, newValue) {
-    let info = data(this);
-    let attributeToPropertyMap = info.attributeToPropertyMap || {};
-
     callback(this, {
       name: name,
-      newValue: newValue,
-      oldValue: oldValue
+      newValue: newValue === null ? undefined : newValue,
+      oldValue: oldValue === null ? undefined : oldValue
```
nicolasv: Why is `undefined` now the value for absence of the attribute? Is for consistency with `removeAttribute(…)` being called if `serialize` returns `undefined`?

src/lifecycle/properties-ready.js
```diff
@@ -0,0 +1,6 @@
+export default function propertiesApply (elem, props) {
+  Object.keys(props).forEach(function (name) {
+    let prop = props[name];
+    prop.set && prop.set.call(elem, elem[name]);
```
nicolasv: `prop.set` will always be defined, right? That is, this isn't the `set` *option*.

src/lifecycle/property.js
```diff
@@ -0,0 +1,149 @@
...
+    // TODO Refactor
+    if (info.linkedAttribute) {
+      if (!info.attributeMap) {
+        info.attributeMap = {};
+
+        elem.removeAttribute = function (attrName) {
+          info.updatingAttribute = true;
+          info.removeAttribute.call(this, attrName);
+          if (attrName in info.attributeMap) {
+            elem[info.attributeMap[attrName]] = undefined;
```
nicolasv: Extracting `info.attributeMap[attrName]` into a `propertyName` variable would make this code (and below) a bit easier to read.

README.md
```diff
@@ -263,47 +261,45 @@ skate('my-element', {
...
-      //
-      // To make a property "readonly", specify a getter without a setter.
-      get: function () {},
+      // Deserializes the attribute value when it is set so that it can be
+      // set as the property value. This only is called if this property is
+      // linked to an attribute.
+      deserializeFromAttribute: function () {},
+
+      // This will be used as the default value for the property. If you specify
+      // a function then it will be invoked and the return value will be used.
```
nicolasv: Is it worth mentioning that the function is invoked every time the getter returns `undefined`? Or, alternatively, should the implementation cache the return value the first time the function is invoked?

src/lifecycle/property.js
```diff
@@ -0,0 +1,149 @@
...
+          info.updatingAttribute = false;
+        };
+
+        elem.setAttribute = function (attrName, attrValue) {
+          info.updatingAttribute = true;
+          info.setAttribute.call(this, attrName, attrValue);
+          if (attrName in info.attributeMap) {
+            // Could also call getAttribute() but this does the same thing.
+            attrValue = String(attrValue);
+            elem[info.attributeMap[attrName]] = opts.deserialize ? opts.deserialize(attrValue) : attrValue;
```
nicolasv: Perhaps `opts` should have a default value for `deserialize` and `serialize` that is just the identity function so that you don't have to check if it was specified and also so you don't forget call it anywhere (since you would just always call it)?

lib/lifecycle/property.js
```diff
@@ -0,0 +1,169 @@
...
+        newValue: newValue,
+        oldValue: oldValue
+      };
+
+      if (opts.set) {
+        opts.set(this, changeData);
+      }
+
+      if (opts.emit) {
+        var eventName = opts.emit;
```
nicolasv: If `opts.emit !== true` then should we `String(opts.emit)` to get the event name?

README.md
treshugart: I've removed that line. It does in fact default to `false`.

README.md
treshugart: I've updated.

README.md
treshugart: Nice catch. Admittedly I didn't spend enough time updating docs.

README.md
treshugart: That's correct. Updated.

README.md
treshugart: Renamed this back to `serialize`. I like the verbosity of the former. What do you think?

src/lifecycle/property.js
treshugart: Yep. Will remove.

src/api/property/float.js
treshugart: I was unsure of that when I was spiking them. I think that's probably a good idea. What are your thoughts?

src/lifecycle/attribute.js
treshugart: Yeah, but moreover it's for consistency across the board. We return `undefined` for empty values instead of null.

src/lifecycle/properties-ready.js
treshugart: Not necessarily. If you were to define your own property function that returns a definition, you don't have to define `set`. This is just defensive of that possibility.

src/lifecycle/property.js
treshugart: I like that idea.

lib/lifecycle/property.js
treshugart: Yeah, that's probably a good idea to be defensive.

treshugart:
I created an issue to address the feedback for this issue: https://github.com/skatejs/skatejs/issues/341.

src/api/property/float.js
treshugart: As discussed IRL, I'll just keep these as is for now. The only difference is that once an attribute is set, it will always exist on the element, even if it is the default. If no default is specified and `undefined` is passed, then the attribute *should* be removed.

README.md
treshugart: Fixed in dc1ab32.

src/lifecycle/property.js
treshugart: Done in ec7fc2d.

lib/lifecycle/property.js
treshugart: Actually, it's currently API to be able to specify an `array` to emit several events so stringifying it breaks tests (had to rewrite the test to ensure this doesn't break). Do you have any arguments against using an array?

Issue #335, opened 2015-09-22
Go back to using an argument instead of "this" in the lifecycle callbacks.


treshugart:
We spiked using `this` as the element in the lifecycle callbacks for awhile. While it felt nice, it had its drawbacks:

- The lifecycle callbacks aren't actually defined on the prototype. This means that `this` actually refers to the component definition rather than the element. Rewriting the context means that you can no longer access the component definition which makes extending more difficult.
- Since you're applying these functions to the function definition, by nature, the context is not the element. Rewriting this could have unexpected results.
- It's a breaking change from 0.13. While we're still in dev releases, this is still something to consider.
- When using ES6 classes, it seems to break inheritance when rewriting contexts.

PR #336, opened 2015-09-22
#335 - Go back to using an argument instead of 'this'

treshugart:
This implements #335. Changes:

- All functions *not* on the prototype now receive an argument which is the element instead of using `this`. The function is always the first argument.
- The `attribute` callback now receives an object literal as its second argument that contains what was previously the `name`, `newValue` and `oldValue` arguments. This does not conform to native custom elements, however, it's a better design choice for an API because you do not have to remember argument ordering, only names.
- The `properties` callback `update()` now follows the same convention as the `attribute` callback.

### DIFF
--- a/src/lifecycle/created.js
+++ b/src/lifecycle/created.js
@@ -11,12 +11,12 @@ let elProto = window.Element.prototype;
 let oldSetAttribute = elProto.setAttribute;
 let oldRemoveAttribute = elProto.removeAttribute;
 
-function applyPrototype (proto) {
-  let prototypes = protos(proto);
-  return function () {
+function applyPrototype (opts) {
+  let prototypes = protos(opts.prototype);
+  return function (elem) {
     prototypes.forEach(proto => {
-      if (!proto.isPrototypeOf(this)) {
-        assignSafe(this, proto);
+      if (!proto.isPrototypeOf(elem)) {
+        assignSafe(elem, proto);
       }
     });
   };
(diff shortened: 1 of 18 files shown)


src/lifecycle/created.js
```diff
@@ -55,21 +56,25 @@ function createCallUpdateOnProperties (opts) {
       if (prop && prop.type) {
         val = (prop.type === Boolean && elem.hasAttribute(typeof prop.attr === 'string' ? prop.attr : name)) || prop.type(val);
       }
-      update && update.call(elem, val);
+      update && update(elem, {
+        name: name,
+        newValue: val,
+        oldValue: null
```
bqx: For a newly created element, shouldn't `oldValue` be `undefined`?

src/lifecycle/created.js
```diff
@@ -80,14 +85,14 @@ export default function (opts) {
...
+    properties(this, opts);
+    events(this, opts);
     render(this, opts);
-    callCreatedOnDescendants(this, opts.id);
+    callCreatedOnDescendants(this, opts);
     callUpdateOnProperties(this);
-    opts.ready && ready.call(this);
-    isResolved || markAsResolved(this, opts.resolvedAttribute, opts.unresolvedAttribute);
+    ready && ready(this);
+    isResolved || markAsResolved(this, opts);
```
bqx: Nice, this is much cleaner.

src/lifecycle/created.js
treshugart: Yeah I'll fix that. Writing some unit tests for this stuff specifically right now.

bqx:
> The attribute callback now receives an object literal as its second argument that contains what was previously the name, newValue and oldValue arguments.

This is nice API-wise. Do we want skate to be a polyfill that follows the spec closely or is it more of a library that provides a wrapping API and extra features? If its the former I'd hesitate to make this change, but if its the latter then this is great.

treshugart:
@bqx not sure why it can't be considered both. It is a wrapper API, but it also closely follows the spec. It may not implement the same exact API as the spec, but for convenience, that's a good thing. So long as it doesn't stray too far from it.

Issue #302, opened 2015-09-01
Switch currentTarget and delegateTarget so that the event object is similar to jQuery's


treshugart:
When originally naming this, I named it to be similar to jQuery, however, jQuery goes against the grain and rewrites `currentTarget` to mean what we are calling `delegateTarget`. I'd like to avoid this confusion and give it a name that's a little bit more appropriate.

a-ignatov-parc:
Why not to use `e.currentTarget`? Most of the developers are familiar with this name and will be easier to understand what it is for without digging docs

treshugart:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget):

> "It always refers to the element the event handler has been attached to..."

For a comparison of native vs jQuery see [this bin](http://jsbin.com/litupa/5).

> Most of the developers are familiar with this name and will be easier to understand what it is for without digging docs.

I presume what you mean by "most of the developers" are people who use jQuery. As someone who has always preferred native to jQuery, I find the fact that jQuery overwrites `currentTarget` to be the element that matched the selector odd since it contradicts what the spec says `currentTarget` should be.

a-ignatov-parc:
Hmmm... very interesting. But jquery doc says:

> The current DOM element within the event bubbling phase.

Which is exactly (almost) what MDN says:

> Identifies the current target for the event, as the event traverses the DOM.

Also if you change delegate event handlers to normal `currentTarget` begin to work properly http://jsfiddle.net/80g1woco/1/

a-ignatov-parc:
Ahh sorry. i misunderstood this issue.

a-ignatov-parc:
w3c spec does not provide api to delegate events. So i think it is normal for jquery (and skatejs) to mask event delegation as direct bindings through `currentTarget`.

treshugart:
That's actually a pretty good argument for using `currentTarget`. It's worth considering anyway.

marcoms:
I suppose `currentTarget` could be retrieved thought `this`, no? Even so, overriding a standardised property I'm not so comfortable with. I think it could be quite confusing to those expecting standard behaviour without reading the docs.

treshugart:
After thinking about this I think @a-ignatov-parc is probably right. I think it's a good design choice to make it so the developer doesn't have to think about whether or not they're delegating. If you want to get the host element you can use `this`.

PR #343, opened 2015-10-01
#302 - Make currentTarget and delegateTarget consistent with jQuery event objects.

treshugart:
Implements #302.

### DIFF
--- a/src/lifecycle/events.js
+++ b/src/lifecycle/events.js
@@ -1,5 +1,14 @@
 import matches from '../util/matches-selector';
 
+function readonly (obj, prop, val) {
+  Object.defineProperty(obj, prop, {
+    configurable: true,
+    get () {
+      return val;
+    }
+  });
+}
+
 function parseEvent (e) {
   let parts = e.split(' ');
   let name = parts.shift();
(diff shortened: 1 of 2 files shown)


src/lifecycle/events.js
```diff
@@ -1,5 +1,14 @@
 import matches from '../util/matches-selector';
 
+function readonly (obj, prop, val) {
+  Object.defineProperty(obj, prop, {
+    configurable: true,
```
bqx: Does a read only property need `configurable: true`?

src/lifecycle/events.js
```diff
@@ -16,7 +25,8 @@ function makeDelegateHandler (elem, handler, parsed) {
     let selector = parsed.selector;
     while (current && current !== elem.parentNode) {
       if (matches(current, selector)) {
-        e.delegateTarget = current;
+        readonly(e, 'currentTarget', current);
+        readonly(e, 'delegateTarget', elem);
```
bqx: Is there any particular reason that this uses `Object.defineProperty` instead of just `e.delegateTarget = elem`?

bqx:
lgtm :+1:

src/lifecycle/events.js
treshugart: Yeah, it's a readonly property, but still `configurable` so in order to override the value we must re-define it.

src/lifecycle/events.js
treshugart: Yeah, if you plan on redefining it. This is true if you reuse an event object and it passes through the same lifecycle. Since you can override it in native, we should probably try and stick as closely as possible to that.

Issue #351, opened 2015-10-08
Remove sync init of descendants from a host component.


treshugart:
This was done in order to try and bring some semblance of order to now native custom elements behave, which is all over the place. However, if you follow some simple best-practices, it's possible to avoid it altogether so we're going to remove this and eventually form some documentation around the best way to build components using different techniques. Removing this hardly imposes any limits on the user, it just means that you can't querySelector from inside a host and expect a descendant element to be initialised yet.

marcoms:
"querySelector from inside a host" - does this extend to all stages of the host's lifecycle? Will `skate.ready` be reintroduced?

treshugart:
Currently, without forcing init on descendants, you can still find them and set properties / attributes on them and they will get the changes no matter what. You just can't do stuff with their eventual prototype because they may or may not have been upgraded yet. This was always the case. To this effect, not much is actually changing, though we should provide something like `skate.ready()` to call something when an element is eventually ready so that in edge-cases you can call prototyped stuff on an element.

treshugart:
@marcoms implemented in #352.

marcoms:
But doesn't the revised lifecycle order described in #285 push property initialisation after `created`? Components designed with this assumption might break if their property setters are called *before* `created`, no?

treshugart:
If you're only interacting with the descendant custom element and not its
descendants (which we should never be doing, that's the responsibility of
the descendant host) then it'll work exactly as before. The prototype bug
existed before, the removal of this.

PR #352, opened 2015-10-08
#351 - Remove sync init of descendants from a host component.

treshugart:


src/api/ready.js
```diff
@@ -0,0 +1,54 @@
+import data from '../util/data';
+import registry from '../global/registry';
+
+function ready (element) {
+  const components = registry.find(element);
+  for (let a = 0; a < components.length; a++) {
```
joscha: New browsers get better at this, but `let a = 0, l = components.length; a < l; a++) {`?

src/api/ready.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export default function (elements, callback) {
+  // If a selector is provided we convert it to to a collection of elements.
+  if (typeof elements === 'string') {
+    elements = document.querySelectorAll(elements);
+  }
+
+  // Ensure the collection is traversable.
+  const collection = elements.length === undefined ? [elements] : elements;
```
joscha: `typeof elements.length === 'undefined'`

src/api/ready.js
```diff
@@ -0,0 +1,54 @@
...
+  const collectionLength = collection.length;
+
+  // If there's no elements we don't do anything.
+  if (!collection.length) {
```
joscha: reuse `collectionLength`?

joscha: `callback` is not supposed to be invoked with an empty element list in this case?

test/unit/api/ready.js
```diff
@@ -0,0 +1,74 @@
...
+
+  it('should fire for an element if no definitions are registered for it', function (done) {
+    ready(elem, function (shouldBeElem) {
+      expect(arguments.length).to.equal(1);
+      expect(shouldBeElem).to.equal(elem);
+      done();
+    });
+  });
+
+  it('should fire for an element when it is already ready', function (done) {
```
bqx: It seems like this doesn't test the timing of the ready callback, it would be good to ensure the callback fires immediately once the element is ready. If possible doing something like this might be good:

    let readyCallbackFired = false;
    setup();
    ready(elem, function() {
       readyCallbackFired = true;
    });
    initialise();
    // wait a tick if we need to?
    expect(readyCallbackFired).to.equal(true);

src/api/ready.js
treshugart: I'd rather `const` it outside the loop, but yeah, may as well do that for older browsers :)

src/api/ready.js
treshugart: What's the benefit of that vs checking equality?

joscha: http://stackoverflow.com/a/3550319

src/api/ready.js
treshugart: There's no elements to execute anything on. Not sure if there's use-cases that contradict this, though. Ideas?

joscha: add an node-style error parameter to the callback that you populate in that case? But I think the API should be stable, e.g. if there is a callback passed, it should be called at some point in time. For no elements (e.g. all of the 0 elements are ready :-)) I'd have expected this to be called straight away I think.

test/unit/api/ready.js
treshugart: Done. Let me know if it fits with what you were looking for.

src/api/ready.js
```diff
@@ -0,0 +1,45 @@
...
+      elem.addEventListener('skate.ready', function () {
+        ++readyCount;
+        callbackIfReady();
+      });
+    }
+  }
+
+  // If the elements are all ready by this time that means nothing was ever
+  // bound to skate.ready above.
+  callbackIfReady();
```
joscha: :+1:

Issue #354, opened 2015-10-12
Public API for rendering / re-rendering a component.


treshugart:
This is useful if you are taking a more functional approach with your custom element and want to re-render when something changes. This wouldn't be very useful if you're rendering once and using setters to mutate the rendered DOM because this would just re-execute the render lifecycle on the element.

Proposed API:

```js
var hello = skate('x-hello', {
  render: function (props) {
    return `Hello, ${props.name || 'World'}!`;
  }
});

// This is already part of the API. This would create the component, set some
// props and render as part of the normal lifecycle.
// <x-hello>Hello, Wilson!</x-hello>
var hi = hello({ name: 'Wilson' });

// Sets a new value to be rendered with.
hi.name = 'Bobbo';

// This is the new API function that re-renders the element.
// <x-hello>Hello, Bobbo!</x-hello>
skate.render(elem);
```

alexreardon:
Very cool! However, would this be confusing as the standard behaviour is that arguments to the skate constructor set's props?

treshugart:
Yeah, that's standard behaviour. The only addition here is `skate.render()`.

alexreardon:
is `state` = `props`?

treshugart:
It's essentially the element instance, but the idea of the render function is to treat the incoming object as state / props (in Skate they're considered the same thing).

alexreardon:
For consistence I think using the word `props` would be best

```
var hello = skate('x-hello', {
  render: function (props) {
    return `Hello, ${props.name || 'World'}!`;
  }
});
```

I like the idea of a new render function that takes `props` as it's arguments. When would the `render` function be fired in the lifecycle? Does it open up the possibility of being confused with `template`?

treshugart:
@alexreardon https://github.com/skatejs/skatejs/issues/337. Added `render` and `renderer` in that. No more `template`.

alexreardon:
:+1:

alexreardon:
Looks very cool

bqx:
Nice. I actually thought you could do this already. 

Is it possible to refer to the `render` function from inside another lifecycle callback? (e.g. inside an event or property handler) or would you have to use `skate.render()`? i.e.:

```
skate('x-el', {
  render: function (el) {
    ..
  },
  events: {
    click: function (e) {
      // Do I have to use skate.render() instead of this.render() here since `this` is bound to the element?
      skate.render({ prop: 'value' });  
    }
  }
}
```

[...]

PR #355, opened 2015-10-12
#354 - Public API for rendering / re-rendering a component.

treshugart:
Implements #354.

### DIFF
--- a/src/api/fragment.js
+++ b/src/api/fragment.js
@@ -35,7 +35,8 @@ export default function (html) {
     let par = createFromHtml(html);
 
     if (par.nodeType !== 1) {
-      return par;
+      frag.appendChild(par);
+      return frag;
     }
 
     while (par.firstElementChild) {
--- a/src/api/fragment.js
+++ b/src/api/fragment.js
@@ -1,5 +1,5 @@
 import init from './init';
-import createFromHtml from '../util/create-from-html'
+import createFromHtml from '../util/create-from-html';
 
 const DocumentFragmentPrototype = DocumentFragment.prototype;
 const slice = Array.prototype.slice;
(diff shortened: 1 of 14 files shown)


src/api/fragment.js
```diff
@@ -29,10 +29,15 @@ function decorateFragmentMethods (frag) {
...
 
 export default function (html) {
-  var frag = document.createDocumentFragment();
+  const frag = document.createDocumentFragment();
   decorateFragmentMethods(frag);
   if (typeof html === 'string') {
-    var par = createFromHtml(html);
+    let par = createFromHtml(html);
+
+    if (par.nodeType !== 1) {
```
treshugart: Fragments never worked for text-nodes.

treshugart: @nicolasv what do you think about this? Before, we couldn't use `skate.fragment()` for templating unless it had a wrapper element. I'm not sure if we should do it this way or use `par.firstChild` instead of `par.firstElementChild` so that all node types are captured.

src/api/render.js
```diff
@@ -0,0 +1,9 @@
+import registry from '../global/registry';
```
treshugart: This is the meat of the PR. This just forces the render lifecycle on the element.

src/lifecycle/created.js
```diff
@@ -37,10 +37,15 @@ function notifyReady (elem) {
   });
 }
 
+function renderIfNotResolved (elem, opts) {
```
treshugart: Had to move the resolved checks to a function here. This used to be in the `renderer` lifecycle, but since we want to be able to re-use that and not have it check if it's resolved or not, we had to move this here.

src/lifecycle/renderer.js
```diff
@@ -1,20 +1,18 @@
...
     elem.removeChild(elem.childNodes[0]);
   }
-  elem.appendChild(fragment(render()));
-};
-
-export default function renderer (opts) {
-  let render = opts.render ? opts.render : null;
-  let renderer = opts.renderer ? opts.renderer.bind(opts) : defaultRenderer;
-  let resolvedAttribute = opts.resolvedAttribute;
+  elem.appendChild(fragment(opts.render(elem)));
```
treshugart: If you were to return only text from your `render()` method, then this default renderer would fail because `fragment()` didn't accept text nodes.

src/util/create-from-html.js
```diff
@@ -30,11 +30,14 @@ function resolveParent (tag, html) {
 }
 
 function matchTag (html) {
-  var tag = html.match(/\s*<([^\s>]+)/);
+  var tag = html.match(/^<([^\s>]+)/);
```
treshugart: Since we trim, we can make this regex more specific.

PR #357, opened 2015-10-12
#356 - skate.fragment() should take a DOM node.

treshugart:
Implements #356.


test/unit/api/fragment.js
```diff
@@ -119,4 +119,24 @@ describe('api/fragment', function () {
       expect(frag.childNodes[1].tagName).to.equal('TD');
     });
   });
+
+  describe('node', function () {
+    it('should work with element nodes', function () {
+      const node = document.createElement('div');
+      const frag = skate.fragment(node);
+      expect(frag.childNodes[0]).to.equal(node);
```
nicolasv: Although there are separate tests to make sure the overridden methods (`appendChild` in this case) actually `init` the element passed in, might be worth asserting that the `node` has been resovled here.

nicolasv:
LGTM

test/unit/api/fragment.js
treshugart: Ended up writing a separate test in 4b0a923 since that test is working with a plain `<div>`.

Issue #362, opened 2015-10-13
Add change() property callback that only gets fired when the property changes.


treshugart:
### Use case

This is especially handy when wanting to re-render using `skate.render()` when the property changes, but not wanting to do it whenever `set` is called because `set` is called when the element is initialised and every time it is set, even if it is the same value.

```js
skate('my-element', {
  properties: {
    textContent: {
      change: skate.render
    }
  },
  render (elem) {
    return `Hello, ${elem.textContent}!`;
  }
});
```

### Signature

It should receive the same arguments as `set` and be called `change`.

### Tests

- is not called if the property has no default value, when it is set up
- is not called if the property has a default value, when it is set up
- is not called if the property has a default value and it does not change
- is not called if the property value does not change
- is called if the property has no default value and it is set
- is called if the property has a default value and it is set to a different value
- is called if the property value changes

bqx:
It seems this is a convenience function only - nothing is stopping the user from doing the same check for `changeData.oldValue !== change.newValue` inside the the setter if they need this kind of behaviour. Is it really worth providing the convenience here?

joscha:
why is `is not called if the property has no default value`? It is still a change, or no?

joscha:
those two:
`is not called if the property has no default value`
and
`is not called if the property has a default value`
kinda sounds as if its never called?

treshugart:
@joscha Reworded. I'll update the test descriptions to suit.

treshugart:
Ended up removing the spiked callback.

PR #363, opened 2015-10-13
#362 - Add change() property callback that only gets fired when the property changes.

treshugart:
Implements #362.


test/unit/lifecycle/property.js
```diff
@@ -323,5 +323,107 @@ describe('lifecycle/property', function () {
...
+          change () {
+            ++calls;
+          }
+        });
+        expect(calls).to.equal(0);
+      });
+
+      it('is called if the property has no default value and it is set', function () {
+        let calls = 0;
+        let elem = create({
```
bqx: We could use `const` for declarations like this where we don't need to assign again

src/lifecycle/property.js
```diff
@@ -125,6 +125,10 @@ function createNativePropertyDefinition (name, opts) {
       opts.set(this, changeData);
     }
 
+    if (opts.change && oldValue !== newValue) {
```
joscha: `typeof opts.change === 'function'` is more sane

test/unit/lifecycle/property.js
```diff
@@ -323,5 +323,107 @@ describe('lifecycle/property', function () {
         create(opts);
       });
     });
+
+    describe('change()', function () {
+      it('is not called if the property has no default value', function () {
+        let calls = 0;
```
joscha: this looks like a pure [sinonJS](http://sinonjs.org/) use-case to me.

test/unit/lifecycle/property.js
```diff
@@ -323,5 +323,107 @@ describe('lifecycle/property', function () {
...
+      it('is not called if the property value does not change', function () {
+        let calls = 0;
+        let elem = create({
+          default: true,
+          change () {
+            ++calls;
+          }
+        });
+        elem.test = true;
+        elem.test = true;
```
joscha: copy & paste glitch?

treshugart: That's to test whether or not `change()` is called if the value being set again is the same.

joscha: O wow - observer power :)

Issue #348, opened 2015-10-06
Write server-side function for rendering / expanding an HTML string of web components.


treshugart:
Expands on #319.

We should be able to give a function a string that will parse the string as HTML and initialise each component that is found only going through a subset of the lifecycle and return a string of the rendered / expanded component HTML.

The lifecycle elements that will need to be called on a virtual element are:

- properties
- created
- render

Given the following definitions:

```js
skate('some-parent', {
  render: function (state) {
    return `
      <div>${state.innerHTML}</div>
    `;
  }
});

skate('some-child', {
  render: function (state) {
    return `<span>${state.textContent}</span>`;
  }
});
```

The following:

```js
skate.render(`
  <some-parent>
    <some-child>text content</some-child>
  </some-parent>
`);
```

Would render a string of:

```html
<some-parent>
  <div>
    <some-child>
      <span>text content</span>
    </some-child>
  </div>
</some-parent>
```

treshugart:
I don't think this is a responsibility of Skate. Server-side rendering really should be done on a per-use-case basis as Skate has no opinion on how to render the components. All Skate should provide is a way to skip the `render()` callback if a flag is provided.

PR #364, opened 2015-10-14
#348 - Server-side rendering.

treshugart:
This implements #348: full server-side rendering support for Skate components.

This implements a single function that takes input as HTML with unexpanded web components (so the same thing you'd expect to write on the client) and it will render everything out and return the resulting string.

The string can be output directly to the client from the server and Skate will not render the components on the client, but it will go through the rest of the lifecycle.

The supported parts of the lifecycle that gets called on the server is:

1. defines `properties`
2. calls `created`
3. readies `properties`
4. calls `render` (but not `renderer`) and passes in the `properties`.
5. adds the `resolved` attribute (this is what prevents it from rendering on the client)

In order to render components on the server-side, all you do is `require('skatejs')` and `require()` your component definitions. From there, all you do is call the node render function.

For example, the following:

```js
const render = require('skatejs/node');
const skate = require('skatejs');

skate('x-hello', {
  properties: {
    textContent: null
  },
  render (props) {
    return `Hello, ${props.textContent}!`;
  }
});

console.log(render('<x-hello>World</x-hello>'));
```

Would log:

```html
<x-hello>Hello, World!</x-hello>
```

Several parts of the codebase had to be updated to run in a node environment. This includes normalising the global object in a module and using that instead of `window` as well as ignoring certain parts of the `created` lifecycle when not running in a browser.


treshugart:
Declined because we're not quite ready to implement this. This was a spike to see if it's possible and roughly how.

PR #377, opened 2015-10-19
Fix mutation observer / lifecycle management on IE

joscha:
This fixes the lifecycle management on IE (e.g. incorrect calls to `fallback` instead of `created` and `removed`).

### DIFF
--- a/src/document-observer.js
+++ b/src/document-observer.js
@@ -5,7 +5,7 @@ import {
   initElements,
   removeElements
 } from './lifecycle';
-import MutationObserver from './mutation-observer';
+import './mutation-observer';
 import {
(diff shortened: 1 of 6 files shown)


treshugart:
Not tooooo much more green in this so it seems a lot has changed in their polyfill. Hopefully the performance tests check out. If so let's merge and recommend it's usage in master (soon to be 0.14)!

joscha:
I created a JSPerf here: http://jsperf.com/mutationobserver-perf
mutationobserver_ _jsperf

joscha:
I added a `childList` test to the jsPerf fiddle, however it seems as if the Skate polyfill does not correctly recognize all element changes.

PR #378, opened 2015-10-19
Backport saucelabs

joscha:
This enbles Saucelabs tests for the 0.13.x branch


joscha:
only two fails of 1198 tests - both in
```console
IE 9/10 (Windows 7 0.0.0) utils getClosestIgnoredElement() should not fail called on an element in an iframe FAILED
	TypeError: Unable to get property 'appendChild' of undefined or null reference
```

joscha:
:heart:

PR #387, opened 2015-10-20
Add lifecycle tests

joscha:
This ports the lifecycle tests from 0.13.x to master

### DIFF
--- a/test/unit/attributes.js
+++ b/test/unit/attributes.js
@@ -1,30 +1,48 @@
 import helperElement from '../lib/element';
+import helperReady from '../lib/ready';
 import skate from '../../src/index';
 
 describe('lifecycle/attribute', function () {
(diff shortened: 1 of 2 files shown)


treshugart:
I think the current tests in https://github.com/skatejs/skatejs/blob/master/test/unit/attributes.js already do a lot of what is here (except maybe the `resolved` attribute test). Have a look at those and see if there's some way we can merge them.

We should eventually move that file to `tests/unit/lifecycle/attribute.js`, I reckon. The tests in `tests/unit/lifecycle.js` probably need organising at some point, too.

joscha:
Cool, I'll merge them in tomorrow - the `sinon`-based tests also test that the amount of calls are correct - the simple one with overwriting the `data`-closure would not detect duplicate calls.

PR #389, opened 2015-10-20
perf: early exit in case we did not get a callback

joscha:
no need to create closure and process any attributes


treshugart:
I don't really have any problems with this, but was there a noticeable performance issue with it? Also, tests :(

joscha:
no, no, I just noticed when looking through the code - the tests are broken because I hadn't merged the fix for master, yet.

PR #392, opened 2015-10-21
Fail on vital browsers

joscha:
This makes our PR tests fail for vital browsers (Chrome and Firefox for now). As soon as we had all our tests green for all supported browsers once, we should just generally fail for any failing browser.


treshugart:
:+1:

PR #407, opened 2015-10-23
Remove change() property callback and make set() get called with unde…

treshugart:
…fined as the oldValue on the initial set.

The rationale behind this is that `set()` was always getting called with an `oldValue` that was the same as `newValue` if the attribute / property existed on the element which is undesirable in most circumstances because you can't do `oldValue !== newValue` and do something based on that for the initial call to `set()`.

The removal of `change()` just means that we're leaving the checking of the aforementioned up to the developer. To wrap this themselves they could just do something like: 

```js
prop: changeProperty({ change: doSomething });
```

Where `changeProperty` might be something like:

```js
function changeProperty (opts) {
  const oldSet = opts.set;
  opts.set = function (elem, data) {
    if (oldSet) {
      oldSet.call(opts, elem, data);
    }
    if (opts.change && data.newValue !== data.oldValue) {
      opts.change(elem, data);
    }
  }
  return opts;
}
```

Or just have a simple abstraction around it:

```js
function didPropertyChange (data) {
  return data.newValue !== data.oldValue;
}
```

I think adding in `change` may have been a mistake because the check is so simple and it was added to facilitate functional UI rendering which can easily be done as part of a separate abstraction for helping with that.


wbinnssmith:
as long as folks are okay with the removal of `change`, this is a :+1: from me.

PR #408, opened 2015-10-23
fix: MutationObserver for IE11 and innerHTML fixes

joscha:
1__npm_run_test-watch_--_--browsers_chrome__google_chrome_he__and_hipchat


joscha:
@treshugart will enable saucelabs breaking the build when a karma test fails...

PR #414, opened 2015-10-25
fix(test): perf tests

joscha:
This fixes the performance tests and runs them automatically in our travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ before_script:
 - bower install
 script:
 - npm run lint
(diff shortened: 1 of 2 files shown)


treshugart:
Noice.

joscha:
uh wait, I think the tests are not green...

joscha:
will look into the fix, seems to "works on my machine" :dancers:

joscha:
Fixed in https://github.com/skatejs/build/pull/2 and ce0360f.

PR #430, opened 2015-11-02
Make property / properties public api references consistent -> props.

treshugart:
This changes all public API instances in the following way:

- `skate.property` renamed to `skate.properties`
- `properties.type` renamed to `properties.coerce`
- `properties.attribute` now defaults to false


marcoms:
There is a little bit of an inconsistency in my view, as `properties` et al. are now abbreviated, but `attribute`, `prop.attribute`, `unresolvedAttribute`, and `resolvedAttribute` are not. As I understand it the full-length "attribute" name (for the property option at least) was chosen in #281, but then shouldn't we keep `properties` unabbreviated as well?

treshugart:
@marcoms just made it `properties` across the board. Will merge once the build is green, but if you've any further feedback let's discuss.

Issue #445, opened 2015-12-02
Add a created callback to the property definition.


treshugart:
The property definition should support a `created` callback that is called during the created callback phase. This would essentially be called during the `created` phase of the element lifecycle.

marcoms:
Is this to logically group code, that would be valid in `el.created`, but is more relevant to the property?

treshugart:
Yeah. The use case here is more of an edge-case but see https://bitbucket.org/atlassian/aui/src/8e311c60418a952b83fb5e8f12295b049ca79c1b/src/js/aui/internal/properties/content.js?fileviewer=file-view-default.

That module is a property that projects content into a specified content area node in your template. When the property is set up (during the create lifecycle), we need to get the references to the element childNodes to use for the first call to `set`. On all subsequent `set`s we use the passed in value (`data.newValue`).

We can't do all of this work in the `set` callback because that is invoked *after* the `render()` callback is called meaning that you'll lose the references to the initial DOM structure that existed when the element is `created`. We could put that code in the `created` callback, but then that means we can't modularise all of the code for this property into a single spot. We'd have to separate it between the property definition and the `created` callback.

PR #446, opened 2015-12-02
#445 - Add a created() callback to the property definition.

treshugart:


cyberdash:
Could you achieve the same thing by checking `data.oldValue === undefined` in the property set callback?

```
set function (el, data) {
   if (data.oldValue === undefined) {
        //created
    }
}
```

treshugart:
Unfortunately not because the `set` callback is fired after `render()` is invoked. That would also fire if the property is ever set back to `undefined`.

cyberdash:
I think this is a fair addition the API, if you want to make standalone property definitions that can be shared, we need to be able to hook in to the "created" part of the lifecycle.

:+1:

PR #450, opened 2015-12-12
Missing default of `render` 

antitoxic:
My first code-contribution to skate.

The missing default is actually preventing copying over `render` from user options when they are passed as ES6 class.

Another thing I noticed is that options are treated as enumerable but they can also be a class which is not enumerable. I only fixed the defaults here. Issue should be opened for the enumerable problem.

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -11,6 +11,10 @@ export default {
   // called on them.
   created: function () {},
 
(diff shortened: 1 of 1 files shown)


treshugart:
Nice, thanks! Also, re the enumerable stuff, sounds good.

Issue #346, opened 2015-10-03
api/property.js still being referenced


marcoms:
Tried building a project today with skate@master and it breaks trying to reference src/api/property.js (recently split into modular chunks, as I'm sure you're aware). Looks like https://github.com/skatejs/skatejs/blob/master/src/index.js#L5 is the offender.

treshugart:
That's probably because we're using node's `require()` semantics to load the `api/property/index.js` file via `api/property`. I'll update this to load all the files directly.

treshugart:
Fixed in 6e56e3435c42ab1519cec63a692a15e622d6f367. Let me know if this is still happening.

marcoms:
all good :+1:

PR #452, opened 2015-12-19
Content property

treshugart:
This is a PR spiking a new property called `content`. This property will allow consumers to define one or more properties of their choice that act as content areas, or nodes that will project content to parts of their template.

It is designed to work with multiple rendering techniques.

### Initial render, mutate after

```js
const MyEl = skate('my-el', {
  properties: {
    content: skate.properties.content({
      default: '<p>Initial content.</p>',
      selector: '.content' 
    }),
    title: skate.properties.content({
      default: 'Initial title',
      selector: 'h1' 
    });
  },
  render: skate.render.html(function (elem) {
    return `
      <h1></h1>
      <section class="content"></section>
    `;
  });
});

const el = MyEl({
  title: 'Updated title',
  content: '<p>Updated content</p>'
});
```

### Functional (skatejs-dom-diff style)

```js
const MyEl = skate('my-el', {
  properties: {
    content: skate.properties.content({ change: skate.render }),
    title: skate.properties.content({ change: skate.render });
  },
  render: skateDomDiff.render(function (elem) {
    return (
      <h1>{elem.title.childNodes || 'Initial title'}</h1>
      <section class="content">
        {elem.content.childNodes || <p>Initial content</p>}
      </section>
    );
  });
});

const el = MyEl({
  title: 'Updated title',
  content: '<p>Updated content</p>'
});
```

### Same public API, different implementation

Both examples have the same declarative and imperative APIs, they're just handled differently internally, especially in more complex examples. The content property would also allow you to call certain methods on it. The initial API would be something that could be diffed and patched by something like [skatejs-dom-diff](https://github.com/skatejs/dom-diff). This means that it'd need to implement at least:

#### Properties

- `childNodes` (as an array)

#### Methods

The following would also need to support document fragments for whatever the native variants support.

- `appendChild(newNode)`
- `insertBefore(newNode, referenceNode)`
- `removeChild(oldNode)`
- `replaceChild(newNode, oldNode)`

This also means that not only can you `set` the properties directly, but you can use these methods and properties as you normally would.

### Declarative API

Both rendering methods should allow the same declarative API. However the example above only gives you an API for manually adding nodes. For example:

```html
<my-component>
  <h1>Title</h1>
  <p>Content.</p>
</my-component>
```

Would render:

```html
<my-component>
  <h1></h1>
  <section class="content">
    <h1>Title</h1>
    <p>Content.</p>
  </section>
</my-component>
```

When in fact what we really want is:

```html
<my-component>
  <h1>Title</h1>
  <section class="content">
    <p>Content.</p>
  </section>
</my-component>
```

In order to offer a declarative API where things can be distributed we'd need to be able to be given a blob of arbitrary elements and be able to decide where they should go. The [Shadow DOM named-slot API](https://w3c.github.io/webcomponents/spec/shadow/#h-slots) is [gaining vendor confidence](https://webkit.org/blog/4096/introducing-shadow-dom-api/) (Apple was probably the biggest critic of the old API) and is becoming more stable. It might be a good example to model this after it especially if it's something that can use native means underneath the hood in browsers that support it.

If we did, our declarative API becomes:

```html
<my-component>
  <h1 slot="title">Title</h1>
  <p>Content.</p>
</my-component>
```

Notice the `<p>` doesn't have a `slot` meaning it would be projected into the main content area (or the `content` property as we have it here).


marcoms:
Nice to have an official way to do one-way data-binding :)
BTW, why not make a call to skate.render automatically if no parameters are supplied, since looking at the functional style example, its beginning to look like an observer/observable system where you can attach an arbitrary function when the value changes

src/api/properties/content.js
```diff
@@ -0,0 +1,67 @@
...
+      return oldNode;
+    }
+  };
+}
+
+export default {
+  created (elem) {
+    const eldata = data(elem);
+    eldata.contentProperty = createDomArray(elem, update(elem, this.change));
+    eldata.contentPropertyInitiaState = [].slice.call(elem.childNodes);
```
marcoms: `initiaState` - typo?

src/index.js
```diff
@@ -2,7 +2,7 @@ import apiCreate from './api/create';
 import apiEmit from './api/emit';
 import apiFragment from './api/fragment';
 import apiInit from './api/init';
-import apiProperties from './api/properties/index';
```
marcoms: I use Rollup now instead of JSPM where #346 was giving me issues, but I think removing the explicit `index` would trigger it again for JSPM users (the SystemJS loader doesn't support the implied `index.js` import)

src/api/properties/content.js
treshugart: Totes. Fixed.

src/index.js
treshugart: Yeah I did this and then realised that it was because I was including stuff wrong. Fixed.

treshugart:
@marcoms I wouldn't necessarily call it "data-binding". It's more or less a way to project content without having to know the implementation details of the component. It's much like an imperative version of the Shadow DOM named-slot API, which we could also mimic declaratively if we wanted to. We'd just have to create function that distributes content, which can easily be done as an override to `innerHTML`. Maybe it's worth spiking this as well. This would allow a user to not only have a single content area that they can put stuff in, but multiple ones. I'll write this into the PR description.

antitoxic:
This looks interesting to me but the examples confuse me a bit. 

First and upmost - it reminds me of a problem when trying to do ShadowDOM-like/shade-like rendering. It's the problem with `[resolved]` and `[unresolved]` that are meant to prevent FOUC.

First and upmost, if this:
```
<my-component>
  <h1>Title</h1>
  <p>Content.</p>
</my-component>
```
was :
```
<my-component>
  <my-other-custom-component>
    <my-third-custom-component>
    Title
    </my-third-custom-component>
  </my-other-custom-component>
  <p>Content.</p>
</my-component>
```
and selectors were changed accordingly, how would `resolved/unresolved` attributes know when to be applied. As far as I know, skate doesn't track descendants being rendered? I have this problem now with a script of mine. `resolved` is applied before nested elements are rendered and a lot of FOUC happens and baiscally `resolved` attributes are kind of useless.

Second question: Is this react:

```
      <h1>{elem.title.value || 'Initial title'}</h1>
      <section class="content">
        {elem.content.value || <p>Initial content</p>}
      </section>
```
? How is this and skate combined? I mean is React needed for this to work? Also what will happen if my example with nested custom elements was used? Would skate-dom-diff detect nested custom-elements, render them and compare the rendered html?

marcoms:
Re: the revised OP

Well this is a lot of new... Content... For me to take in with this API, but I have a question with this example:

```html
<my-component>
  <h1>Title</h1>
  <p>Content.</p>
</my-component>

Would render:

<my-component>
  <h1></h1>
  <section class="content">
    <h1>Title</h1>
    <p>Content.</p>
  </section>
</my-component>
```

Is this an example of the imperative API? How is this done without appending to the `el.content` property? I've looked at the first two examples, and I can't really understand this in particular.

treshugart:
@antitoxic

> As far as I know, skate doesn't track descendants being rendered?

That is correct. This is actually terrible for performance as you have to crawl the entire DOM tree checking every element.

> `resolved` is applied before nested elements are rendered and a lot of FOUC happens and baiscally `resolved` attributes are kind of useless.

I tend to disagree. If the parent element knows about the descendants it is interacting with, you can use `skate.ready()` to know when select elements are ready to be shown. That is more for interacting from parent -> child via API, though. In this situation, you should be handling descendant DOM `resolved` styling on a per-component basis. There are too many use-cases here to be solved with a bit of magic code, at least in a reasonably performant manner.

You can do something like:

```css
x-parent x-descendant:not([resolved]) {}
/* or (if you add the unresolved attribute) */
x-parent [unresolved] {}
```

The former works only for `x-descendant` and is selective at the CSS level. The latter works for any component, but is selective at the component level. Both would apply unresolved styling to the descendant regardless of the parent state, which is way better than trying to be too smart at the library level. Without knowing your specific use-case, it's hard to make specific recommendations.

> How is this and skate combined? I mean is React needed for this to work?

No, React is not required. This *is* using JSX, but that's a separate technology.

> Also what will happen if my example with nested custom elements was used? Would skate-dom-diff detect nested custom-elements, render them and compare the rendered html?

Depends on if you're diffing / patching *real* DOM or *virtual* DOM. If you diff real DOM, generally you'd do something like:

```js
skateDomDiff.merge({
  destination: skate.fragment(newRenderTree),
  source: currentRenderTree
});
```

The `skate.fragment()` call there would sync init all components and then do a comparison + patch afterward. The differ wouldn't detect components, it would only be concerned with the nodes passed to it, diffing and patching as necessary. This "just works" so long as both trees are in their initialised state, which `skate.fragment()` ensures.

On the other hand, if you're using a virtual tree, then you're comparing vtree to vtree, most likely (if you do real tree to vtree, while possible, it could get you into trouble). When you construct a virtual tree, the diff / patch algorithm only sees the virtual elements. If one of those virtual elements is a web component, the component won't be expanded until it is converted to a real element. This naturally creates an encapsulated diff / patch tree on a per-component basis making it possible to intermix rendering styles without running into issues.

When using virtual trees, you'd keep the last rendered tree in memory and compare the newly generated tree to the previous one. Both would be in unexpanded state. If the patcher says "update attribute-a on component-a", then all it does is change it. The logic is handled internally by the web component since the public API of web components are properties and light dom.

As a side note, if you're using a functional style rendering pipeline, then it's worth using the virtual DOM because you don't have the overhead of component initialisation happening simply for a diff algorithm to be executed.

treshugart:
@marcoms

> Well this is a lot of new... Content... For me to take in with this API

Well played :)

> Is this an example of the imperative API?

No, this is the declarative one; where you'd blat out the former blob of HTML and internally it would be converted into something like the latter.

> How is this done without appending to the el.content property?

[Looking at the code](https://github.com/skatejs/skatejs/pull/452/files#diff-720334ddfa070115462ad0888090a585R53), the `created` callback for a property gets called before the property is applied and before rendering happens. This means that it can get the initial state of the element and record that so that it can be used within the `render` callback.

antitoxic:
@treshugart Thank you. In my usecase I use `[resolved]` and `[unresolved]` to show a page loader. I understand the concept of loading each element on its own but that looks very glitchy as height and width can change. And having a page-wide loader basically needs all things to be resolved which as you said is hard.

----
I know JSX is different from React. I just don't know what other compilers are there for it? I mean, even BABEL transpiles to React.

----
If we don't compile the virtual tree, how do we get the actual rendering output that needs to be compared? I'm sorry if I sound silly. I just don't get if we have content slots and we only render the virtual tree without the subtree which populates the slots (*so we don't know what goes in the slot*) how do we know the exact DOM difference? Or am I relating this pull request too much with `<content select="">` concept?

marcoms:
(unrelated to my last question)

@treshugart If the aim is to provide an imperative alternative to shadow DOM-like distribution, then perhaps the `selector` property should be restricted to ids only, since in shadow DOM, elements can only be distributed once (if I'm not mistaken)

treshugart:
@marcoms that's a fair point. In the named-slot API, you'd normally use a `<slot>` element that is invisible in the rendered DOM. For posterity, https://gist.github.com/treshugart/b67f47118bd89d5ab70e is a gist that incorporates the named slot API through a property. Maybe we could work that into this.

Also, I'm thinking that since this PR is growing in scope that it might be good to pull out into its own module / repo.

treshugart:
@antitoxic re: If we don't compile the virtual tree, how do we get the actual rendering output that needs to be compared?

So what happens is that the initial mount records the last virtual tree that was rendered. When it goes to render the new tree, it compares it to the last tree. That way, it doesn't need to concern itself with the expanded output, just what it knows is there.

antitoxic:
I think I'm pausing on this issue. May be skate as a whole. I feel out of context. I can't see how comparing 2 virtual trees is helping for things like:

> The skate.ready() function should not be confused with the ready lifecycle callback. The lifecycle callback is called when the component element is ready to be worked with. It means that it's been templated out and all properties have been set up completely. It does not mean, however, that descendant components have been initialised.

If skate element is not aware of descendants and their logic, respectively their `render` logic, how can we compare stuff that are not initialised. It's all seems to boil down to the freaky mutation observation. That's why virtual-dom and virtual-element libraries doesn't seem to have that problem (React, riot, ember) - because they dont use mutation observations.

I've done so much reading on frontend the last year. The vast amount on it feels silly. It's just UI, it shouldn't be that brain-consuming. ref: https://medium.com/@ericclemmons/javascript-fatigue-48d4011b6fc4

I feel I need to be suuuper-careful on each new library that I use. It comes with its own dependencies and its own opinionated way of doing stuff. Even for object-merging there're several popular things (object.asssign, lodash.merge, deepmerge). And having 3 libraries , each using one different of those is simply crazy.

Sorry for the rant. However I think it's important for skate to point this out.There's a "war" on frontend practices right now. Web components traction is slowing, v-dom and v-components are speeding.

Skate should focus on strong sides and talk with the rest of the group. Otherwise a huge JS fatigue is put on developers that choose to use it.

treshugart:
> I think I'm pausing on this issue. May be skate as a whole.

@antitoxic Sucks to hear, but each to their own. If you have any specific questions please feel free to DM me.

> I can't see how comparing 2 virtual trees is helping for things like:

You don't have to concern yourself about the timing of descendant components at all if you're using a differ because everything is happening in `render()`. You shouldn't ever need to use `ready()`. Even if you're using the other rendering method of stamping out a template and mutating stuff in property setters, most of the time you're not going to have to worry about it. It's there for the edge-cases.

> It's all seems to boil down to the freaky mutation observation.

The problem also exists in native and is dependent on when and where your components are loaded. It's not specific to mutation observers. It's actually possible to make mutation observer / polyfill based custom elements behave more consistent and predictable than native, unfortunately. This behaviour is detailed in https://github.com/skatejs/skatejs#background.

It seems to me like your frustration, with Skate at least, is the fact that you can have FOUC and it's not handled automatically by Skate itself?

treshugart:
@antitoxic 

> Sorry for the rant. However I think it's important for skate to point this out. There's a "war" on frontend practices right now. Web components traction is slowing, v-dom and v-components are speeding.

You're right, and it's really sad. I see it stemming from two issues:

1. The web component spec authors are working at a much slower pace than the alternative solutions to the problems they're trying to solve (ES6 modules vs HTML imports, custom elements vs React). This is bad. Especially when there's been so much unnecessary churn.
2. Many loud and influential people in the React (et al) community are openly in opposition to web components, and unnecessarily, IMO. This is also bad because it diverges the community, hence the so-called war.

Back to Skate, I'm not exactly sure what there is for it to "point out". I'm open to ideas, though.

> Skate should focus on strong sides and talk with the rest of the group. Otherwise a huge JS fatigue is put on developers that choose to use it.

Which strong sides would you suggest? By "group" do you mean the Skate user-base, or with other communities like React and Polymer?

antitoxic:
@treshugart You're right, Facebook is behind React and they are powerful. But it's also true that Google is behind Polymer and most push of webcomponents. Google also is pushing `incremental-dom`.

> What to point out?

As many state, npm is very saturated. JS went from a moduless state to a state of chaos and thousands loosely coupled modules that a developer must spend a day to glue together and keep spending time to make them work.

What skate (*my opinion of course, I can't say it is globally true*) should do is add examples of intended uses that include best practices. Yes this goes to `opinionated` area but I as a developer want to see the opinion of the creators of a product that I like. It's highly likely that my opinion matches theirs. Simple examples that show the advertised power of simplicity. Simple examples that let me develop "as it's intended". Web apps are not "Hello world". They include parallel animations, modals, input validation, tooltips, powerful dropdowns, page transitions. How do we do all these?

How do skate components communicate? What libraries best compliment skate? - this includes matching dependencies, speed, matching pattern of workflow so on. What router do you think is best fit for skate? What's best matching package for ajax-request making? How do you do bullet-proof input validation? - I currently use `parsleyjs` - what I can replace it with? How do you do tooltips that are not influenced by `overflow:hidden`? - this usually means appending tooltips and the end of the `<body>` but webcomponents promote encapsulation, this again goes back to How components communicate? I currently used `magnific popup` for modals, its the most robust UX-wise, it accounts for document scrollbar and many other stuff. What should I use in replacement of selectize/select2? - these two have become the goto replacement for advanced dropdowns, lazy loading,tagging..... 

Right now I the docs show simple stuff, but If I'm using components I'm not making simple stuff, I'm making an app. And if I'm sticking to best practices JS-wise I definitely do best practices in graphic design and user interaction which include huge amount of animation.

antitoxic:
Another example comes to mind: Skate stays away from giving examples of templating as it allows any templating. Ok, I get that. But templating is a thing you can't do without. What is your best-practice template accroding to you? I see examples of virtual dom at some issues. I see JSX here. So what tempting you think as optimal?

antitoxic:
And last but not least: If virtual dom is recommended how do I do server-side rendering?

treshugart:
@antitoxic I see what you're saying and there is some work to be done there. That said, have a look at https://github.com/skatejs/skatejs#examples. In terms of recommending libraries that go well with it such as routers and ajax requests, that's really hard because any one of them can pair well. What might fare better is picking one of each and showing examples.

I'm more than happy for people (you?) to raise issues and discuss creating examples for these. That's how the TodoMVC example was created. I have very limited time, but it's easier to prioritise when there is something there to prioritise.

It'd also be great if you could raise an issue for the problems you're having regarding `resolved` etc so that we can discuss that stuff there. I'm happy to try and come to a solution that makes that easier for you. At the very least, some best practices might come out of it.

As for this PR, I'd like to get it back on track.

antitoxic:
:+1:

marcoms:
@treshugart, FWIW I would be more comfortable with this api being separate from the main skate project. I think being a relatively vanilla, unopinionated library is a good position to have at this time.

I also agree with @antitoxic that we could have more examples of recommended "best practices". Inline snippets in the readme I think would be very nice.

treshugart:
@marcoms :+1: 

In terms of examples, since you guys have been working with it, I'd encourage you guys to raise issues around introducing some examples. Blog etc. I can do some of it, but I'm so close to the code, it's probably better if others start doing that.

I'm going to keep this PR open for now, but we should start discussing how we might create a separate solution.

treshugart:
I'm going to close this. In retrospect it's a little obtuse and there seems to be a little confusion about it. I've started some ideas on named slot partial-filling over here: https://github.com/skatejs/named-slots.

Issue #447, opened 2015-12-02
Allow the passing of a function as the component definition


treshugart:
If a function is passed as the component definition, it should be used as the `created` callback. https://github.com/w3c/webcomponents/wiki/Custom-Elements:-Contentious-Bits#constructor-vs-createdcallback.

The rest of the API should not change (e.g. return value). When this is natively supported, we can still wrap the returned constructor so that it allows it to be called like a function (and can accept props).

treshugart:
Unfortunately we'll have to wait for native support because you can't call native constructors as functions. If you try and new them and you're transpiler has transpiled the es6 -> es5, it will try and call the native constructor as a function. The only scenario this would work for is passing a user defined constructor. We must wait for better native support at this time, I believe.

treshugart:
Re-opening because this will need to be done. Adding a link to the W3C issue about authoring guidance  for reference: https://github.com/w3c/webcomponents/issues/409.

treshugart:
Actually, I think that this isn't very relevant unless we decide to mimic the spec verbatim. It may be better to keep the `created` callback and to do automation stuff in the constructor. Closing this for now until we come up with a good reason to use `constructor` directly.

PR #453, opened 2015-12-19
Enable enumerating ES6 class properties as additional options

antitoxic:
This takes in account when options are being passed as ES6 class. Related to: https://github.com/skatejs/skatejs/pull/450

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -32,11 +32,7 @@ function makeOptions (userOptions) {
   }
 
   // Copy over non-standard options.
-  for (let name of Object.getOwnPropertyNames(userOptions)) {
-    options[name] = userOptions[name];
-  }
-
-  return options;
+  return assignSafe(options, userOptions);
 }
 
 function makeNonNewableWrapper (Ctor, opts) {
--- a/src/index.js
+++ b/src/index.js
@@ -32,7 +32,7 @@ function makeOptions (userOptions) {
   }
 
   // Copy over non-standard options.
-  for (let name in userOptions) {
+  for (let name of Object.getOwnPropertyNames(userOptions)) {
     options[name] = userOptions[name];
   }
 


treshugart:
Unfortunately using `for of` requires a `Symbol` polyfill which I don't think we should be using at this stage (see build). You can try using `for in` with `hasOwnProperty()` checks, but I had issues with only doing own properties and ES6 classes at one point (don't recall the exact issue). Maybe it's ok now.

src/index.js
```diff
@@ -32,7 +32,7 @@ function makeOptions (userOptions) {
   }
 
   // Copy over non-standard options.
-  for (let name in userOptions) {
+  for (let name of Object.getOwnPropertyNames(userOptions)) {
```
treshugart: You might be able to `assign(options, userOptions)` (via `import assign from 'object-assign') here instead which may fix the CI (might cause other issues we don't know about).

antitoxic: I saw `object-assign` a bit later on. I will switch to it for this.

antitoxic:
:christmas_tree: commit with `assign-safe` jingle while watching "Home Alone" for the 100th time.

treshugart:
@antitoxic I wonder if we should just implement https://github.com/skatejs/skatejs/issues/447. I don't think this is copying over props because ones inherited from classes aren't "own".

treshugart:
Tried fixing in #462. Maybe we can write a test for this in that one?

treshugart:
Hey @antitoxic, just reading this in relation to #450, which is now merged, I'm not sure what exact problem this is solving. Could you elaborate? If I read this correctly, it's making class properties enumerable when they're not by default. It sounds ok to me to leave them as is, unless it's solving a use-case. Thoughts?

antitoxic:
Probably not solving much. I'm simply mimicking the behavior that skate has when it's passed a plain javascript object as options.

Without this patch, behaviors differ. It doesn't bring any problems right now. However a inconsistency might bring problems later on. The #450 is an example of one such inconsistency. #450 is not a bug for passing options as plain object but it is when passing options as class (function).

treshugart:
I wonder if #450 is still an issue in master. I think this behaviour is perfectly fine. You could pass a normal object with enumerable props and you'd get the same behaviour as a class. Going to close, but I'd be happy to re-open if there is a use-case or bug related to this.

antitoxic:
If this pull request or something similar is not merged the small problem that might not be obvious but will remain for the future is that  if skate options were changed but "default options" were not plain object settings will work but class setttings won't. Just to keep in mind.

treshugart:
Having defaults out of sync with the API is a bug, so those should always be correct as your other PR corrected :)

antitoxic:
That begs the question why is than this needed:

```
// Copy over non-standard options.
for (let name in userOptions) {
  options[name] = userOptions[name];
}
```

at all?

treshugart:
It's not in master anymore. This is what the file looks like now: https://github.com/skatejs/skatejs/blob/master/src/index.js#L61.

antitoxic:
:+1: My bad. That's cool.

PR #454, opened 2015-12-21
Adds the web-components keyword

zenorocha:
That way this project can be fetched by customelements.io


treshugart:
Thanks @zenorocha!

PR #458, opened 2015-12-23
FIx issue where the internal value wasn't being passed through to the…

treshugart:
… setter. Also added more information to the get function.

The goal of this PR originally was to be able to pass the internal value through to the get function so that we can get the current value and return a different value based on that if necessary. The other goal was to be able to return different values from `get` than was passed into `set` by not actually calling the getter for the initial value in the `created` lifecycle.

Another part of this PR was to make properties behave consistently for `undefined` and `null` values. If we define a value as empty when it is either `undefined` or `null`, then we can assert that an attribute will be removed if a property or default value is considered empty. The same thing happens for non-boolean properties. If you pass an empty value, then it will not be coerced, but set to strictly `undefined`. This is to take any ambiguity away from how props accept empty values. How some things use undefined vs null in JavaScript is one of the more frustrating aspects of it IMO.

The final piece in this PR deals with attribute linkage. In order to take away any confusion in how they are linked together (special cases etc), we've made it so that an attribute and property are *always* linked if they are specified as linked in the options. This means that `undefined` / `null` will remove an attribute. Any other value will set it, including a default value if specified.


src/lifecycle/properties-init.js
```diff
@@ -79,13 +79,20 @@ function createNativePropertyDefinition (name, opts) {
 
   prop.get = function () {
     const info = data(this, `api/property/${name}`);
+    const internalValue = info.internalValue;
 
     if (opts.get) {
-      return opts.get(this);
+      return opts.get(this, { name, internalValue });
```
treshugart: Do we like this name?

src/lifecycle/properties-init.js
```diff
@@ -79,13 +79,20 @@ function createNativePropertyDefinition (name, opts) {
...
     }
 
-    return info.internalValue;
+    return internalValue;
   };
+  
+  prop.init = function () {
```
treshugart: Instead of putting more logic in the `properties-ready` lifecycle, this seemed a better place as it co-locates the internal `data`.

src/lifecycle/properties-init.js
```diff
@@ -97,20 +104,11 @@ function createNativePropertyDefinition (name, opts) {
...
-    }
-
     if (typeof opts.coerce === 'function') {
       newValue = opts.coerce(newValue);
     }
 
-    if (!opts.get) {
-      info.internalValue = typeof newValue === 'undefined' ? info.defaultValue : newValue;
-    }
+    info.internalValue = typeof newValue === 'undefined' ? info.defaultValue : newValue;
```
treshugart: If we want access to the internal value then we need to always set it.

src/lifecycle/properties-init.js
```diff
@@ -97,20 +104,11 @@ function createNativePropertyDefinition (name, opts) {
 
     info.updatingProperty = true;
 
-    if (info.hasBeenSetOnce) {
```
treshugart: The `init` function takes care of the initial setting, so no need to set this anymore.

test/unit/lifecycle/properties.js
```diff
@@ -3,6 +3,24 @@ import propertiesCreated from '../../../src/lifecycle/properties-created';
 import propertiesReady from '../../../src/lifecycle/properties-ready';
 
 describe('lifecycle/property', function () {
+  function initProperty (elem, definition, name = 'test') {
```
treshugart: Made the utils more accessible to other scopes in these tests.

test/unit/lifecycle/properties.js
```diff
@@ -199,14 +199,27 @@ describe('lifecycle/property', function () {
         expect(elem.test).to.equal('something');
       });
 
-      it('context and arguments', function () {
+      it('context and arguments', function (done) {
```
treshugart: This test wasn't fleshed out enough, but I did add more parts to it to test the new params.

test/unit/lifecycle/properties.js
```diff
@@ -325,4 +338,21 @@ describe('lifecycle/property', function () {
       });
     });
   });
+  
+  describe('patterns', function () {
```
treshugart: This was the actual bug where you wanted to use the internal value in the setter when you specify a custom getter.

src/lifecycle/properties-init.js
bqx: Is this new API? i.e. you have access to an object with `internalValue` when you call the `get()` method?

src/lifecycle/properties-init.js
```diff
@@ -79,13 +79,20 @@ function createNativePropertyDefinition (name, opts) {
 
   prop.get = function () {
     const info = data(this, `api/property/${name}`);
+    const internalValue = info.internalValue;
```
bqx: Could shorten this to one line:
```
const internalValue = data(this, `api/property/${name}`).internalValue
```

src/lifecycle/properties-init.js
```diff
@@ -79,13 +79,20 @@ function createNativePropertyDefinition (name, opts) {
...
+      return opts.get(this, { name, internalValue });
     }
 
-    return info.internalValue;
+    return internalValue;
   };
+  
+  prop.init = function () {
+    const info = data(this, `api/property/${name}`);
+    const init = info.internalValue;
```
bqx: Could also shorten to one line by removing `info`

src/lifecycle/properties-init.js
```diff
@@ -79,13 +79,20 @@ function createNativePropertyDefinition (name, opts) {
...
+    this[name] = typeof init === 'undefined' ? this[name] : init;
+  },
```
bqx: Is the `,` meant to be a `;`?

src/lifecycle/properties-ready.js
```diff
@@ -1,6 +1,5 @@
 export default function propertiesApply (elem, properties) {
   Object.keys(properties).forEach(function (name) {
-    let prop = properties[name];
-    prop.set && prop.set.call(elem, elem[name]);
+    properties[name].init.call(elem);
```
bqx: Does this mean that `set` is no longer called when the element is initialised? Does this have any other side effects? (It looks like `init` doesn't do a lot of things that `set` did)

test/unit/lifecycle/properties.js
bqx: Would be good to test that you have access to `data.internalValue` inside `get()` as well

src/lifecycle/properties-init.js
treshugart: Yep.

src/lifecycle/properties-init.js
treshugart: Good catch.

src/lifecycle/properties-ready.js
treshugart: In the `init` function, it does call set. This `init` function isn't part of public API.

test/unit/lifecycle/properties.js
treshugart: That's already tested here https://github.com/skatejs/skatejs/pull/458/files#diff-c58358d2c478f1fec77c97e11121e54fR211.

src/lifecycle/properties-init.js
treshugart: Fixed.

src/lifecycle/properties-init.js
treshugart: I'll try that.

src/lifecycle/properties-init.js
```diff
@@ -23,22 +28,30 @@ function createNativePropertyDefinition (name, opts) {
...
+    // TODO Refactor to be cleaner.
     if (info.linkedAttribute) {
       if (!info.attributeMap) {
         info.attributeMap = {};
 
         elem.removeAttribute = function (attrName) {
           info.updatingAttribute = true;
-          info.removeAttribute.call(this, attrName);
+
+          if (defaultValueIsEmpty) {
```
treshugart: This ensures that if a default value exists, that the attribute is returned to that value so it is always in sync with the prop.

src/lifecycle/properties-init.js
```diff
@@ -23,22 +28,30 @@ function createNativePropertyDefinition (name, opts) {
...
+            info.removeAttribute.call(this, attrName);
+          } else {
+            info.setAttribute.call(this, attrName, defaultValue);
+          }
 
           if (attrName in info.attributeMap) {
             const propertyName = info.attributeMap[attrName];
-            elem[propertyName] = undefined;
+            elem[propertyName] = defaultValueIsEmpty ? undefined : defaultValue;
```
treshugart: Default value for prop.

src/lifecycle/properties-init.js
bqx: Gotcha, sounds good.

bqx:
lgtm

PR #460, opened 2015-12-29
Document the 'detail' option for skate.emit

bqx:
Add a short section to document the `detail` option for `eventOptions` in `skate.emit`


bqx:
Was asked how to do this and realised it was missing from the docs, so here it is.

treshugart:
lgtm

PR #465, opened 2016-01-07
skate.emit(el) on a disabled event does not fire an event. Fixes GH-463

jpnelson:
Fixes GH-463
This is because we cannot due it in firefox due to https://bugzilla.mozilla.org/show_bug.cgi?id=867418 .

For consistency, we will not trigger events on any browser if the element is disabled.

In addition, events cannot bubble through a disabled element.


jpnelson:
Need to do docs for this too.

Issue #456, opened 2015-12-22
Remove data-skate-ignore attribute.


treshugart:
This was originally added to mitigate performance issues in IE9 when there were 50k+ elements being mutated at the same time. It was - and still is possibly only used by us at Atlassian - and it'd probably be a good idea to remove this from the API before going 1.0.

This is only breaking in the sense that it may cause unexpected performance hits in old IE.

jpnelson:
Is there an alternative to this? I imagine it could still be useful for people trying to mitigate performance issues in other browsers

treshugart:
That's a good point. We need some data on this for a more solid "yes". This really should only be needed for IE9 and 10, though. It's non-functional if native web components are supported.

treshugart:
I removed this from the docs in #548 so that people don't start using it. We'll leave it in for 1.0 but look at removing it in a major release after that.

treshugart:
This has in fact been removed for 1.0.0 as we're using webcomponentsjs v1 polyfills.

PR #467, opened 2016-01-12
#456 remove data-skate-ignore functionality.

treshugart:
Implements #456.


treshugart:
Closing this without merging as this will be a breaking change down the road once we can be sure products don't need them.

PR #470, opened 2016-01-19
More prop consistency and minor additions.

treshugart:
- `default()` now has a data param with "name" in it
- `data.initialValue` -> `data.internalValue` (consistency with `created` and `get` data)
- `prop.init` -> `prop.ready` (better name that corresponds with the element lifecycle)
- `prop.initial` is now a function that can return the initial value that is set as the internal value before anything is called. This is called before the prop is even defined. This may change in the future depending on when WebKit fixes [this bug from 2010](https://bugs.webkit.org/show_bug.cgi?id=49739).

This came out of a need for being able to return the initial value of a property if that differs from the `default` value. One thing to remember is that `default` is called and used as the value for the prop every time an empty (`undefined` or `null`) value is set, which includes the initial value. The `initial` option is used to return an initial value for the property. Before we just tried to "get" the prop, but now this is an alternative to that, which defaults to that.

A generic use case for this is creating a property whose initial value should be calculated. Both `default` and `coerce` are called multiple times and have different semantics, so they're not suitable for this. The `created` callback does not return a value and is called once *after* the property has been defined. The `initial` callback should be called *before* the property is `created` so that when it is created, this initial value can be used in subsequent algorithms.


src/lifecycle/properties-created.js
```diff
@@ -17,6 +17,6 @@ export default function propertiesApply (elem, properties) {
     // Once that bug is fixed, the initial value being passed as the second
     // argument to prop.created() can use the overridden property definition to
     // get the initial value.
-    prop.created && prop.created(elem, initialValue);
+    prop.created(elem, initialValue);
```
jpnelson: Does this have a default value now? I wonder why it checked `props.created` before but doesn't now.

lib/lifecycle/properties-init.js
```diff
@@ -126,11 +126,11 @@
       }
 
       // We must coerce the initial value just in case it wasn't already.
-      info.internalValue = opts.coerce ? opts.coerce(initialValue) : initialValue;
+      var internalValue = info.internalValue = opts.coerce ? opts.coerce(initialValue) : initialValue;
```
jpnelson: `const`?

lib/lifecycle/properties-created.js
```diff
@@ -18,14 +18,14 @@
   function propertiesApply(elem, properties) {
     Object.keys(properties).forEach(function (name) {
       var prop = properties[name];
-      var initialValue = elem[name];
+      var initialValue = prop.initial(elem);
```
jpnelson: `const` for these?

jpnelson:
> so that when it is created, this initial value can be used in subsequent algorithms.

What's an example use case where you'd need to `initial`ise a property, and then use `created` afterwards? I wonder if we can get away with just one of these

test/unit/lifecycle/properties.js
```diff
@@ -464,9 +465,58 @@ describe('lifecycle/property', function () {
...
+            expect(data.name).to.equal('test');
+            expect(data.internalValue).to.equal(undefined);
+            done();
+          }
+        };
+        create(opts);
+      });
+    });
+
+    describe('initial', function () {
```
jpnelson: Is there a test where the `initial` callback returns a value that is then used as the internal value? I think that would be useful to have

src/lifecycle/properties-created.js
treshugart: It did because we were possibly thinking about allowing a function to return a user-defined prop definition, but we haven't documented this and don't publicly support it yet. So currently, there will always be a `created` function here.

test/unit/lifecycle/properties.js
```diff
@@ -29,35 +29,35 @@ describe('lifecycle/property', function () {
...
+
   it('should not leak options to other definitions', function () {
     const elem = element().skate({
       properties: {
-        test1: { 
-          attribute: true, 
-          default: 'test1', 
+        test1: {
+          attribute: true,
+          default: 'test1',
```
jpnelson: What happens if you specify a `default` as well as a callback for `initial`?

lib/lifecycle/properties-init.js
treshugart: Visual studio code was complaining with `const` and a double assignment there so I played it safe.

treshugart: This is `lib/`.

lib/lifecycle/properties-created.js
treshugart: This is `lib/`.

treshugart:
> What's an example use case where you'd need to initialise a property, and then use created afterwards? I wonder if we can get away with just one of these

https://github.com/skatejs/named-slots/blob/master/index.js#L179

test/unit/lifecycle/properties.js
treshugart: `default` is only used if the value, or initial value is "empty" (`null` or `undefined`). This also means that `default` can be used not only to initialise it if `initial` isn't specified, or doesn't return a value, it is also used if the user sets the value to empty.

jpnelson: Ok, I think I understand. So if `default` and `initial` were both specified, the return value for `initial` would be used?

jpnelson:
LGTM

test/unit/lifecycle/properties.js
treshugart: > Ok, I think I understand. So if default and initial were both specified, the return value for initial would be used?

That's correct.

> Only thing would be documentation around the differences between default, initial and created, as I imagine this would be initially confusing.

Totes. I'll leave this in PR until we can document it then. It'll finally get me to document those options :)

treshugart:
lgtm

Issue #479, opened 2016-02-11
Modify internals of skate.ready() so that it doesn't use events, but element data instead.


treshugart:
References #478 

This currently causes performance issues because of the overhead in triggering an event in the created callback. We could improve performance by using data that `skate.ready()` attaches to an element instead of triggering an event.

In the `created` callback after `ready`:

1. Check for registered callbacks.
2. If callbacks exist, execute the callbacks in the order they were attached.
3. Remove the callbacks if there were any.

In `skate.ready()`, for each element selected, if the element is not ready:

1. If no data exists, create an empty array to store callbacks.
2. If data does exist, use that as the array to add the callback to.
3. Add the callback to the array of callbacks.

If the element is ready:

1. Execute the callback.

marcoms:
In native, loading scripts at the bottom and importing dependencies inside those definitions works fine as a replacement (as described in `skate.ready()` docs), but is there an alternative for polyfilled web components?

treshugart:
@marcoms

> but is there an alternative for polyfilled web components

No there isn't. I'm inclined to keep it but to modify the internals of it so that it doesn't trigger events and uses data stored on the element. It'd be nice if there was a performant way of doing something when all descendants are set up so that the `ready` lifecycle callback could be used instead, but I'm not sure how to do that. That still wouldn't solve the public API case where you select the element and need to work with it which `skate.ready()` does for all cases internally and externally.

I reckon we just modify the internals if we can. I'll update the issue to reflect that.

PR #482, opened 2016-02-16
#479 - Fix performance issues with skate.ready()

treshugart:
Also removed spots that didn't need to use a data namespace.

Results:

44,758 -> 169,962 ops/sec (Chrome) and 146,570 opts/sec (Firefox)

This is compared to 314,351 ops/sec for the native custom element implementation (Chrome) and 89,190 ops/sec for the custom element polyfill (Firefox).

Breaking changes are that this only now calls the callback for component elements. It used to fire for any element by checking to see if it had a component definition. However this causes problems because you may not have registered the element yet and don't want the callback to fire until you do. You can't have both. It also simplifies the behaviour of this function and you should only be passing in component elements anyways.

### DIFF
--- a/src/api/ready.js
+++ b/src/api/ready.js
@@ -3,7 +3,7 @@ import registry from '../shared/registry';
 
 function ready (element) {
   const component = registry.find(element);
-  return !component || data(element, `lifecycle/${component.id}`).created;
+  return component && data(element).created;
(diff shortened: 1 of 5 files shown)


treshugart:
@a-ignatov-parc you may be interested in this

PR #487, opened 2016-02-23
chore(bower): remove install step from build

joscha:
as we got rid of all direct bower dependencies, the `bower` install step is not needed any more in the travis build.


joscha:

PR #488, opened 2016-02-23
chore: use stable node for testing

joscha:
This uses node latest for our build


joscha:

PR #489, opened 2016-02-23
chore: add more saucelabs browsers

### LINKED ISSUE #457: Increase number of tested browsers in Sauce labs.
Probably good to go as far back as we possibly can. At the very least we could assess whether or not we should fix any failures in older browsers that may still be getting a lot of use.

joscha:
* removed Chrome 43, because `constructor should be accessible` was failing for some strange reason
* removed Opera 11 because it couldn't deal with `.stopImmediatePropagation`

68747470733a2f2f73617563656c6162732e636f6d2f62726f777365722d6d61747269782f736b6174656a732e737667_svg

closes #457.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "devDependencies": {
     "eslint": "2.2.0",
     "gulp": "gulpjs/gulp#73eced0",
-    "skatejs-build": "skatejs/build#1242c84",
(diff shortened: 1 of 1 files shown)


joscha:
sha needs to be updated once https://github.com/skatejs/build/pull/72 has been merged.

joscha:

Issue #491, opened 2016-02-25
Add array() property type.


treshugart:
- `serialize()` using `JSON.stringify()`
- `deserialize()` using `JSON.parse()`
- `coerce()` should ensure that whatever is passed is an array (any non-array becomes the first element in a new array, an array remains unchanged)
- `default()` should return an empty array

marcoms:
Maybe an optional `delimeter` could be the second argument?

PR #492, opened 2016-02-26
#491 - Add array() prop type.

treshugart:
Implements #491.


src/api/properties/array.js
```diff
@@ -0,0 +1,14 @@
+export default {
+  coerce (val) {
+    return Array.isArray(val) ? val : [val];
+  },
+  default () {
+    return [];
+  },
+  deserialize (val) {
+    return val.split(',');
```
joscha: if the serialized value contains a `,` this fails.

src/api/properties/array.js
```diff
@@ -0,0 +1,14 @@
...
+  },
+  serialize (val) {
+    return val.join(',');
```
joscha: if the array value contains a `,` this fails.

treshugart: This is true. Maybe it's worth being able to specify a `token` or something to `split` or `join` using? It's worth nothing that this should be a simple implementation. If users need complex behaviour, they could throw an abstraction on top of it or write their own property.

joscha: what about `JSON.stringify`? Should be fairly fast due to native impl.

treshugart: Could do that. Maybe just change this prop type to json()?

treshugart: But if we did that, then you'd have to escape double quotes, and it'd be a mess. I reckon for simple enum-style values, this'd be simpler.

marcoms:
I agree, not great for the user to have to ensure valid JSON if they just want a simple list. If they want the flexibility of JSON, I think it should be a separate property type.

@treshugarts' suggestion of a token or delimiter parameter I like :).

src/api/properties/array.js
joscha: HTML allows both `'` and `"`, so it should be okay, no quoting needed.

joscha:
@marcoms true it's a bit more verbose, but at least it works (tm), whereas specifying a token delimiter is another param to take into account and won't work for non-controlled array values.

marcoms:
Don't these two types serve different purposes though? A list property type might be more useful when linked to an attribute, and would prioritise ergonomics (often controlled data - element ids, etc.), while JSON would make more sense if used internally or as part of a component's API, prioritising structure and consistency (safe for uncontrolled values).

treshugart:
I see both `array()` and `json()` as helpful. The former more in the short-term, though, as I use the behaviour quite often. I do agree that it should possibly be renamed. Maybe `enum()`? Once this is merged was thinking about proposing an argument that you can specify a whitelist and an optional default so it behaves more like an HTML enum attribute. With that, it's also worth extending it with a `delimitter` option so things other than commas can be used.

marcoms:
Enum is technically correct but it's use in many languages give the idea of a fixed set of values you are allowed to use. Is that the intention from the second sentence?

Also, what do you mean by an HTML enum attribute?

treshugart:
@marcoms actually, you're right. It might not be clear calling it `enum()` as it allows any value by default. Maybe `list()` is better? The thing is, it's really just an array. The delimiter is strictly for going to / from an attribute, which may or may not be enabled. If that's the case, maybe JSON encoding it, or JSON decoding it is better.

joscha:
I think specifying delimiters is kind of 1998 :) - it's virtually just workarounds. Using JSON encoding will keep the API concise and clear and will take care of any hidden quirks.

treshugart:
@marcoms @joscha I've tried to do the best of both worlds here. Instead of manual parsing, I've decided to use `JSON.parse()`. Much more robust and faster. However, I'm still going to call it `array` as it's intended to coerce non-array's to arrays and is strictly an array.

treshugart:
I'm going to go ahead and merge this if the build passes. It seems like we can agree on initial functionality, just maybe not the implementation detail.

PR #494, opened 2016-03-02
fix(globals): do not share globals between skate versions

### LINKED ISSUE #493: Skate 0.13 and 0.12 use the same global namespace
They both use `window.__skate`, but they are not compatible in the way they store their lifecycle information (e.g. `created` string vs object). In order to fix 0.13 and 0.12, we need to backport https://github.com/skatejs/skatejs/blob/master/src/shared/vars.js

cc @seancurtis @jpnelson @treshugart

joscha:
This fixes #493 for 0.13.x

refs #495

@treshugart @jpnelson @seancurtis PTAL


joscha:
tests are actually green - only karma dies afterwards, probably a build dep problem.

joscha:
build fail was due to https://github.com/karma-runner/karma/issues/1782 - updated karma in order to get this green.

jpnelson:
lgtm

chrisdarroch:
I'm not terribly familiar with this part of the code, but... is the implication that skate will cease to create a window global? Or is it that a shared mutable state has been removed, and a `window.skate` global will still continue to exist once installing the library?

joscha:
@chrisdarroch `window.skate` is not touched, only `window.__skate` which always has been internal.

seancurtis:
Looks good

PR #495, opened 2016-03-02
fix(globals): do not share globals between skate versions

### LINKED ISSUE #493: Skate 0.13 and 0.12 use the same global namespace

joscha:
This fixes #493 for 0.12.x

refs #494 

@treshugart @jpnelson @seancurtis PTAL


jpnelson:
lgtm

seancurtis:
lgtm

joscha:
skatejs@0.12.7 released

Issue #501, opened 2016-03-07
Modularise shared vars.


treshugart:
See #498 for more information. Basically this removes the global sharing of the registry and document observers.

PR #502, opened 2016-03-07
#501 - Modularise shared components.

treshugart:
Implements #501.


joscha:
Not sure why that failed. It seems to work in #503

treshugart:
Looks like it was a problem with the Sauce tunnel.

joscha:
we should be able to delete `vars.js` I think?

joscha:
LGTM

joscha:
I think we need @jpnelson to add his LGTM, given that we both made changes...

treshugart:
Good point @joscha, probably should delete the `vars.js`. @jpnelson what do you think about the performance implications here?

joscha:
@treshugart - I removed it, however we need @jpnelson to review it :)

treshugart:
Right, thanks @joscha, I hand't checked commits.

treshugart:
lgtm

joscha:
LGTM

PR #503, opened 2016-03-07
Add compat test with 0.14.x

joscha:
refs #501


joscha:
can you check that locally again @treshugart ?

treshugart:
I'll merge #501 before testing this and merge this if things look good locally and on Travis.

PR #515, opened 2016-03-16
fix: use saucelabs latest

joscha:
This uses the Saucelabs latest feature.


treshugart:
Hmm, lots of "Cannot start latest..."

joscha:
yes, because I didn't specify the platform :-/ I hoped SL would figure out a platform automatically.

joscha:
Opened a ticket with SL support.

treshugart:
(thumbsup) Thanks for looking into this!

treshugart:
### Sauce issue thread

#### skatejs wrote:

I was trying to use: http://sauceio.com/index.php/2016/03/new-browser-version-shortcuts/

I noticed that when using a definition with “latest-n” and not specifying a platform, Saucelabs doesn’t figure out a valid one by itself, the build just doesn’t start.
E.g.

ERROR [launcher.sauce]: Can not start safari latest-1
[init({“version”:”latest-1″,”platform”:”ANY”,”tags”:[],”name”:”Skate unit tests (master)”,”tunnel-identifier”:”karma1458098888″,”record-video”:false,”record-screenshots”:false,”device-orientation”:null,”disable-popup-handler”:true,”build”:”2520″,”public”:null,”customData”:{},”browserName”:”safari”,”base”:”SauceLabs”})] The environment you requested was unavailable.
Unsupported OS/browser/version/device combo: OS: ‘unspecified’, Browser: ‘safari’, Version: ‘8.0.’, Device: ‘unspecified’

Would be good if that can be added, otherwise the definition needs to be changed with every major release that shifts the platform.

The definition is here: https://github.com/skatejs/build/commit/8646983a1b7417531cafd81e1b22adeb7cc9041a

#### Matt Dunn replied:

Hi,

The intention behind the latest-n browser version capability is really for browsers where we support multiple versions of a browser per OS. That’s Chrome and Firefox really and yes, you still have to specify an OS. For Safari, we support one version of Safari per OS X version.

That said, it would be a good idea to be able to just specify a browser with “latest” and have Sauce Labs work out the correct OS to run on rather than have to specify this yourself. I’m going to speak to the Product Manager who’s responsible for this area and see whether that’s on the roadmap.

I’ll let you know what I find out.

Matt

joscha:
still 9 failures on Safari 7.0.6 - Saucelabs seems to only return 7.0.6 now

joscha:
I disabled Safari 7 for now and opened a ticket with SL - seems as if even when specifying `7.17` explicitly, you still get a `7.0.6`.

treshugart:
I disabled the sauce image in the readme as it was severely broken. Might be good to uncomment that.

treshugart:
lgtm if tests pass

Issue #513, opened 2016-03-16
Property change event


jpnelson:
It'd be useful to have a property change event for creating stateless components. This could then be prevented, and a stateless component made.

We could make this optional on the property level – whether or not it triggers an event, and is preventable.

treshugart:
```js
skate('my-element', {
  properties: {
    event: true // or 'event-name'?
  }
});
```

^ ??

Event should probably be:

```js
{ bubbles: false, cancelable: true }
```

treshugart:
On second thought, it should probably be cancelable, we'll just need to be aware of the above issue when automating the React integration.

treshugart:
### Outstanding items

The outstanding things for this as I see it are:

#### Default value

What's the default value? I'd vote for `false`, or some form of falsy. It'd be a big assumption to have on by default as it would create a significant amount of overhead for those who wouldn't be using it.

#### Allow boolean `true` or just `String`

Should we allow `true` to be specified? If so, what's the default event name we want to use? I'd vote for not supporting `true` and have it always coerce to a `String` if not falsy. You've got to opt-in anyways so the consumer may as well choose an event name. That way when reading the definition, it's clear to the reader what's going on, rather than making then have knowledge of what Skate's default is.

#### Property option name

What should the property option name be? I vote for either `event` or `notify`.

### IRL decisions:

1. The event should not bubble - performance and accuracy (a bubbled event could fire on an element who's property didn't actually change.
2. The event should be cancelable - allows for truly stateless components.

PR #521, opened 2016-03-22
#513 Prop change event

jpnelson:


src/lifecycle/properties-init.js
```diff
@@ -17,6 +19,21 @@ function getLinkedAttribute (name, attr) {
   return attr === true ? dashCase(name) : attr;
 }
 
+function triggerPropertyChangedEvent (elem, {propertyName, eventName, oldValue, newValue}) {
+  let propertyChangeEvent = createCustomEvent(eventName, {
```
treshugart: `const`

src/lifecycle/properties-init.js
```diff
@@ -17,6 +19,21 @@ function getLinkedAttribute (name, attr) {
...
+    bubbles: false,
+    cancelable: true,
+    detail: {
+      name: propertyName,
```
treshugart: You can just just do:

```js
{ name, oldValue, newValue }
```

Where `name` is `propertyName`.

README.md
```diff
@@ -461,6 +461,22 @@ The parameters passed to the function are:
...
+##### `event`
+
+An event name to trigger whenever the property changes. This event is cancelable, and does not bubble.
+
+```js
+skate('my-component', {
+  properties: {
+    myProp: {
+      event: 'my-prop-changed'
+  }
```
treshugart: Needs closing `}` for the `myProp` definition.

src/api/emit.js
```diff
@@ -1,25 +1,8 @@
 import createElement from '../native/create-element';
-import createEvent from '../native/create-event';
+import createCustomEvent from '../util/create-custom-event';
```
treshugart: Is there any reasoning for separating this out and not just using `emit()` directly? I guess you don't need all the behaviour that emit provides. More curious that anything else.

src/lifecycle/properties-init.js
```diff
@@ -17,6 +19,21 @@ function getLinkedAttribute (name, attr) {
...
+      oldValue: oldValue,
+      newValue: newValue
+    }
+  });
+
+  const cancelled = !dispatch(elem, propertyChangeEvent);
```
treshugart: Could just return here.

test/unit/api/properties.js
```diff
@@ -20,6 +20,70 @@ function testTypeValues (type, values) {
 }
 
 describe('api/properties', function () {
+  describe.only('property change event', function () {
```
treshugart: `.only` ;)

test/unit/api/properties.js
```diff
@@ -20,6 +20,70 @@ function testTypeValues (type, values) {
...
+
+    it('contains property name, and change details', function () {
+      let elem = create(assign({event: 'propertychange'}, properties.boolean()));
+      let event = null;
+      elem.test = false;
+
+      elem.addEventListener('propertychange', (e) => event = e);
+      elem.test = true;
+
+      expect(event).to.not.equal(null);
```
treshugart: Maybe `to.be.an('object')` is a more precise assertion here.

jpnelson:
@treshugart it seems like there's a problem with the saucelab build – I've updated the PR and the tests are now passing locally, but Chrome is failing to launch (on master too)

treshugart:
@jpnelson I just restarted the builds. Definitely seems to be a Sauce Labs issue. We've been having some trouble with them.

test/unit/api/properties.js
```diff
@@ -20,6 +20,70 @@ function testTypeValues (type, values) {
...
+      let fooCalled = false;
+      elem.addEventListener('propertychange', () => propertyChangeCalled = true);
+      elem.addEventListener('foo', () => fooCalled = true);
+      elem.test = true;
+
+      expect(propertyChangeCalled).to.equal(false);
+      expect(fooCalled).to.equal(true);
+    });
+
+    it('is triggered only when properties change', function () {
```
treshugart: Should probably also have a test to make sure it doesn't bubble.

src/lifecycle/properties-init.js
```diff
@@ -150,6 +161,18 @@ function createNativePropertyDefinition (name, opts) {
       newValue = opts.coerce(newValue);
     }
 
+    const propertyHasChanged = newValue !== oldValue;
+    if (propertyHasChanged && opts.event) {
+      const cancelled = triggerPropertyChangedEvent(this, {
```
treshugart: Is it worth having to call out to another function for this? I don't see it abstracting much.

treshugart:
lgtm

joscha:
LGTM - As a side note: I'd like to propose we replace the CustomEvent impl. with the one from AUI in the longterm.

Issue #508, opened 2016-03-15
Change resolved to defined, remove unresolved


treshugart:
As per the recent spec change: https://github.com/w3c/webcomponents/issues/418.

This issue involves:

1. Rename `resolved` to `defined` in the component definition, both for the option name and value.
2. Remove `unresolved` from the options. Users can use `:not([defined])` instead.
3. This will involve adding a separate flag for an element that has been "rendered" since that reuses the old `resolved` attribute for assisting in server-side rendering, or skipping the `render` callback if the flag is present.

In native upgrading is a nanotask, so between an element becoming defined and actually upgraded, there isn't an observable delay in JS. However, in polyfill land, a mutation observer must fire so there's definitely an observable delay. Should we have a separate flag for detecting when an element is finally upgraded, or should we just use `[defined]` as is for polyfill? This should be ok because it won't be added to the element until Skate upgrades it anyways. Maybe the only difference is that we add this earlier on in the element upgrading algorithm rather than later.

PR #537, opened 2016-04-28
#508 - Rename resolved to defined and add flag for rendered.

treshugart:
Implements #508.

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -6,14 +6,14 @@ export default {
   attached: nope,
   attribute: nope,
   created: nope,
+  definedAttribute: 'defined',
   render: nope,
   detached: nope,
   events: nope,
   extends: nope,
   properties: nope,
   prototype: {},
-  resolvedAttribute: 'resolved',
   ready: nope,
-  type,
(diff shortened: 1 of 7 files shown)


nadiam84:
We have styles that depend on "resolve" attribute, won't this affect them?

treshugart:
@nadiam84 yeah we do. Unfortunately we're just going to have to update those. Since this is a change in the actual spec I'd rather stay as inline as we can at the cost of some selector renames.

nadiam84:
if "resolved" transformed to "defined" then why
`if (render && !resolved) {`
transformed to
`if (render && !this.hasAttribute(renderedAttribute)) `
and not to this
`if (render && !this.hasAttribute(definedAttribute))`

treshugart:
That was part of the issue; I think the third bullet point.

nadiam84:
lgtm

treshugart:
But that reminds me I don't think we need to add the rendered attribute, I
think we only need to check for its existence as a server renderer would
specifically add this after it renders it out.

Issue #541, opened 2016-05-05
Merge in Kickflip to Skate proper


treshugart:
This will make Skate opinionated about rendering and what it uses to do so:

- Named slots is polyfilled (possibly can use v0 shadow DOM in Blink)
- Incremental DOM is the defacto virtual DOM implementation.

Pros:

1. Skate puts less of the onus on users to decide the best way to do things.
2. You can still use any templating language you want as long as it compiles down to Incremental DOM.
3. We push for functional web components.

Cons:

1. It gets a little bigger.
2. People can no longer choose how they want to render things, but this could be alleviated by being able to choose your own templating language if it compiles down to Incremental DOM.

joscha:
I think if it increases the consumeability of skate without taking away actual widely used features (e.g. just adding a bit of opinionated flavour) it is a good idea. Anyone wanting to get into the topic fast would use Skate+kickflip anyway, so merging them would make it easier. The only thing I am a little concerned about is the amout of polyfilling - skate currently only fills tiny amounts and I think that is actually an advantage of skate by itself.

treshugart:
The amount of polyfilling is a temporary issue I'm willing to risk at this point for the sake of simplicity and more widespread adoption.

treshugart:
@marcoms You'd probably have an opinion here.

@antitoxic you've been vocal in the past about Skate not having an opinionated way to render. This would fix that. Input is appreciated.

marcoms:
I must admit I don't have experience with functional approaches like React uses. One thing I've been wary of is how animation would be possible between two different DOM states (which is why I have in the past been interacting with the DOM directly on property changes instead of re-rendering). I believe Aurelia automatically animates layout changes but something like that would probably be non-trivial and require more thought.

treshugart:
@marcoms that's a very valid concern. There can be issues with this. For example, if 

1. a render occurs during the middle of an animation and your animation is based on changes to the `style` attribute that Incremental DOM isn't aware of, or
2. the position of the animating element changes relative to its siblings and there is no `key` defined ,it would be blown away and re-rendered

For 1, there are a couple of work arounds. Perhaps you could simply use the `render()` function to animate the style property similar to how react-motion does things:

```js
skate('my-element', {
  properties: {
    animationWidth: skate.properties.number()
  },
  render (elem) {
    <div style={ width: elem.animationWidth } />
  }
});
```

Or you could simply tell Incremental DOM not to diff the `style` attribute using the `statics` prop on an element:

[...]

PR #542, opened 2016-05-06
#541 (1) - Refactor properties.

treshugart:
This PR removes all of the redundant options and code since we're making Skate a little bit more opinionated. This also simplifies the property lifecycles since it now syncs attribute to property from the `attributeChangedCallback` in v0 and whatever they're calling it now in v1.


src/lifecycle/created.js
```diff
@@ -27,9 +23,23 @@ function ensurePropertyDefinitions (elem, propertyFunctions) {
   }, {});
 }
 
+function iniitaliseProperties (elem, propertyDefinitions) {
```
joscha: just a small thing, but there us a common typo :)

src/lifecycle/properties-init.js
```diff
@@ -4,147 +4,90 @@ import data from '../util/data';
...
-function getData (elem, name) {
-  return data(elem, `api/property/${name}`);
+function getDefaultValue (elem, name, opts) {
+  return typeof opts.default === 'function' ? opts.default(elem, { name }) : opts.default;
 }
 
-function getDataForAttribute (elem, name) {
-  return getData(elem, getData(elem, name).linkedProperty);
+function getInitialValue (elem, name, opts) {
+  typeof opts.initial === 'function' ? opts.initial(elem, { name }) : opts.initial;
```
nadiam84: no "return"?

joscha: 👍  I think there might be a test missing as well then :)

treshugart: Added some tests. It required a tiny bit of refactoring in the test to accommodate `initial` tests, but happy they got written. Totally my oversight they didn't get in initially (no pun intended).

src/lifecycle/created.js
treshugart: Fixed.

PR #543, opened 2016-05-06
#541 (2) - Support only custom elements to the spec.

treshugart:


src/lifecycle/created.js
```diff
@@ -27,13 +24,26 @@ function ensurePropertyDefinitions (elem, propertyFunctions) {
   }, {});
 }
 
+function iniitaliseProperties (elem, propertyDefinitions) {
```
joscha: function seems duplicated and also conains a typo.

treshugart: That got fixed in #542.

PR #544, opened 2016-05-10
#541 (3) - "merge" in Kickflip functionality

treshugart:


treshugart:
Hey @nadiam84, seems Safari >=8 are still failing for me on this, which is to be expected until we drop support for them.

IE >= 9 tests are failing all with the same error. I'm thinking this is due to `__lookupGetter__` and `__lookupSetter__` taking precedence, but unsure. Maybe you'd have some insight here since you were doing some testing today with IE.

Thoughts?

treshugart:
Whoops, sorry @nadiam84, that comment was meant for named-slots. However, the remaining build issues seem directly related to the same reasons https://github.com/skatejs/named-slots/pull/63 is failing. Maybe we can pair on that tomorrow.

package.json
```diff
@@ -22,7 +22,9 @@
   "devDependencies": {
     "eslint": "2.8.0",
     "gulp": "gulpjs/gulp#73eced0",
-    "skatejs-build": "skatejs/build#22dc0e2",
+    "incremental-dom": "^0.3.0",
```
joscha: are we deliberately not fixing the version here?

package.json
```diff
@@ -22,7 +22,9 @@
...
+    "skatejs-build": "skatejs/build#e25d583fea5f3fc590aed2ee274d37daebf279a2",
+    "skatejs-named-slots": "skatejs/named-slots#00daacd",
```
joscha: For a 1.0 it would be cool to also have a named-slots release instead of a sha I think

src/api/create.js
```diff
@@ -1,11 +1,30 @@
...
 import init from './init';
-import registry from '../shared/registry';
+import support from '../native/support';
 
 export default function (name, props) {
-  const Ctor = registry.get(name);
-  const elem = Ctor ? Ctor.type.create(Ctor) : createElement(name);
-  Ctor && init(elem);
+  let elem;
+  const Ctor = customElements.get(name);
```
joscha: This feels like I have seen it before in PR (2)

src/lifecycle/created.js
```diff
@@ -27,13 +24,26 @@ function ensurePropertyDefinitions (elem, propertyFunctions) {
   }, {});
 }
 
+function iniitaliseProperties (elem, propertyDefinitions) {
```
joscha: duplicated method, with typo again.

package.json
treshugart: Good point. I'll check the version and fix.

package.json
treshugart: Yep, I'll add that to #549.

package.json
treshugart: Fixed.

package.json
treshugart: Added.

src/api/create.js
treshugart: You would have. Each PR was branched off the previous ones since the work is tightly coupled but I still wanted to break them up into chunks. Once we start merging in order we'll see less files in the diffs.

src/lifecycle/created.js
treshugart: Yep, fixed in #542.

test/unit/api/properties.js
```diff
@@ -77,7 +77,7 @@ describe('api/properties', function () {
         let elem = create(properties.boolean());
         elem.setAttribute('test', value);
```
nadiam84: This test doesn't check what it suppose to check, I think. It's not in the changes, it has been here before, but still.
When we are doing `[undefined, null, false, 0, '', 'something'].forEach(` the first thing that we do is `value = JSON.stringify(value);`, which doesn't make sense for the test here because all that we check now is setting string for an attribute, not the different zero-values.

treshugart: Can you add an issue for that and we can fix as a separate PR?

nadiam84: done: https://github.com/skatejs/skatejs/issues/550

Issue #547, opened 2016-05-12
Remove skate.render()


treshugart:
Since becoming more opinionated about the rendering pipeline, there is no longer a need to imperatively render a component.

treshugart:
That said, there may be a possibility that this would warrant use in a test where you wanted to synchronously render a component. An alternative to that would be to use the [`state()`](https://github.com/skatejs/skatejs#state-elem-state) function and force a synchronous render in that. The question then being, are there any use-cases for keeping `render()`?

seancurtis:
I prefer the 3rd option. Adding setTimeouts is messy and could be brittle. Having to `render()` on some state changes, but not others might end up with people just calling it "in case". The third option hides the entire issue from them and should result in a lot cleaner test files.

treshugart:
Third option implemented in #556.

PR #556, opened 2016-05-16
#547 - Remove skate.render() and sync render in skate.state()

treshugart:
Implements option three from #547.

This currently does not cater to the case where you may set properties around a call to `state()`:

```js
elem.someProp = true;
state(elem, { someOtherProp: true });
elem.someOtherOtherProp = true;
```

In that case, `someProp` and `someOtherOtherProp` sets would cause a render to happen async after the call to `state()` while the call to `state()` also invokes a render. In the case above there would be two render calls: one for the property sets and one for the call to `state()`.

I don't see it worth catering to this use case at this point in time because one would either want to use `state()` or set properties but not both. It's likely also state would be used in apps where they have access to the `state()` function that corresponds to the version of Skate they are using. In environments where Skate is used to distribute web components to end users, `state()` likely wouldn't be used because the consumers would just set properties.


lukebatchelor:
LGTM 👍

Issue #511, opened 2016-03-16
Add support for observedAttributes


treshugart:
https://github.com/w3c/webcomponents/issues/367

```js
skate('my-element', {
  attributes: ['my-attribute']
});
```

Another way of specifying these is finding which attributes are linked to properties in the existing API:

```js
skate('my-element', {
  properties: {
    myAttribute: { attribute: true }
  }
});
```

treshugart:
This is "breaking" because when this is implemented by browsers, only those attributes in the whitelist here will be observed causing the `attributeChangedCallback` callback to not fire for some that it may have already been firing for.

Issue #483, opened 2016-02-16
Changing of behaviour for attributeChanged


treshugart:
The current behaviour is to not fire for existing attributes. See: https://github.com/skatejs/skatejs/blob/master/README.md#attribute-1. However, they've recently decided on some changes.

However, there have been several issues raised that are currently open that we should keep track of and action once they've been closed:

- https://github.com/w3c/webcomponents/issues/364 and https://github.com/w3c/webcomponents/issues/366
- https://github.com/w3c/webcomponents/issues/367

If this list grows, we should split these out but until it's final this may serve as a better aggregation point.

This is also dependent upon #511.

PR #557, opened 2016-05-16
#483 - Ensure attribute lifecycle is invoked for attributes existing on the element at the time of upgrade

treshugart:
Implements #483, #511 and #559.


src/lifecycle/properties-init.js
```diff
@@ -46,6 +46,11 @@ function createNativePropertyDefinition (name, opts) {
     const attributeName = getLinkedAttribute(name, opts.attribute);
     let initialValue = elem[name];
 
+    const observedAttributes = elem.constructor.observedAttributes;
```
treshugart: It'd be ideal to do this up front when the element is registered rather than checking and adding when the properties are initialised for each element. The difficulty was refactoring linked attribute names to be calculated as soon as possible. This can be done, but requires a little more effort than this right here. I'd like to do that before merging this PR.

src/lifecycle/created.js
```diff
@@ -24,6 +27,18 @@ function ensurePropertyDefinitions (elem, propertyFunctions) {
   }, {});
 }
 
+function callAttributeChangedForEachAttribute (elem, observedAttributes, definedAttribute) {
+  observedAttributes.forEach(function (name) {
+    const attr = elem.attributes[name];
+
+    // We don't call it for the defined attribute because that will have
+    // already called the handler via setAttribute().
+    if (attr && name !== definedAttribute) {
```
treshugart: I believe if you set an observed attribute in `created` that it'd fire the `attributedChanged` callback. I'm not sure what the behaviour for native should be here in V1. I'm unclear as to what the spec means here. Should this fire even if `attributeChanged` hasn't been called for existing properties? I'd think not, but then again you never know.

treshugart: This was fixed. Massive hacks for Chrome here, though.

lukebatchelor:
LGTM 👍

Issue #566, opened 2016-05-25
Problems with iDOM default attribute setter


treshugart:
The check for `name in elem` will set properties for svg elements that have only getters. To rectify this, we should:

1. remove that check and move it into the event setter so that it can set events as props if available
2. remove the need for internalData as we can just check `elem.constructor.properties`

A test should be written for creating an SVG element and setting `height` which has only a getter.

treshugart:
This is happening again. Incoming fix.

PR #567, opened 2016-05-25
#565 and #566  - Fix iDOM props setter and clean up how props are set

treshugart:
Fixes both #565 and #566.

### DIFF
--- a/test/unit/vdom/incremental-dom.js
+++ b/test/unit/vdom/incremental-dom.js
@@ -12,13 +12,13 @@ describe('IncrementalDOM', function () {
 describe('VdomIncrementalDOM', function () {
   describe('attributes (default)', function () {
     it('should not set properties on SVG elements', function () {
-      // This should throw but the error is different depending on the browser,
-      // so we don't expect any particular message.
-      element().skate({
-        render () {
-          vdom('svg', { height: 100 });
-        }
-      })();
+      expect(function () {
+        element().skate({
+          render () {
+            vdom('svg', { height: 100 });
+          }
+        })();
+      }).to.not.throw(Error);
     });
 
     it('should set built-in event handlers as properties', function () {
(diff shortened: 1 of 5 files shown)


test/unit/vdom/incremental-dom.js
```diff
@@ -1,8 +1,34 @@
...
   it('should export all the same members as the incremental-dom we consume', function () {
     expect(VdomIncrementalDOM).to.contain(IncrementalDOM);
   });
 });
+
+describe('VdomIncrementalDOM', function () {
+  describe('attributes (default)', function () {
+    it('should not set properties on SVG elements', function () {
+      // This should throw but the error is different depending on the browser,
+      // so we don't expect any particular message.
```
nadiam84: If we're expecting an error everywhere, why not check it with `expect.to.throw` ?
```
expect(function() {
          //... func body ...
        }).to.throw(Error);
```

treshugart: I wasn't aware you could specify the instance type. Nice. Will do.

test/unit/vdom/incremental-dom.js
```diff
@@ -12,13 +12,13 @@ describe('IncrementalDOM', function () {
...
-        }
-      })();
+      expect(function () {
+        element().skate({
+          render () {
+            vdom('svg', { height: 100 });
+          }
+        })();
+      }).to.not.throw(Error);
     });
```
nadiam84: Confused a bit.
Was: 'This should throw but the error is different depending on the browser', i.e. we are expecting an error.
Now: `.to.not.throw(Error);`, i.e. we are not expecting an error.
Which one is correct?

treshugart: That was a bad comment that should have had a "not" in between "should" and "throw" :).

nadiam84: it says 'should not set properties on SVG elements', but you actually checking whether or not initialization went well. Shouldn't you check for existence of an actual property as well?

LGTM otherwise

nadiam84:
LGTM

test/unit/vdom/incremental-dom.js
treshugart: The property would be set to the same value the attribute had been set to by just setting the attribute (the property is a getter for the attribute).

PR #568, opened 2016-05-25
Edge existing attr mutation poc

bengummer:
- just Edge. Chrome, IE, FF all fine.
- Plus a hacky lifecycle fix for AUI but still need to fix failing test

The new test is the only one that fails on Edge - it describes an existing element with an existing attribute. I may be missing something in the skate definition but the failing test is consistent with the failing tests i'm seeing in AUI 5.9.x which is based on the same 0.13.x skate branch (but please let me know the definition needs extra config)

### DIFF
--- a/lib/lifecycle.js
+++ b/lib/lifecycle.js
@@ -40,15 +40,6 @@
     return nativeMatchesSelector.call(element, selector);
   };
 
-  // Edge MutationObserver is polyfilled because Edge has a bug where the
-  // MutationRecord provided when removeAttribute is called has an incorrect
-  // oldValue of null rather than the previous attribute value. This Edge
-  // polyfill can be removed when http://jsbin.com/nirofo works.
-  var isEdge = /Edge/.test(navigator.userAgent);
-  if (isEdge && hasOwn(window, "JsMutationObserver") && !window.JsMutationObserver._isPolyfilled) {
-    window.MutationObserver = window.JsMutationObserver;
-  }
-
   /**
    * Parses an event definition and returns information about it.
    *
--- a/lib/lifecycle.js
+++ b/lib/lifecycle.js
@@ -40,6 +40,15 @@
     return nativeMatchesSelector.call(element, selector);
   };
 
+  // Edge MutationObserver is polyfilled because Edge has a bug where the
+  // MutationRecord provided when removeAttribute is called has an incorrect
+  // oldValue of null rather than the previous attribute value. This Edge
+  // polyfill can be removed when http://jsbin.com/nirofo works.
+  var isEdge = /Edge/.test(navigator.userAgent);
+  if (isEdge && window.hasOwn("JsMutationObserver") && !JsMutationObserver._isPolyfilled) {
+    window.MutationObserver = window.JsMutationObserver;
+  }
+
   /**
    * Parses an event definition and returns information about it.
    *
--- a/lib/lifecycle.js
+++ b/lib/lifecycle.js
@@ -45,7 +45,7 @@
   // oldValue of null rather than the previous attribute value. This Edge
   // polyfill can be removed when http://jsbin.com/nirofo works.
   var isEdge = /Edge/.test(navigator.userAgent);
-  if (isEdge && window.hasOwn("JsMutationObserver") && !JsMutationObserver._isPolyfilled) {
+  if (isEdge && hasOwn(window, "JsMutationObserver") && !JsMutationObserver._isPolyfilled) {
     window.MutationObserver = window.JsMutationObserver;
   }
 
--- a/lib/lifecycle.js
+++ b/lib/lifecycle.js
@@ -45,7 +45,7 @@
   // oldValue of null rather than the previous attribute value. This Edge
   // polyfill can be removed when http://jsbin.com/nirofo works.
   var isEdge = /Edge/.test(navigator.userAgent);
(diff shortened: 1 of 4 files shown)


bengummer:
I looked into potentially porting back the `removeAttribute` patches from master but the issue causing the bug is actually in Edge's `MutationObserver` the solution in this PR is the way to go it seems.

bengummer:
This is ready for PR now.

Tested this against sauce (this skate branch and also in AUI 5.9.x).

Tried to backport the `removeAttribute` patches from master to avoid having to override Edge's `MutationObverserver`, but no luck since it's a MutationObserver bug in Edge.

I'll run `npm run dist` just before merge to keep the PR clean.

lib/lifecycle.js
```diff
@@ -40,6 +40,15 @@
     return nativeMatchesSelector.call(element, selector);
   };
 
+  // Edge MutationObserver is polyfilled because Edge has a bug where the
```
treshugart: Just wondering if this is a known edge bug. If not it'd be worth reporting it.

lib/lifecycle.js
```diff
@@ -40,6 +40,15 @@
...
+  // MutationRecord provided when removeAttribute is called has an incorrect
+  // oldValue of null rather than the previous attribute value. This Edge
+  // polyfill can be removed when http://jsbin.com/nirofo works.
+  var isEdge = /Edge/.test(navigator.userAgent);
+  if (isEdge && hasOwn(window, "JsMutationObserver") && !window.JsMutationObserver._isPolyfilled) {
```
treshugart: Nitpick: single quotes

src/lifecycle.js
```diff
@@ -34,6 +34,15 @@ var matchesSelector = function (element, selector) {
   return nativeMatchesSelector.call(element, selector);
 };
 
+// Edge MutationObserver is polyfilled because Edge has a bug where the
```
treshugart: Woop, see comments on `lib/` files above.

treshugart:
Have you considered checking to see if we could manually store the previous value in https://github.com/skatejs/skatejs/blob/0.13.x/src/lifecycle.js#L122 and use that instead of the mutation records value?

bengummer:
@treshugart thanks for the PR feedback, i've rolled back the Edge MutationObvserver override in favour of attribute caching (Edge only for now for performance reasons but other browsers can be added easily in the future if needed).

Raised a ticket for the Edge bug at https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7711883/

src/lifecycle.js
```diff
@@ -34,6 +34,11 @@ var matchesSelector = function (element, selector) {
   return nativeMatchesSelector.call(element, selector);
 };
 
+// Edge has a bug where oldValue is sent as null instead of the true oldValue
+// when an element attribute is removed. Bug raised at
+// https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7711883/
+const needsAttrCaching = /Edge/.test(navigator.userAgent);
```
treshugart: In case anyone is wondering, to feature check this would take creating a mutation observer for an element add an attribute, then remove the attribute and then check the old value. Since mutation observers are run at the end of microtasks it could make the check flaky if other component code on the page is executed before the check.

src/lifecycle.js
```diff
@@ -124,7 +129,14 @@ function triggerAttributeChanged(target, component, data) {
   var type;
   var name = data.name;
   var newValue = data.newValue;
-  var oldValue = data.oldValue;
+
+  // Read the old attribute value from cache if needed, otherwise use native oldValue
+  var cachedAttributes;
+  if (needsAttrCaching) {
+    cachedAttributes = hasOwn(target, '_cachedAttrs') && target._cachedAttrs;
```
treshugart: Instead of using a property on the element, you can use the imported `data()` function already in this scope. Something like:

```js
var cachedAtributes = data(target, 'cachedAttributes');
```

It creates an object if one doesn't already exist so you don't need to create the object if it doesn't exist in `addAttributeListeners()`.

treshugart: Nitpick: something that describes the problem for the data name might be better. Naming is hard, though. `cachedAttributeOldValues`, `cachedOldValues`, `attributeOldValues`. On it's own, I'm not really sure what `cachedAttributes` really means.

src/lifecycle.js
```diff
@@ -138,6 +150,14 @@ function triggerAttributeChanged(target, component, data) {
     type = 'removed';
   }
 
+  if (needsAttrCaching) {
+    if (type === 'created' || type === 'updated') {
```
treshugart: You might be able to invert this check and simplify it:

```js
if (type === 'removed') {
  delete cachedAttributes[name];
} else {
  cachedAttributes[name] = newValue;
}
```

src/lifecycle.js
```diff
@@ -188,6 +208,10 @@ function triggerAttributesCreated (target, component) {
 }
 
 function addAttributeListeners (target, component) {
+  if (needsAttrCaching) {
```
treshugart: This won't be necessary if you use `data()` in the `triggerAttributeChanged()` function.

bengummer:
Hey @treshugart good tip on the `data()` function :)

I've switched to using data() and addressed the other PR comments. Switching to use data() caused a separate test to break but it just needed a `helpers.afterMutations()`.

treshugart:
lgtm

lukebatchelor:
LGTM 👍

PR #570, opened 2016-05-25
#569 - Fix discrepancy between vdom('slot') and vdom.slot() variants

treshugart:


lukebatchelor:
LGTM 👍

test/unit/vdom/elements.js
```diff
@@ -0,0 +1,34 @@
+import element from '../../lib/element';
+import { symbols, vdom } from '../../../src/index';
+
+const shadowDomV0 = !!document.createElement('div').createShadowRoot;
```
nadiam84: What would happen here in Chrome if we want to use named slots with skate?
It would be `true`, since it's native property in Chrome, but everything would be polyfilled by namedslots because we are using expression `if (!('attachShadow' in document.createElement('div'))) {`and there is no native `attachShadow` in there. But we are using `slot` insead of `content`, so everything would be broken.

treshugart: In this case it would just use Chrome's native implementation. Underneath the hood it would still write a `<content>` element when the user outputs a slot using `vdom('slot')` or `vdom.slot()`. I see your point, though. In terms of performance, trying to use a native implementation first will be better than using a polyfill. Maybe it's worth inverting the check to use a v1 flag, but then it's likely Chrome would use a polyfill when it has native support, even if it's v0. Thoughts?

nadiam84: What if we just do this in namedslots? `if (!('attachShadow' in document.createElement('div') || 'createShadowRoot' in document.createElement('div')))`

nadiam84: And/or make a polyfill in namedslots which transforms v0 to v1

treshugart: We could check for v1 support first. That might be saner in the long run.

> And/or make a polyfill in namedslots which transforms v0 to v1

Can we do that easily?

treshugart: Update to use v1 over v0 even if it's a polyfill.

test/unit/vdom/elements.js
```diff
@@ -1,7 +1,8 @@
 import element from '../../lib/element';
+import support from '../../../src/native/support';
 import { symbols, vdom } from '../../../src/index';
 
-const shadowDomV0 = !!document.createElement('div').createShadowRoot;
+const shadowDomV0 = !support.shadowDomV1 && support.shadowDomV0;
```
nadiam84: same code here and there `src/api/vdom.js`
`!support.shadowDomV1 && support.shadowDomV0;`
can't you just import it from some common place?
LGTM otherwise

nadiam84:
LGTM

test/unit/vdom/elements.js
treshugart: Updated the teste to test for all support scenarios. The code is now different :)

PR #571, opened 2016-05-27
fix broken skate.prop permalink

marcoms:


treshugart:
lgtm

marcoms:
Odd - the lgtm check hasn't passed 🤔

jpnelson:
lgtm

treshugart:
Yeah not sure what's up with the lgtm plugin. I removed the maintainers file because we have the maintainers team. We should all be on it. I'll have a look.

treshugart:
Looking at this now, we have a `properties` option and now the `prop()` function and namespace. Thinking that the former should become the latter for consistency. Maybe `props` so it's clear that you can define multiple props there.

The backstory in why the namespace was changed is:

1. `prop()` is now a function not just a namespace as it is a factory for creating a custom property type
2. it's easier to type and `prop` pretty much means `property` across the board

marcoms:
I'm in favour of `skate.prop.*` (singular form reconciles well with enum naming schemes in languages like Rust) and the `props` option. For numerous properties it would definitely be more convenient.

Issue #581, opened 2016-05-31
Simplify emit()


treshugart:
We added some code to simulate bubbling in detached elements. We've found over time that this is only useful in testing so we're going to remove this from the codebase.

PR #582, opened 2016-05-31
#581 - Simplify emit()

treshugart:
Implements #581.

The following features were removed as they weren't used very much:

- Simulating bubbled events for elements that aren't in the document
- Space-separated event names
- Array of event names
- ^ thus no more firiing of multiple events, no longer need array return value; now it's `true` or `false`

`emit()` is now a thin layer on top of `CustomEvent` that polyfills it (since it's really small) if it doesn't exist and will default `bubbles` and `cancelable` to `true` since we've found that we use that more often than the alternatives.


src/api/emit.js
```diff
@@ -15,76 +13,14 @@ function createCustomEvent (name, opts = {}) {
   if (CustomEvent) {
     return new CustomEvent(name, opts);
   }
-
-  var e = document.createEvent('CustomEvent');
+  const e = document.createEvent('CustomEvent');
```
joscha: maybe we can re-use what we used in named slots for this at a later point in time:
```
    "custom-event-polyfill": "joscha/custom-event-polyfill#de24bd9",
```

test/unit/api/emit.js
```diff
@@ -8,68 +9,62 @@ describe('api/emit', function () {
...
     triggered = 0;
 
+    // All listeners should increment triggered.
+    // Child should always prevent default.
     child.addEventListener('test', () => ++triggered);
     child.addEventListener('test', e => e.preventDefault());
     parent.addEventListener('test', () => ++triggered);
     disabled.addEventListener('test', () => ++triggered);
+
+    // It's not spec'd whether or not detached elements should bubble. In some
```
joscha: some = browsers?

joscha:
LGTM, eventually replacing the custom poly for `CustomEvent` with the one we use in https://github.com/skatejs/named-slots/blob/master/package.json#L37

treshugart:
Yeah, that's the plan. That said, I'd prefer to bundle it and that implementation is smaller (not by much, though) and working for now and it's doubtful it'll ever need to be "maintained".

joscha:
The one in AUI was broken in regards to defaultPrevented, yours seem to have the same issue as far as I can see. Also, the other one is already in named slots, so technically re-using it will save space.

PR #590, opened 2016-06-03
Readme example added missed colon

dmi3y:
Babel gave error on this, apology if that some new javascript syntax though... like ES2222 no colons in object declarations :P


treshugart:
Hah! Thanks. That would be kinda cool, tbh :)

PR #594, opened 2016-06-15
V1 prep

treshugart:
cc @nadiam84 @jpnelson

### DIFF
--- a/src/lifecycle/render.js
+++ b/src/lifecycle/render.js
@@ -1,5 +1,6 @@
 import { IncrementalDOM } from '../api/vdom';
 import { shadowRoot } from '../api/symbols';
+import { shadowDomV0, shadowDomV1 } from '../util/support';
 
 const { patch } = IncrementalDOM;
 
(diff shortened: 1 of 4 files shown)


src/lifecycle/render.js
```diff
@@ -14,9 +15,9 @@ export default function (Ctor) {
...
       let sr;
 
-      if (elem.attachShadow) {
+      if (shadowDomV1) {
         sr = elem.attachShadow({ mode: 'open' });
-      } else if (elem.createShadowRoot) {
+      } else if (shadowDomV0) {
         sr = elem.createShadowRoot();
       } else {
         sr = elem;
```
nadiam84: It looks a bit ugly, why you don't use `switch/case` instead?

treshugart: Personal style, I guess. I prefer `if else` for small situations. Happy to change if there's a good reason for it.

joscha:
LGTM

PR #620, opened 2016-06-22
#619 - Remove version module and tests.

treshugart:
The build doesn't bump this anymore and it doesn't make sense when consumed via a package manager
because it's in the package manifest already. It only makes sense in a global world which is quickly
going away.

BREAKING CHANGE: remove version module #619

### DIFF
--- a/src/api/version.js
+++ b/src/api/version.js
@@ -1,1 +1,0 @@
-export default '0.15.3';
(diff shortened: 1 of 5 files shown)


jpnelson:
I've always found it useful for debugging in the console (sanity check, is this the version I think it is?) however if it's going to be a headache to maintain, it's not a great loss.

treshugart:
It's mostly because it will always be a custom build step (I think?) to maintain and it really only makes sense when you don't use a package manager.

Issue #603, opened 2016-06-20
vDOM event bindings should fire for bubbled events


treshugart:
This is consistent with react's events when using `onClick` and so forth and is more permissive than restricting to the element itself which allows the user to catch bubbled events as well as restrict them if they want to (i.e. `e.target === e.currentTarget`).

PR #628, opened 2016-06-23
#603 - Events bound via the on* syntax should catch bubbled events

treshugart:
#603


treshugart:
Closing to retarget to master.

Issue #601, opened 2016-06-18
prop.number() should default to 0


treshugart:
It seems safer than the default being `undefined` which used in math becomes `NaN` (but oddly `isNaN(undefined) === true`). Removes having to type:

```js
{ default: 0 }
```

...every time and if you want `undefined` or something else, just do:

```js
{ default: undefined }
```

PR #631, opened 2016-06-24
feat: number prop now defaults to zero

bengummer:
related issue #601

### DIFF
--- a/src/api/prop.js
+++ b/src/api/prop.js
@@ -1,7 +1,7 @@
 import assign from 'object-assign';
 import empty from '../util/empty';
 
-const alwaysZeroIfNotANumberOrNumber = val => isNaN(val) ? 0 : Number(val);
+const alwaysUndefinedIfNotANumberOrNumber = val => isNaN(val) ? undefined : Number(val);
 const alwaysUndefinedIfEmptyOrString = val => empty(val) ? undefined : String(val);
 
 export function create (def) {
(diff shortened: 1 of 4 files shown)


src/api/prop.js
```diff
@@ -26,9 +26,10 @@ export const boolean = create({
 });
 
 export const number = create({
-  coerce: alwaysUndefinedIfEmptyOrNumber,
-  deserialize: alwaysUndefinedIfEmptyOrNumber,
-  serialize: alwaysUndefinedIfEmptyOrNumber
+  default: 0,
+  coerce: alwaysZeroIfNotANumberOrNumber,
```
treshugart: What happens if you don't change `undefined` to `0`? I would think that the default would still be `0`, right?

treshugart:
lgtm

src/api/prop.js
bengummer: That works in every case except where the input value is a string e.g. `'test'`. In that case it would get coerced to `NaN` rather than `0` because `Number('test') == NaN`.

treshugart: Right. Hmm, I wonder what the expected behaviour here would be. `String('test')` really isn't a number.

bengummer: Chatted with @treshugart and we're going to get some extra opinions on whether passing `test` should use the default value (`0`) or use `NaN`. Easy code change either way, just need to decide.

bengummer: Informal poll seems like `myNumericProp.value = 'test'` should return `NaN`, will update PR to behave that way

nadiam84:
lgtm

bengummer:
Fixes #601

Issue #638, opened 2016-06-28
Move all special incremental dom functionality out of vdom.element()


treshugart:
This functionality should be abstracted into the `elementOpen()` and `elementOpenStart()`. vdom.element should act as a shorthand for calling elementOpen() with attrs / props and children as normal.

The behaviour can be described as follows:

1. passing a constructor instead of a tag name: vdom.element(MyElement)
2. abstraction between <content select="selector"> (v0 shadow DOM) and <slot name="slotname"> (v1 shadow DOM)
3. { key: 'unique-key' } automatically setting a key for the element
4. { statics: [ ... ] } automatically setting statics for the element
5. { skip: true } automatically skipping the element

treshugart:
References https://github.com/google/incremental-dom/issues/281.

PR #642, opened 2016-06-30
#638 - Move special VDOM behaviour into Incremental DOM overrides.

treshugart:
Implements #638. This also references https://github.com/google/incremental-dom/issues/281 as referenced in the issue for this. It would be better if we could have built-in support for "components" as described in the [babel-plugin-incremental-dom docs](https://github.com/jridgewell/babel-plugin-incremental-dom#components).

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -1,22 +1,23 @@
 import * as IncrementalDOM from 'incremental-dom';
 import * as skateSymbols from './symbols';
 import { shadowDomV0, shadowDomV1 } from '../util/support';
+import assign from '../util/assign';
 
-// Could import these, but we have to import all of IncrementalDOM anyways so
-// that we can export our configured IncrementalDOM.
 const {
-  applyProp,
   attr,
+  applyProp,
   attributes,
   elementClose,
   elementOpen,
   elementOpenEnd,
   elementOpenStart,
+  elementVoid,
   skip,
   symbols,
-  text
+  text,
 } = IncrementalDOM;
 
+const fallbackToV0 = !shadowDomV1 && shadowDomV0;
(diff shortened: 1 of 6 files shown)


lukebatchelor:
lgtm 👍

src/api/vdom.js
```diff
@@ -1,22 +1,20 @@
...
   skip,
   symbols,
-  text
-} = IncrementalDOM;
+  text,
+} from 'incremental-dom';
+import * as skateSymbols from './symbols';
+import { shadowDomV0, shadowDomV1 } from '../util/support';
 
+const fallbackToV0 = !shadowDomV1 && shadowDomV0;
```
joscha: should this be a global helper?

src/api/vdom.js
```diff
@@ -43,6 +46,13 @@ attributes[symbols.default] = function (elem, name, value) {
     return name in elem ? applyProp(elem, name, value) : applyEvent(elem, name.substring(2), name, value);
   }
 
+  // Set the select attribute instead of name if it was a <slot> translated to 
+  // a <content> for v0.
+  if (name === 'name' && elem.tagName === 'CONTENT') {
+    name = 'select';
+    value = `[slot="${value}"]`;
```
joscha: do we need to encode value?

joscha:
LGTM

src/api/vdom.js
treshugart: I don't think so. They're allowed inside attribute values without escaping them.

src/api/vdom.js
treshugart: I thought about that but decided against it since this is the only pace it's used. Happy to refactor later.

joscha:
LGTM

Issue #646, opened 2016-07-05
Ability to pass functions as tag names


treshugart:
Currently you can pass Skate element constructors as tag names (not raw web component constructors, yet), but you cannot pass a raw function.

They have different outcomes. The Skate custom element constructors convert into a string that is the custom element tag name.

Functions should get executed and passed the attrs / props and children so that they can act as templates. Essentially they become stateless, "private" custom elements that behave like templates, but are just functions and are composable.

JSX would look something like this:

```js
function List (props, chren) {
  return <ul {...props}>{chren()}</ul>;
}

function Item (props, chren) {
  return <li {...props}>{chren()}</li>;
}

<List>
  <Item>item 1</Item>
  <Item>item 2</Item>
</List>
```

This has the greatest impact to templating languages and their ability to compose in functions as tag names, but also can be used by the raw `vdom.element()` function or Incremental DOM functions as they'd be implemented at that level.

```js
vdom.elementOpen(List);
  vdom.elementOpen(Item);
    vdom.text('item 1');
  vdom.elementClose(Item);
  vdom.elementOpen(Item);
    vdom.text('item 1');
  vdom.elementClose(Item);
vdom.elementClose(List);
```

or:

```js
vdom.element(List, () => {
  vdom.element(Item, 'item 1');
  vdom.element(Item, 'item 1');
});
```

jpnelson:
I like it. Looks like https://facebook.github.io/react/docs/reusable-components.html#stateless-functions

treshugart:
The only possible issues with this is memory spikes. If you use this, everything between the opening and closing tag will be deferred until later. Worst-case scenario is that it brings it down to the memory usage of something like React or Virtual DOM that build and then discard large trees.

PR #647, opened 2016-07-05
#646 - Implement ability to pass functions as tag names to all v…

treshugart:
Implements #646.

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -90,80 +90,55 @@ function resolveTagName (tname) {
   return tname;
 }
 
-let buildQueueChren;
-let buildQueueProps;
+const stackChren = [];
+const stackProps = [];
 
-function buildQueueEmpty () {
-  buildQueueChren = null;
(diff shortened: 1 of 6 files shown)


test/unit/vdom/properties.js
```diff
@@ -21,6 +21,16 @@ describe('properties', function () {
     expect(elem[symbols.shadowRoot].firstChild.hasAttribute('test')).to.equal(false);
   });
 
+  it('should not set properties on SVG elements', function () {
```
treshugart: This was a previous test that was just moved to a better spot.

src/api/vdom.js
```diff
@@ -80,44 +82,64 @@ function applyEvent (elem, ename, name, value) {
 
 // Returns the tag name of the element if a custom element constructor was
```
bradleyayers: Might want to update this for the function use-case.

treshugart: Fixed in 030cb2a.

PR #653, opened 2016-07-07
#652 - Test with the stable v0 custom element polyfill and fix tests in it

### LINKED ISSUE #652: Constructors are failing with v0 custom element polyfill
The behaviour between native and polyfilled custom elements v0 behaves differently with constructors. The problem is because native allows you to extend `HTMLElement` and the polyfill does not.

treshugart:
I removed the v1 polyfill as I'd prefer to test in the more common, stable polyfill and create an
issue for testing with both polyfills separately.

Fixes #652.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -13,7 +13,7 @@ module.exports = function (config) {
 
   // Shims for testing.
   config.files = [
-    'node_modules/es6-shim/es6-shim.js',
+    'https://npmcdn.com/webcomponents.js/CustomElements.js'
   ].concat(config.files);
 
   // Ensure mobile browsers have enough time to run.
--- a/package.json
+++ b/package.json
@@ -28,8 +28,7 @@
   "devDependencies": {
     "es6-shim": "0.35.1",
     "eslint": "3.0.1",
-    "skatejs-build": "5.0.1",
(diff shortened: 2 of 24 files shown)


karma.conf.js
```diff
@@ -13,7 +13,7 @@ module.exports = function (config) {
 
   // Shims for testing.
   config.files = [
-    'node_modules/es6-shim/es6-shim.js',
+    'https://npmcdn.com/webcomponents.js/CustomElements.js'
```
joscha: is this a good idea? That makes the tests dependent on network.

joscha:
Opera keeps on dying on SL...maybe retry later.

karma.conf.js
treshugart: You're right, I'll make it an NPM dependency.

treshugart:
Someone also experienced the root of this problem in https://github.com/webcomponents/webcomponentsjs/issues/563. Should probably see if we can land this in the polyfill instead. Will keep this open until we do so.

treshugart:
screen shot 2016-07-11 at 2 27 03 pm

Looks like Opera is actually passing.

treshugart:
Going to merge this as is and add a todo to follow up about possible fixes in the polyfills.

PR #661, opened 2016-07-12
fix: Fixes issue with polyfills overriding native shadowDOM functions

lukebatchelor:


src/api/symbols.js
```diff
@@ -5,3 +5,5 @@ export const props = '____props';
 export const renderer = '____renderer';
 export const rendererDebounced = '____rendererDebounced';
 export const shadowRoot = '____shadowRoot';
+export const v0ShadowDOMProperty = 'createShadowRoot'; // property used to check for shadowDOMv0 support
```
treshugart: Eventually these symbols will actually use `Symbol()`, which you wouldn't want to do with these. I'd suggest just co-locating these in the support file.

src/util/support.js
```diff
@@ -1,5 +1,36 @@
+import isNativeRegex from 'regex-native-function'; // checks if a function is native using regex
```
treshugart: You may not need this. You can check for native v0 by doing `'HTMLContentElement' in window` and v1 by doing `'HTMLSlotElement' in window`.

nadiam84: webcomponents polyfilling HTMLContentElement

lukebatchelor: Yea, we were trying to make sure that the check wont return true if a polyfill had set the property already. Unless there's another reason not to use the regex way?

treshugart: Not really, I was just exploring built-in ways to do this. Just trying to keep deps to a minimum, if possible.

src/util/support.js
```diff
@@ -1,5 +1,36 @@
...
+import { v0ShadowDOMProperty, v1ShadowDOMProperty } from '../api/symbols';
+
 const div = document.createElement('div');
+const isNative = (elem, prop) => isNativeRegex.test(elem[prop]);
+const isPolyfill = (elem, prop) => !!elem[prop] && !isNative(elem, prop);
```
treshugart: Should the regex return false if `elem[prop]` is not defined? If so, this would mean you don't need that first check, right?

nadiam84: Yes, but it also will return `false` if this property was polyfilled.

treshugart: Saw that just after I commented :)

README.md
```diff
@@ -60,6 +60,7 @@ Skate requires Custom Element support. In [browsers that don't support it](http:
 
 Skate prefers v0 support if it detects both as that yields the best performance in browsers that support it natively. If you supply only a v1 polyfill, it will use v0 where supported by native and v1 in other browsers.
 
+Skate will always prefer a native implementation to a polyfill as they provide the best performance. Fo example, if you had native v0 support and v1 support through a polyfill, Skate will use v0. Likewise, if native support was detected for both, Skate would use v1.
```
nadiam84: 'for'

test/unit/util/support.js
```diff
@@ -0,0 +1,49 @@
+import { shouldUseShadowDomV0, shouldUseShadowDomV1, v0ShadowDOMProperty, v1ShadowDOMProperty } from '../../../src/util/support';
+
+describe('util/support', () => {
+  describe('shadowDOM', () => {
+    /* We take a copy of a known native property so that we can mock how our code will behave when 
+       it finds a native implmentation of a function. The polyfilled property just needs to match
```
nadiam84: 'implementation'

README.md
lukebatchelor: Nuh, I'm speaking Gangsta'

test/unit/util/support.js
lukebatchelor: Nuh, I'm speaking... ..

I've got nothing.

nadiam84: :D :D

src/util/support.js
```diff
@@ -1,5 +1,38 @@
+import isNativeRegex from 'regex-native-function'; // checks if a function is native using regex
+
 const div = document.createElement('div');
+const isNative = (elem, prop) => isNativeRegex.test(elem[prop]);
+const isPolyfill = (elem, prop) => !!elem[prop] && !isNative(elem, prop);
+
+export const v0ShadowDOMProperty = 'createShadowRoot'; // property used to check for shadowDOMv0 support
```
treshugart: nit: comments go above associated lines, empty newline should separate comment and code above it, i.e.:

```js
// comment 1
const code1;

// comment 2
const code2;
```

The aim is readability and comment association.

test/unit/util/support.js
```diff
@@ -0,0 +1,49 @@
+import { shouldUseShadowDomV0, shouldUseShadowDomV1, v0ShadowDOMProperty, v1ShadowDOMProperty } from '../../../src/util/support';
+
+describe('util/support', () => {
+  describe('shadowDOM', () => {
+    /* We take a copy of a known native property so that we can mock how our code will behave when
```
treshugart: nit: multiline comments should still use `//`

lukebatchelor: So just use /**/ for headers and jsdoc annotations and stuff?

treshugart:
lgtm

lukebatchelor:
Can I LGTM?

Issue #663, opened 2016-07-13
Multiple components with the same name


treshugart:
To be able to use Skate in Webpack's Hot-module-reloader, we require the ability to be able to register a component multiple times.

This makes the tag name non-deterministic, but that doesn't matter if you can use the constructor.

PR #664, opened 2016-07-13
#564, #663 - Multiple components with the same name

### LINKED ISSUE #564: Issue with webpack HRM
Is there anyone here using webpack Hot Module Replacement successfully with skatejs components? I currently got this issue.
```
[HMR] Cannot apply update. Need to do a full reload!
[HMR] Error: Failed to execute 'registerElement' on 'Document': Registration failed for type 'tiny-button'. A type with that name is already registered.
```

treshugart:
If a component with the same name has already been registered, Skate will use the specified name as  a prefix and append a number to the name before registering it.

Fixes #564 and implements #663.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,8 @@ Without native support and if you do not supply a Shadow DOM polyfill, any compo
       - [Background](#background)
     - [`state (elem[, state])`](#state-elem-state)
     - [`symbols`](#symbols)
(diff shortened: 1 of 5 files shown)


treshugart:
Might be good to only allow this in dev mode for now.

treshugart:
Merged. We can add dev mode later if need be.

Issue #657, opened 2016-07-08
Test with v0 and v1 polyfills separately.


treshugart:
Right now we only test with one of the custom element polyfills. We should test against both. We can't load them in the same Karma session so we'll have to create a separate runs for each polyfill.

treshugart:
Not supporting v0 polyfills' problems. Issues should be taken up with the v0 polyfills respectively as Skate works flawlessly with native v0 implementations. Tests only load polyfills if necessary.

PR #665, opened 2016-07-13
#657 - Test with v0 and v1 polyfills

nadiam84:
Possible solution for this issue: #657
Although I'm not sure what is correct import suppose to be in test/polyfill.v0.js


test/polyfill.v0.js
```diff
@@ -0,0 +1,4 @@
+import 'webcomponents.js/src/ShadowDOM/ShadowDOM';
+import 'webcomponents.js/src/WeakMap/WeakMap';
+import 'webcomponents.js/src/MutationObserver/MutationObserver';
+import 'webcomponents.js/src/CustomElements/v1/CustomElements';
```
treshugart: Should probably use the v0 custom element polyfill here.

treshugart: You'll also want to check to see if you need those other polyfills with it (I just remember possibly not needing them).

karma.conf.js
```diff
@@ -11,10 +13,19 @@ module.exports = function (config) {
     config.browsers.push('internet_explorer_11');
   }
 
+  config.preprocessors['test/polyfill.v1.js'] = [ 'webpack' ];
```
treshugart: Can these just be concatenated into the `files` array?

nadiam84:
@treshugart updated the PR, lots of broken tests, can you take a look?

karma.conf.js
```diff
@@ -12,20 +10,28 @@ module.exports = function (config) {
     // Only test IE latest.
     config.browsers.push('internet_explorer_11');
   }
+  var polyOptions = process.argv.filter((val) => {
```
treshugart: Do you think it'd be worth bringing something in like [`minimist`](https://github.com/substack/minimist)?

.travis.yml
```diff
@@ -11,10 +11,11 @@ before_script:
 script:
   - sk-bundle
   - if [ "${TRAVIS_SECURE_ENV_VARS}" = "true" ]; then
-    npm test -- --saucelabs --v0;
-    npm test -- --saucelabs;
+    npm test -- --saucelabs --polyfills=dom0,elem0;
```
treshugart: This is awesome! Total coverage on integrations.

karma.conf.js
nadiam84: Definitely if we're going to add more logic to this. If not, I think it would be a little overkill. Reminds me a bit of this http://m.imgur.com/Q3mkcnl?r

treshugart: I get it, it would just turn several LoC into a single one and normalise how the argument is passed which makes passing args a little bit less brittle.

karma.conf.js
```diff
@@ -10,6 +10,30 @@ module.exports = function (config) {
...
+    return val.search('--polyfills=') == 0;
```
treshugart: nit: `val.indexOf()` is probably more appropriate because we don't require a regex (also `===`)

karma.conf.js
nadiam84: Ok, I'll fix it

treshugart:
LGTM pending fixes for broken tests. Happy to pair if on those if that helps.

nadiam84:
Disabled CustomElementV1 testing
Fixed one broken test (separate PR #672 since it's unrelated to this issue)

treshugart:
Closing this now since we've decided to simplify the polyfill support matrix. We've listed know issues with v0 polyfills (works in native v0) and if there's any problems there we can raise issues with the polyfill projects.

Issue #666, opened 2016-07-13
Support ref callbacks


treshugart:
Ref callbacks can be specified on any element:

```js
const ref = input => input.focus();
vdom.element('input', { ref });
```

The ref should:

- only be called for elements (not possible on text nodes anyways)
- be called when the element is closed so that:
  - it has access to all attributes / properties
  - it has access to all descendants of the ref element
  - basically it has access to the current state of the element when it is closed

As noted in the React docs, they're discouraging ref strings in favour of ref callbacks. See: https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute. We never planned on doing strings, so I think this is a testament that the callback API should be the only one for now.

PR #668, opened 2016-07-13
#666 - Support ref() callbacks passed as props

treshugart:
Implements #666.


nadiam84:
Don't see any changes in the Readme.md file?

nadiam84:
LGTM other than Readme.md file

treshugart:
Nice catch, will update.

treshugart:
This actually passes on iPhone. Sauce Labs is failing because of their infrastructure and I don't have an ETA, and we can't just wait for them, so merging.

PR #672, opened 2016-07-17
fix: fixed native polyfill detection

nadiam84:
Changed to the `Function.prototype` in `nativeProperty`, since webcomponents shadowDOM polyfill polyfillng everything and old approach didn't work as expected.

`shouldUseShadowDomV0` and `shouldUseShadowDomV1 ` became simpler.

### DIFF
--- a/test/unit/util/support.js
+++ b/test/unit/util/support.js
@@ -5,7 +5,7 @@ describe('util/support', () => {
     // We take a copy of a known native property so that we can mock how our code will behave when
     // it finds a native implementation of a function. The polyfilled property just needs to match
     // what a non-native property looks like on an element (must have a `value` function)
-    const nativeProperty = Object.getOwnPropertyDescriptor(Element.prototype, 'getAttribute');
+    const nativeProperty = Object.getOwnPropertyDescriptor(Function.prototype, 'toString');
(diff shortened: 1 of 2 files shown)


treshugart:
lgtm

test/unit/util/support.js
```diff
@@ -5,7 +5,7 @@ describe('util/support', () => {
     // We take a copy of a known native property so that we can mock how our code will behave when
     // it finds a native implementation of a function. The polyfilled property just needs to match
     // what a non-native property looks like on an element (must have a `value` function)
-    const nativeProperty = Object.getOwnPropertyDescriptor(Element.prototype, 'getAttribute');
+    const nativeProperty = Object.getOwnPropertyDescriptor(Function.prototype, 'toString');
```
jpnelson: What was the unexpected behaviour? Is `Element.getAttribute` different now?

nadiam84: If we're using webcomponents to polyfill ShadowDOM, then everything gets wrapped because they're re-assigning the prototype for everything starting from the EventTarget.
Because of this attempt to get a native property from the Element results in a property that is not native at all but just a wrapper-function.

PR #674, opened 2016-07-18
fixed registerUniqueName and added a few tests

nadiam84:
fixed registerUniqueName and added a few tests to prove the fix

### DIFF
--- a/src/api/define.js
+++ b/src/api/define.js
@@ -102,9 +102,9 @@ function generateUniqueName(name) {
     registry[name] = true;
     return name;
   }
-  // copy-pasted from here, looks like a good solution : http://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript/2117523#2117523
-  const rand = 'xxxxxxxx'.replace(/[xy]/g, function(c) {
-    var r = Math.random()*16|0, v = c == 'x' ? r : (r&0x3|0x8);
+  // http://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript/2117523#2117523
+  const rand = 'xxxxxxxx'.replace(/[xy]/g, (c) => {
+    const r = Math.random()*16|0, v = c == 'x' ? r : (r&0x3|0x8);
     return v.toString(16);
   });
 
--- a/src/api/define.js
+++ b/src/api/define.js
@@ -102,10 +102,19 @@ function generateUniqueName(name) {
 }
 
 function registerUniqueName(name) {
-  registry[name] = registry[name] ? registry[name] + 1 : 1;
+  if (registry[name]) {
+    registerUniqueName(name + '-'+ registry[name]);
+    registry[name] += 1;
+  } else {
+    registry[name] = 1;
+  }
 }
 
 export default function (name, opts) {
+  if (opts === undefined) {
+    throw new Error('You have to define options to register a component');
+  }
+
   const uniqueName = generateUniqueName(name);
   const Ctor = typeof opts === 'object' ? Component.extend(opts) : opts;
 
--- a/src/api/define.js
+++ b/src/api/define.js
@@ -97,17 +97,18 @@ function createInitProps (Ctor) {
 }
 
 function generateUniqueName(name) {
-  const registered = registry[name];
-  return registered ? `${name}-${registered}` : name;
-}
-
-function registerUniqueName(name) {
-  if (registry[name]) {
-    registerUniqueName(name + '-'+ registry[name]);
-    registry[name] += 1;
-  } else {
-    registry[name] = 1;
+  // we don't need to generate a unique name if it's the first time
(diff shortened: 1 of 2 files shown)


src/api/define.js
```diff
@@ -102,10 +102,19 @@ function generateUniqueName(name) {
 }
 
 function registerUniqueName(name) {
-  registry[name] = registry[name] ? registry[name] + 1 : 1;
+  if (registry[name]) {
+    registerUniqueName(name + '-'+ registry[name]);
```
treshugart: If you're using HMR in dev mode and you haven't refreshed the page, but it's reloaded x number of times, this has x - 1 levels of recursion. What was the problem with the previous implementation?

src/api/define.js
```diff
@@ -102,10 +102,19 @@ function generateUniqueName(name) {
...
+    registry[name] += 1;
+  } else {
+    registry[name] = 1;
+  }
 }
 
 export default function (name, opts) {
+  if (opts === undefined) {
+    throw new Error('You have to define options to register a component');
```
treshugart: Maybe reword this to include the name of the component?

test/unit/api/define.js
```diff
@@ -3,18 +3,38 @@ import { Component, define, symbols } from '../../../src/index';
 const { name: $name } = symbols;
 
 describe('api/define', () => {
+  it('should not register without any properties', () => {
+    expect(() => {
+      define('x-test');
+    }).to.throw(Error);
+  });
+
+  it('should not register a component without a dash in the name', () => {
```
treshugart: I could be missing something but I don't think this would be testing any code paths in our codebase as this is up to the polyfill and native implementations.

src/api/define.js
nadiam84: The problem is that it doesn't do what it suppose to do, that's all. It just bugs me. For example, those tests would fail with the previous implementation: https://github.com/skatejs/skatejs/pull/674/commits/b1072eb7a9b134383eb279b0f6defd89645ec44b#diff-e52899dbc88c4ba1c1db13686955c644R27.

And it's quite easy to imagine a practical situation based on those tests. For example:
<test-list>
  <test-item></test-item>
  <test-item-2></test-item-2>  
</test-list>
This would fail with HMR. And to make it worth fixing - it would fail unexpectedly for those who doesn't know how exactly skate works from the inside, i.e. for the outside developers. 

BTW, speaking of HMR and recursion. With every reload there would be registered x new components, all of which cannot be unregistered, isn't it? If this is okay, can we just instead of recursion register a new component every time with some unique hash instead of numbers? Using something like this: https://github.com/broofa/node-uuid.

src/api/define.js
nadiam84: Sure, first thing tomorrow

test/unit/api/define.js
nadiam84: Maybe you're right, it just looks nicer for me to see all the limitations in the one place. I'll remove it if you sure this is not worth is.

src/api/define.js
nadiam84: And actually it's not a recursion with HMR, it's just an extra step with every reload.

test/unit/api/define.js
treshugart: If we're going to test this, we would want to test [allthethings](http://w3c.github.io/webcomponents/spec/custom/#valid-custom-element-name) to be consistent, but I don't think it's worth it if we don't have anything to do with it.

src/api/define.js
treshugart: I can see how someone could get caught up by that and am now seeing extra step vs recursion because you're incrementing below. 👍

treshugart: The only concern with uuid (and other implementations) is size and performance. That library is almost 300 LoC which is significant weight when Skate itself is only 1200 LoC. I'd be interested to see if something like http://stackoverflow.com/a/2117523/110840 could work for this. My only concern there is degrading registration performance.

nadiam84:
Updated the PR, @treshugart can you take a look again? :)

src/api/define.js
```diff
@@ -97,19 +97,28 @@ function createInitProps (Ctor) {
...
 function generateUniqueName(name) {
-  const registered = registry[name];
-  return registered ? `${name}-${registered}` : name;
-}
+  // we don't need to generate a unique name if it's the first time
+  if (!registry[name]) {
+    registry[name] = true;
+    return name;
+  }
+  // copy-pasted from here, looks like a good solution : http://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript/2117523#2117523
```
treshugart: Uber nit: it'd be nice to clean up the function a little bit. `const`, `===`, `() => {}`; the works ;). Maybe make the comment just the URL.

treshugart:
lgtm

Issue #675, opened 2016-07-19
Remove "events" functionality


treshugart:
Since we can attach listeners to elements directly, should we remove the ability to specify "events"? This would clean up the API and remove any confusion around the two.

For example (using JSX to simplify the examples):

```js
skate('x-test', {
  events: {
    'click .my-button'() {}
  },
  render() {
    return <button class="my-button" />
  }
});
```

vs just:

```js
skate('x-test', {
  render() {
    return <button onclick={() => {}} />
  }
});
```

If there aren't any major objections to this, a simple workaround is to just `elem.addEventListener()` in one of the callbacks if you need to attach events to the host. Generally events would be attached to the host from higher up in the tree.

bradleyayers:
While removing it from the implementation is an interesting discussion, perhaps something easier to agree on is demoting it from the documentation and promoting the `onfoo={(event) => {…}}` style (here on referred to as the onFoo style). Encouraging the onFoo style is consistent with the vision of "React for web components".

Having two ways to do the same thing — where the difference appears only syntactic — increases the learning curve because it burdens me with needing to weigh up the strengths and weaknesses of each approach, without understanding the subtle differences between them.

I would prefer a single bullet-proof approach that alleviates me from needing to make a judgment call.

treshugart:
I'd prefer nutting this out and removing it altogether if possible.

lukebatchelor:
FWIW, I don't like the inline way of adding events

```
skate('x-test', {
  render() {
    return <button onclick={() => {}} />
  }
});
```

unless the handler is a function described somewhere else.

Other than that I don't mind either way. Which way is closer to Reacts? Would there be benefits in making them closer?

zephor:
I've so far enjoyed the declarative nature of the 'events' object. That said I'm not really that attached to it and the line of thought around simplifying API is a good one. What I'm curious about is what sparked this? The ability to bind directly to elements has always been there so we could have done this a long time ago. 

Also I notice the example is JSX oriented. I guess non-JSX users would just have to start using `addEventListener`, or in our case we would use jquery's `$.on`. Bottom line is I don't see any problem with removing it if we feel like it simplifies the API.

treshugart:
> Also I notice the example is JSX oriented. I guess non-JSX users would just have to start using addEventListener, or in our case we would use jquery's $.on.

Nah, you'd use `vdom.element()` instead:

```js
skate.vdom.element('button', { onclick: () => {} });
```

[...]

PR #676, opened 2016-07-19
#675 - Remove events functionality 

treshugart:
Preferred way is to use the `on*` syntax in render().

BREAKING CHANGE: Remove events spec in favour of on* attributes in render(). Implements #675.

If we ever find that similar functionality is required, it's easier to add rather than subtract API. Doing this now ensures we go into 1.0 with a clean slate.


README.md
```diff
@@ -309,43 +307,6 @@ skate.define('my-component', {
...
-  }
-});
-```
-
-Instead of firing for every click on the component element - or that bubbles to the component element - it will only fire if a descendant `<button>` was clicked.
-
-Event delegation works with or without a shadow root as it will inspect the event `path` if it exists.
-
-
-
```
bqx: As a consumer, it would be good to replace this section with a guide for how to use the `on*` syntax (doesn't appear to be documented at length anywhere else in the README, thought its mentioned in the `skate.link` section).

treshugart: It's [here](https://github.com/skatejs/skatejs#attrson), but I'll add a section called "Binding Events" that describes it and links off to that as well as provides workarounds for the host element.

nadiam84:
What about events for props? (https://github.com/skatejs/skatejs#event)
Shouldn't there be the same logic?

bqx:
LGTM, lots of red is good!

README.md
treshugart: https://github.com/skatejs/skatejs/tree/396d3793e86bbad7e7add639135699724e40dd0b#binding-events

treshugart:
@nadiam84 

> What about events for props? (https://github.com/skatejs/skatejs#event)

That is for triggering events when the property changes. This PR removes the `events` constructor option for automatically binding events to the host to listen for events on the host and descendant elements through event delegation.

README.md
```diff
@@ -1313,14 +1273,47 @@ And it could be used like:
...
+
+
+## Binding Events
+
+Generally, binding events to elements are done using the `vdom` [on* syntax](https://github.com/skatejs/skatejs#attrson):
+
+```js
+skate.define('x-element', {
+  render(elem) {
+    skate.vdom.element('div', { onclick: elem.handleClck });
```
nadiam84: handleClick

nadiam84:
LGTM

README.md
```diff
@@ -1313,14 +1273,47 @@ And it could be used like:
...
+  },
+  prototype: {
+    handleClick() {
```
stevemao: Would be good if we write the complete example?

```
handleClick(event) {
    // do something with `event`
}
```

README.md
```diff
@@ -1313,14 +1273,47 @@ And it could be used like:
...
+For instances where you need to bind listeners directly to your host element, you should do this in one of your lifecycle callbacks:
+
+```js
+skate.define('x-element', {
+  created(elem) {
+    elem.addEventListener('click', elem.handleClick);
+  },
+  prototype: {
+    handleClick() {
+
```
stevemao: Same here. Maybe we can provide a different event example. EG: `onchange`

README.md
treshugart: +1

bradleyayers:
LGTM

PR #678, opened 2016-07-19
#671 - Fix elementVoid() when passed function helpers

### LINKED ISSUE #671: Function helpers don't work with elementVoid()
```js
const Helper = () = <div />;
<Helper />
```

treshugart:
Fixes #671.


bradleyayers:
LGTM

PR #680, opened 2016-07-19
#679 - Fix props being passed to function helpers when calling Incremental DOM directly

### LINKED ISSUE #679: Attributes not honoured for function helper
https://github.com/skatejs/skatejs#function-helper

I wrote [a fiddle](http://jsbin.com/jikaxokuwu/edit?html,js,output) that's basically:

```javascript
const MyComponent = (props) => {
  skate.vdom.element('p', () => skate.vdom.text(props.text));
};

skate.define('x-element', {
  render(elem) {
    skate.vdom.elementVoid(MyComponent, null, null, 'text', 'the text')
  }
});
```

treshugart:
Fixes #679.

Something that we're not doing yet is passing `key` and `statics` on to helpers. This would increase the number of arguments, else we can create a generic object to pass in to the helper instead. Currently the API shape looks like:

```js
const Helper = (props, chren) => {};
```

With `key` and `statics` added:

```js
const Helper = (props, chren, key, statics) => {};
```

We could combine them:

```js
const Helper = (props, chren, { key, statics }) => {};
```

Pass them as `props`:

```js
const Helper = ({ key, statics, ...props }, chren) => {};
```

Or do everything together:

```js
const Helper = ({ key, statics, props, children }) => {};
```

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -102,11 +102,11 @@ function resolveTagName (tname) {
 
 function wrapIdomFunc (func, tnameFuncHandler = () => {}) {
   return function wrap (...args) {
-    const tname = args[0] = resolveTagName(args[0]);
-    if (typeof tname === 'function') {
+    args[0] = resolveTagName(args[0]);
+    if (typeof args[0] === 'function') {
       // If we've encountered a function, handle it according to the type of
       // function that is being wrapped.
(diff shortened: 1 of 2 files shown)


lukebatchelor:
Cool,

LGTM

Issue #684, opened 2016-07-22
Defer rendering to connectedCallback()


treshugart:
Instead of rendering in the `constructor` / `createdCallback`, we should instead render in `attachedCallback` / `connectedCallback`. This ensures that no unnecessary work is being done. If you're inspecting the shadow root in production code, stop it! If you're doing this in tests, then make sure your element is connected.

This change will involve:

- The main code change moving the render call to `attachedCallback`
- Updating tests to ensure everytime we inspect our internal shadow root that the element is connected

This creates a breaking change in that the `render()` lifecycle callback is now redundant because your component may not be rendered by the time it's called.

The `skate.ready()` function is still useful as it can be used to queue a callback for an element that may not have been upgraded yet. This is still a problem in polyfill land.

PR #685, opened 2016-07-22
#684 - Remove prop render() callbacks in favour of a single shouldRender() one

treshugart:
Implements #684.

The breaking change for the `ready()` callback was explained in the issue.

The breaking change for removing the prop `render()` function is because when we started rendering in the `attachedCallback()` / `connectedCallback()`, the number of renders that occurs at any given point in time was inconsistent in tests. To make this consistent, we check at the latest time possible (just before rendering) if we need to render using `shouldRender(elem, prevState, nextState)`. Having this check means that we no longer need checks on each prop.

Most of the changes to the tests involved appending each element to the fixture and adding `afterMutations()` wherever we needed to check the `shadowRoot` for anything. One spot was reorganised. The other spot was removing the tests for the `ready()` callback.


src/api/component.js
```diff
@@ -186,4 +179,12 @@ export default class Component extends HTMLElement {
 
     return Ctor;
   }
+
+  static shouldRender (elem, prev, curr) {
+    for (let name in prev) {
```
joscha: Do we need to compare `Object.keys(...).length`? what if curr got a new element, is that possible?

src/api/component.js
```diff
@@ -32,12 +33,17 @@ export default class Component extends HTMLElement {
   }
 
   connectedCallback () {
-    const cb = this.constructor.attached;
+    const ctor = this.constructor;
+    const cb = ctor.attached;
+    const render = ctor[$renderer];
+    this[$connected] = true;
+    render && render(this);
```
joscha: can this ever be falsy?

src/api/component.js
treshugart: That would mean something mutated `props` after an instance was created which I would think is something we shouldn't support.

src/api/component.js
treshugart: Shouldn't ever be, but it'd be a good idea to do a `typeof` check instead.

src/api/component.js
joscha: Cool, was just checking!

joscha:
LGTM

src/lifecycle/props-init.js
```diff
@@ -113,11 +98,9 @@ function createNativePropertyDefinition (name, opts) {
       opts.set(this, changeData);
     }
 
-    // Re-render on property updates if the should-update check passes.
-    if (prop.render(this, changeData)) {
-      const deb = this[$rendererDebounced] || (this[$rendererDebounced] = debounce(this.constructor[$renderer]));
-      deb(this);
-    }
+    // Queue a re-render.
+    const deb = this[$rendererDebounced] || (this[$rendererDebounced] = debounce(this.constructor[$renderer]));
```
jpnelson: Is there a performance reason for storing the debounced render?

src/lifecycle/render.js
```diff
@@ -1,18 +1,39 @@
 import { patchInner } from 'incremental-dom';
-import { shadowRoot as $shadowRoot } from '../util/symbols';
+import { connected as $connected, shadowRoot as $shadowRoot, state as $state } from '../util/symbols';
```
jpnelson: Could / should we export the symbols with the `$` prefix? Would this be another breaking change?

src/lifecycle/props-init.js
treshugart: Performance wasn't a concern here. Now that you mention it, this should be able to be added when the element is upgraded rather than doing it lazily here.

src/lifecycle/render.js
```diff
@@ -1,18 +1,39 @@
...
+      // We always do the initial render, therefore we only check if we should
+      // render if there is a shadow root. If there is no shadow root, then we
+      // are in the initial render.
+      if (sr && !shouldRender(elem, prevState, currState)) {
+        return;
+      }
+    }
+
+    // Setup the shadow root if it hasn't been setup yet.
+    if (!sr) {
```
jpnelson: Why not move this above the previous block? It would remove the need to check if `sr` is defined above.

test/unit/api/state.js
```diff
@@ -1,48 +1,63 @@
...
-    state(elem, {
-      prop1: 'updated1',
-      prop2: 'updated2',
-      undeclaredProp: 'updated3'
+  it('getting', () => {
+    it('should return only defined properties', function () {
+      const curr = state(elem);
+      expect(curr.prop1).to.equal('test1');
+      expect(curr.prop2).to.equal('test2');
+      expect(curr.undeclaredProp).to.equal(undefined);
```
jpnelson: Is this a new test? Is it related to the above changes?

src/lifecycle/props-init.js
treshugart: @jpnelson I've updated how the debounced renderer is stored.

src/lifecycle/render.js
joscha: yes, it broke some of the AtlasKit code the other week, so we decided to revert that change, as it is cosmetic and you can rename on import anyway.

test/unit/api/state.js
treshugart: I reorganised the tests here. The red was for setting, the green is for getting. Setting sets any prop. Getting only returns `props` you've defined for the component.

treshugart:
Build is failing because iPhone on SauceLabs hasn't been able to start for a few days.

src/lifecycle/render.js
treshugart: If you moved the block up `sr` would be exist (be truthy) and the initial render wouldn't happend because `shouldRender()` returns `false` on the initial render.

jpnelson: I see, so it doesn't really matter because it won't render initially anyway.

jpnelson:
LGTM

PR #689, opened 2016-07-26
Get perf tests running

treshugart:


treshugart:
Closing because the perf updates were upstreamed to #689 which has already been merged back into master.

Issue #688, opened 2016-07-26
Remove defined attribute


treshugart:
Currently we set a `defined` attribute (or customised using the `definedAttribute` option). This has two issues:

1. Can hit performance with lots of web components being initialised at once (minor concern)
2. If a parent component renders a child component, the child component will mutate itself from within the definition by adding the attribute. If the parent component re-renders and decides to remove that attribute because it doesn't know about it (any virtual DOM library could do this) then any styles used to prevent FOUC to undefined components will be reapplied to the element.

GIven the two issues with this, it's probably better to leave it up to the component authors and consumers to come up with a convention for this.

*In native custom elements, you have access to `:defined` to do this, so this really only applies to polyfill-land.*

bradleyayers:
I expect this will be a PITA while we still live in a polyfill world, so if we want to go ahead with this we should have some examples / guides on how to avoid FOUC for _today's_ world.

treshugart:
@bradleyayers some guidance added at https://github.com/skatejs/skatejs#preventing-fouc. We can expand on this as we encounter better workarounds.

treshugart:
[Seems](http://jsbin.com/nivobay/1/edit) that React and Incremental DOM don't actually behave like we had thought and they keep attributes they don't know about, so long as they aren't setting them through their vdom implementations.

Let's revert this change and re-update the docs.

bradleyayers:
Interesting. I knew that React doesn't touch attributes it doesn't know about, but I was under the impression idom did. So does idom have its own virtual DOM representation too? I assumed it traversed and queried the DOM in lock-step with idom function calls?

treshugart:
It does lock-step, but it doesn't query for the current DOM state. Instead stored what they call "node data" on the element with what they currently about it. So in that sense it makes it the same as React. I was quite happy to figure this out about them both (and it seems most, if not all, virtual DOM implementations behave this way). It would be counterintuitive to performance for any implementation to have to query and remove attributes they don't know about. This is actually what caused me to reinvestigate the potential issue. Previously, my tests showed that they removed it, but that's because I had tested it where the vDOM impl knew about the attribute, I believe.

This is good news.

treshugart:
@seancurtis ^. This is good news for a11y.

treshugart:
@jpnelson ^. This may slightly change how we think about smart / dumb components.

bradleyayers:
Cool — thanks for the explanation!

seancurtis:
[...]

PR #690, opened 2016-07-26
#688 - Remove definedAttribute

treshugart:
Implements #688.

Requires https://github.com/skatejs/skatejs/pull/689 for perf testing.


karma.conf.js
```diff
@@ -11,6 +11,14 @@ module.exports = function (config) {
     config.browsers.push('internet_explorer_11');
   }
 
+  config.files = [
+    // React
+    'https://scontent.xx.fbcdn.net/t39.3284-6/13591530_1796350410598576_924751100_n.js',
```
joscha: what are these URLs?

treshugart: If we used installed versions then perf tests aren't accurate because of dev mode (these are prod versions, so they perform better). What I'd like to do here is try minifying the local perf tests so that it does the same for incremental dom. I'll try and see what happens.

joscha:
@treshugart - in a post-"defined"-world, what would be the best way to
determine whether a component has been successfully initialized? Would it
be worth making this a symbol-based property instead of removing it
completely?

treshugart:
@joscha how would you target that from css?

treshugart:
Added some docs here about it https://github.com/skatejs/skatejs#preventing-fouc. I really don't see it being a big deal for component authors to add their own attributes.

joscha:
@treshugart I wanted to use it from selenium - for CSS it is not targetable.

treshugart:
Oh, I see. The `ready()` function triggers a callback when the element is ready (or immediately if it is already upgraded):

```js
skate.ready(elem, callback);
```

Would that work for this case?

joscha:
Totally forgot about ready, yes that would work!

PR #699, opened 2016-07-28
chore(deps): add missing devdeps

stevemao:
Alternatively we could ask user to install them globally. But devdeps is probably better.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,13 +27,20 @@
     "regex-native-function": "^1.0.0"
   },
   "devDependencies": {
+    "babel-core": "^6.11.4",
+    "babel-loader": "^6.2.4",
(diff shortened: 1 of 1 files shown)


treshugart:
Interesting. These weren't getting pulled in by [`skatejs-build`](https://github.com/skatejs/build)?

stevemao:
I notice that I'm using npm v2, so this change should probably be discussed.

treshugart:
Oh right. Try NPM 3. If that's the case we should try and enforce that in the package.json if we can.

stevemao:
Oh yeah, if we purposely pull the deps from skatejs-build this is not needed.

PR #700, opened 2016-07-29
fix: attributeChanged fixed

nadiam84:
According to the documentation, `The attributeChangedCallback() is invoked for every attribute that exists on the element at the time of creation.`.
But for the properties that were defined as attributes this callback was never called.
This PR is fixing it.

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -72,13 +72,12 @@ export default class Component extends HTMLElement {
       // then that causes the attribute to be set again.
       if (propData.syncingAttribute) {
         propData.syncingAttribute = false;
-        return;
+      } else {
+        // Sync up the property.
+        const propOpts = this.constructor.props[propertyName];
+        propData.settingAttribute = true;
+        this[propertyName] = newValue !== null && propOpts.deserialize ? propOpts.deserialize(newValue) : newValue;
       }
-
-      // Sync up the property.
-      const propOpts = this.constructor.props[propertyName];
-      propData.settingAttribute = true;
-      this[propertyName] = newValue !== null && propOpts.deserialize ? propOpts.deserialize(newValue) : newValue;
     }
 
     if (attributeChanged) {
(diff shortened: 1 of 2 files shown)


test/unit/lifecycle/attribute-changed.js
```diff
@@ -18,4 +19,24 @@ describe('lifecycle/attribute-changed', function () {
...
+      attributeChanged () {
+        counter++;
+      },
+      props: {
+        test: {
+          attribute: true
+        },
+      },
+    }));
+    afterMutations(done);
```
treshugart: You might want to make assertions within `afterMutations` and calling `done` last, like:

```js
afterMutations(
  () => expect...,
  () => elem.test = ...,
  ...
  done
);
```

nadiam84: done

src/api/component.js
```diff
@@ -72,13 +72,12 @@ export default class Component extends HTMLElement {
       // then that causes the attribute to be set again.
       if (propData.syncingAttribute) {
         propData.syncingAttribute = false;
-        return;
+      } else {
+        // Sync up the property.
+        const propOpts = this.constructor.props[propertyName];
+        propData.settingAttribute = true;
```
bengummer: i am probably wrong, but should this be `propData.syncingAttribute`? just asking because of the if condition

test/unit/lifecycle/attribute-changed.js
```diff
@@ -18,4 +19,26 @@ describe('lifecycle/attribute-changed', function () {
       }
     });
   });
+
+  it('attributes that are defined as properties should call attributeChanged callback', function (done) {
+    let counter = 0;
+    let elem = new (element().skate({
```
bengummer: tiny thing: `const` could be used here

src/api/component.js
nadiam84: judging by the code those are the two different data settings

PR #701, opened 2016-07-29
style: airbnb linting

nadiam84:
Only linting fixes here


treshugart:
lgtm

treshugart:
Can you have a look at the build? I think one of the linting fixes may have some breakages: https://travis-ci.org/skatejs/skatejs/builds/148266842

nadiam84:
@treshugart fixed

src/api/prop.js
```diff
@@ -1,39 +1,39 @@
...
-  serialize: JSON.stringify
+  serialize: JSON.stringify,
 });
 
 export const boolean = create({
   coerce: value => !!value,
   default: false,
   deserialize: value => !(value === null),
-  serialize: value => value ? '' : undefined
+  serialize: (value) => (value ? '' : undefined),
```
bengummer: any reason why it is `(value)` here but not on the `deserialize` line?

src/api/props.js
```diff
@@ -1,15 +1,16 @@
 import { renderer as $renderer } from '../util/symbols';
 import assign from '../util/assign';
 
-function get (elem) {
+function get(elem) {
   const props = {};
-  for (let key in elem.constructor.props) {
+  Object.keys(elem.constructor.props).forEach((key) => {
```
bengummer: We use a `for..in` loop at https://github.com/skatejs/skatejs/pull/701/files#diff-d3ff8469a13e890ec2a5c73b90df75abR185 - can we use it here as well?

test/boot.js
```diff
@@ -1,10 +1,10 @@
 import 'webcomponents.js/CustomElements';
 import helperFixture from './lib/fixture';
 
-mocha.setup({ timeout: 10000 });
+mocha.setup({ timeout: 10000 }); // eslint-disable-line no-undef
```
bengummer: to avoid the `eslint-disable-line`, you could add `mocha` to the `env` section of `.eslintrc`

test/perf/react.js
```diff
@@ -5,17 +5,17 @@ const { React, ReactDOM } = window;
 
 
 // Skate components.
-const wclist = (props, chren) => ul(props, chren);
-const wcitem = (props, chren) => li(props, chren);
-const [ div, h1, li, ul, item, list ] = ['div', 'h1', 'li', 'ul', wclist, wcitem].map(t => vdom.element.bind(null, t));
+const wclist = (ps, chren) => ul(ps, chren); // eslint-disable-line no-use-before-define
+const wcitem = (ps, chren) => li(ps, chren); // eslint-disable-line no-use-before-define
```
bengummer: optional: not clear how `li(..)` and `ul(..)` are defined? would be cool to import them directly if possible, or if they are globals a short comment would help.

test/perf/react.js
```diff
@@ -94,31 +95,33 @@ describe('update', () => {
...
         comp,
-        count: 0
+        count: 0,
       })
-        .then(ops => console.log(`Skate (update): ${ops} / sec`))
-        .then(() => fixture.innerHTML = '')
+        .then(ops => console.log(`Skate (update): ${ops} / sec`)) // eslint-disable-line no-console
+        .then(() => {
+          fixture.innerHTML = '';
+        })
```
bengummer: optional: could just group these 2 `.then()` blocks into 1

test/unit/constructor.js
```diff
@@ -2,16 +2,16 @@ import { define } from '../../src/index';
 import helperElement from '../lib/element';
 import resolved from '../lib/resolved';
 
-describe('constructor', function () {
-  var id;
+describe('constructor', () => {
```
bengummer: not sure if we should be using arrow functions with mocha based on https://mochajs.org/#arrow-functions ?

src/api/prop.js
nadiam84: no reason, good catch

test/boot.js
treshugart: You might be able to do `const { mocha } = window`, too.

test/perf/react.js
treshugart: That's something I missed in my original review, nice catch. Line 10 there shouldn't have been removed, I think.

src/api/props.js
nadiam84: for...in loop are being yelled at by airbnb
in case of the loop there changing it on Object.keys breaks some tests for some reason, not sure why and should we invest time to fix it now.

test/perf/react.js
treshugart: You can also wrap the assignment in parenthesis instead of having to use curly braces.

test/unit/constructor.js
treshugart: If you require the mocha context, you can convert it to `function`.

test/perf/react.js
nadiam84: it goes beyond the 'linting fixes' scope I think

test/perf/react.js
nadiam84: it goes beyond the 'linting fixes' scope I think

test/unit/constructor.js
nadiam84: we're using arrow-functions everywhere (including AtlasKit), since normal functions are being yelled at by the airbnb linter. I disabled lines where we're actually using `this`, not many of them.

treshugart: The linter should in-fact look at the function and see if it is using this. If it is, then it doesn't make assumptions about which function form you should use.

test/perf/react.js
treshugart: Sure. I'm guessing this was one of the automated fixes.

test/perf/react.js
treshugart: Looking at the full changes @bengummer, they are defined just below the function helpers.

PR #716, opened 2016-08-04
fix(vdom): Fix issue where the attribute applicator was trying to set…

### LINKED ISSUE #566: Problems with iDOM default attribute setter

treshugart:
… readonly properties on SVG ele

Fixes #566

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -94,8 +94,13 @@ attributes[symbols.default] = function (elem, name, value) {
     value = `[slot="${value}"]`;
   }
 
-  // Set defined props on the element directly.
-  if (name in elem) {
+  // Set defined props on the element directly. This ensures properties like
+  // "value" on <input> elements get set correctly. Setting those as attributes
+  // doesn't always work and setting props is faster than attributes.
+  //
+  // However, certain props on SVG elements are readonly and error when you try
+  // to set them.
+  if (name in elem && !('ownerSVGElement' in elem)) {
     applyProp(elem, name, value);
     return;
   }
--- a/test/unit/vdom/properties.js
+++ b/test/unit/vdom/properties.js
@@ -31,14 +31,13 @@ describe('vdom/properties', () => {
 
   it('should not set properties on SVG elements', done => {
     expect(() => {
-      new (define('x-test', { // eslint-disable-line no-new
+      const Test = define('x-test', {
         render() {
-          vdom.element('svg', { height: 100 });
+          vdom.element('svg', { width: 100 });
         },
-      }));
-      afterMutations(
-        done
-      );
+      });
+      fixture(new Test());
+      afterMutations(done);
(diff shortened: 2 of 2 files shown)


lukebatchelor:
lgtm

src/api/vdom.js
```diff
@@ -94,8 +94,13 @@ attributes[symbols.default] = function (elem, name, value) {
...
 
-  // Set defined props on the element directly.
-  if (name in elem) {
+  // Set defined props on the element directly. This ensures properties like
+  // "value" on <input> elements get set correctly. Setting those as attributes
+  // doesn't always work and setting props is faster than attributes.
+  //
+  // However, certain props on SVG elements are readonly and error when you try
+  // to set them.
+  if (name in elem && !('ownerSVGElement' in elem)) {
```
nadiam84: why not `elem.hasOwnProperty('ownerSVGElement')`?

nadiam84:
lgtm

test/unit/vdom/properties.js
```diff
@@ -31,14 +31,13 @@ describe('vdom/properties', () => {
 
   it('should not set properties on SVG elements', done => {
     expect(() => {
-      new (define('x-test', { // eslint-disable-line no-new
+      const Test = define('x-test', {
         render() {
-          vdom.element('svg', { height: 100 });
+          vdom.element('svg', { width: 100 });
```
bqx: Any reason for setting `width` instead of `height` here? Are they both read-only?

src/api/vdom.js
bqx: Is it possible to make this fix more general by checking that the property isn't read-only, or is there some performance consideration here?

bqx:
lgtm pending comments

test/unit/vdom/properties.js
treshugart: Width was the prop I had issues with but height would reproduce it too. It just got changed when I was messing with the test. It won't matter which one it is, though.

src/api/vdom.js
treshugart: @nadiam84 it's not an own property. You'd have to walk the prototype chain depending on how far up `SVGElement` is.

@bqx You'd have to do `Object.getOwnPropertyDescriptor()` and check `writable`, but just like `hasOwnProperty()` you'd have to do it at the prototype level as it wouldn't be an own property on the element instance.

Depending on the element, how far up you'd have to walk can vary and walking prototype chains can be expensive. Even so, this is performance critical code. The less object accessing and function calls the better.

PR #722, opened 2016-08-10
#721 - Fix ref

### LINKED ISSUE #721: ref callback not called for function components
When I pass a function into the `skatejs.vdom.element*` APIs the ref callback is not called and passed the DOM node.

treshugart:
Fixes #721 and #724.

Some minor refactors to simplify the default attribute handler.

Fixed `ref` so that it will only be called if the element changes or if the ref changes. if you specify a closure in `render()` it will always change. If you supply a function that doesn't change, it will only ever be called if the element is re-rendered.

Fixed `skip` so that it actually skips the entire tree. Incremental DOM requires that you either `skip()` or render your tree. Providing this API is syntactic sugar for not having to worry about `if / else` blocks and makes it more declarative. This is being provided as a fix as it was originally intended to work this way, but was broken.

See this [bin](http://jsbin.com/yaralol/12/edit) for how you'd do it with IDOM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,14 +26,12 @@
     "incremental-dom": "0.4.1"
   },
   "devDependencies": {
-    "birdpoo": "^0.2.2",
-    "es6-shim": "^0.35.1",
-    "eslint": "^3.2.0",
-    "react": "^15.3.0",
-    "react-dom": "^15.3.0",
-    "skatejs-build": "^7.0.1",
-    "skatejs-named-slots": "^1.0.0",
-    "webcomponents.js": "webcomponents/webcomponentsjs#b77ca74"
+    "birdpoo": "0.x",
+    "eslint": "3.x",
+    "react": "15.x",
+    "react-dom": "15.x",
+    "skatejs-build": "7.x",
+    "skatejs-web-components": "*"
   },
   "scripts": {
     "prepublish": "rollup --config rollup.config.js && webpack --config webpack.config.js",
@@ -43,6 +41,5 @@
     "commitizen": {
       "path": "cz-conventional-changelog"
     }
-  },
-  "version": "1.0.1"
+  }
 }
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -1,9 +1,11 @@
 import {
   applyProp,
   attributes,
+  currentElement,
   elementClose,
   elementOpen,
   elementVoid,
+  skip,
   symbols,
   text,
 } from 'incremental-dom';
@@ -75,7 +77,10 @@ attributes.ref = function (elem, name, value) {
 // Skip handler.
 attributes.skip = function (elem, name, value) {
   if (value) {
+    skip();
     elem[$skipCurrentElement] = true;
+  } else {
+    delete elem[$skipCurrentElement];
   }
 };
 
@@ -161,7 +166,9 @@ function wrapIdomFunc(func, tnameFuncHandler = () => {}) {
       const isElementClosing = func === elementClose;
       const isElementOpening = func === elementOpen;
 
-      if (skipCurrentTree && !isElementClosing) {
+      // If we're skipping the tree, we must skip everything except for the
+      // closing of the element that originally started the skipping.
+      if (skipCurrentTree && !isElementClosing && !currentElement()[$skipCurrentElement]) {
         return;
       }
 
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -4,7 +4,6 @@ import {
   elementClose,
   elementOpen,
   elementVoid,
-  skip,
   symbols,
   text,
 } from 'incremental-dom';
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -1,11 +1,8 @@
 import {
   applyProp,
-  attr,
   attributes,
   elementClose,
   elementOpen,
-  elementOpenEnd,
-  elementOpenStart,
   elementVoid,
   skip,
   symbols,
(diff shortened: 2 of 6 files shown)


test/unit/vdom/skip.js
```diff
@@ -0,0 +1,60 @@
...
+
+describe('vdom/skip', () => {
+  it('should skip the element children', done => {
+    const Elem = define('x-test', {
+      props: {
+        num: prop.number()
+      },
+      render() {
+        vdom.element('div', { skip: true }, () => {
+          vdom.element('span', () => {
```
bradleyayers: What's the rationale for this *not* throwing an exception? (since skip was declared on the parent). I thought the idom behaviour was for this to throw?

test/unit/vdom/skip.js
```diff
@@ -0,0 +1,60 @@
...
+    );
+  });
+
+  it('should allow conditional rendering', done => {
+    function isEven(num) {
+      return num % 2 === 0;
+    }
+    const Elem = define('x-test', {
+      props: {
+        num: prop.number()
```
bradleyayers: I assume the default value is 0. It might be nice to add that explicitly `{ default: 0 }` to remove any ambiguity.

test/unit/vdom/skip.js
```diff
@@ -0,0 +1,60 @@
...
+            vdom.text('text');
+          });
+        });
+      },
+    });
+    const elem = new Elem();
+    fixture(elem);
+    afterMutations(
+      () => expect(elem[symbols.shadowRoot].textContent).to.equal(''),
+      () => props(elem, { num: elem.num + 1 }),
```
lukebatchelor: Do we know that the props call would have definitely caused a re-render before the next expect would have been called?

package.json
```diff
@@ -43,6 +41,5 @@
     "commitizen": {
       "path": "cz-conventional-changelog"
     }
-  },
-  "version": "1.0.1"
```
lukebatchelor: Just confirming, we dont need the version field anymore? This is handled by semantic release?

test/unit/vdom/skip.js
treshugart: Yeah, this gets kinda complicated.

- Incremental DOM allows you to put content in it, you just have to either call skip(), or generate content.
- Once you call `skip()`, it only throws if you're using the untranspiled source.
- If you're using the `dist/` (which most people are), it doesn't throw. It just does nothing.

This is really just syntactic sugar for doing what you can already do with Incremental DOM, it just removes the need to know about implementation details where you must either call `skip()` or generate content. In our case, you can switch skipping on and off and it will manage either calling `skip()` or rendering your descendant tree.

bradleyayers: Sounds fine to me.

treshugart: The last point there might actually render content. That said, it doesn't really matter. I updated the original description with another link to the updated bin showing how you'd do what Skate does manually.

test/unit/vdom/skip.js
treshugart: Haha! Sure, starting at 2 sounds like a good idea.

package.json
treshugart: Yeah, semantic-release will ensure the correct one is published. This was causing some confusion.

src/api/vdom.js
```diff
@@ -16,15 +14,31 @@ import { shadowDomV0, shadowDomV1 } from '../util/support';
 
 const applyDefault = attributes[symbols.default];
 const fallbackToV0 = !shadowDomV1 && shadowDomV0;
+
+// A stack of children that corresponds to the current function helper being
+// executed.
 const stackChren = [];
-const stackProps = [];
+
+const $skipCurrentElement = '__skip';
```
lukebatchelor: Will these be replaced with symbols at some point?

src/api/vdom.js
```diff
@@ -133,36 +152,62 @@ function resolveTagName(tname) {
...
+      } else if (isElementClosing) {
         const eref = elem[$ref];
+
+        // We delete so that it isn't called again for the same element. If the
+        // ref changes, or the element changes, this will be defined again.
+        delete elem[$ref];
+
+        // Execute the saved ref after esuring we've cleand up after it.
         if (typeof eref === 'function') {
           eref(elem);
```
lukebatchelor: nitpick, but the function name is a bit confusing without context. Even `executeRef` might be easier? (shrug)

src/api/vdom.js
treshugart: At some point, yep.

lukebatchelor:
LGTM

src/api/vdom.js
treshugart: Just changed to `ref`. No reason it needs to be `eref`.

package.json
```diff
@@ -26,14 +26,12 @@
...
-    "react-dom": "^15.3.0",
-    "skatejs-build": "^7.0.1",
-    "skatejs-named-slots": "^1.0.0",
-    "webcomponents.js": "webcomponents/webcomponentsjs#b77ca74"
+    "birdpoo": "0.x",
+    "eslint": "3.x",
+    "react": "15.x",
+    "react-dom": "15.x",
+    "skatejs-build": "7.x",
+    "skatejs-web-components": "*"
```
bradleyayers: Seems a little unrelated, but whatever.

bradleyayers:
LGTM

package.json
treshugart: It was. It was about 1am and I cherry-picked it to master just after.

treshugart:
Brad's lgtm didn't trigger the check to pass because he wasn't a maintainer, but is now. This should make it green.

PR #727, opened 2016-08-12
fix(emit): Fix emit behaviour so that it's consistent with all browsers.

### LINKED ISSUE #726: Preventing emit() on disabled elements prevents it from emitting on custom elements
This was originally a workaround to fix inconsistent behaviour in FIrefox for built-in form elements where events aren't triggered on disabled elements, but for other browsers they were (and Firefox sometimes spit out an error).

See: #463 and #465.

We need to be able to emit events on custom elements that have a `disabled` attribute and the current behaviour is inconsistent with all but Firefox.

treshugart:
Firefox still has a bug regarding this issue. This also fixes a problem where we couldn't emit

events on any elements (custom elements) that have a "disabled" property and this is absolutely

something that people need to do.

Fixes #726.

### DIFF
--- a/src/api/emit.js
+++ b/src/api/emit.js
@@ -25,5 +25,5 @@ export default function (elem, name, opts = {}) {
   if (opts.cancelable === undefined) {
     opts.cancelable = true;
   }
-  return elem.disabled ? true : elem.dispatchEvent(createCustomEvent(name, opts));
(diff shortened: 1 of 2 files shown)


nadiam84:
lgtm

treshugart:
lgtm

PR #728, opened 2016-08-12
feat: Add the "defined" attribute when the element is upgraded.

treshugart:
#688

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -89,6 +89,8 @@ export default class Component extends HTMLElement {
       created(this);
     }
 
+    this.setAttribute('defined', '');
+
     if (readyCallbacks) {
       readyCallbacks.forEach(cb => cb(this));
       delete elemData.readyCallbacks;
--- a/test/unit/lifecycle/defined.js
+++ b/test/unit/lifecycle/defined.js
@@ -1,0 +1,9 @@
+import { define } from '../../../src/index';
+
+describe('lifecycle/defined', () => {
+  it('should add the [defined] attribute when the element is upgraded', () => {
+    const Elem = define('x-test', {});
+    const elem = new Elem();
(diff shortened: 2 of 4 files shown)


test/unit/vdom/incremental-dom.js
```diff
@@ -32,17 +32,17 @@ describe('IncrementalDOM', () => {
 
       patchIt('elementOpen, elementClose', () => {
         vdom.elementOpen(Elem);
-        expect(vdom.elementClose(Elem).outerHTML).to.equal('<x-test></x-test>');
+        expect(vdom.elementClose(Elem).tagName).to.equal('X-TEST');
```
treshugart: These tests had to be updated because they were testing `outerHTML`. The result is different depending on the browser so it's safer to test `tagName` as it's not producing any content anyways.

src/api/component.js
```diff
@@ -89,6 +89,8 @@ export default class Component extends HTMLElement {
       created(this);
     }
 
+    this.setAttribute('defined', '');
```
treshugart: It's probably best to add this before the `readyCallbacks` are fired so that the element has layout when those are called but this is up for discussion.

lukebatchelor:
LGTM

test/unit/lifecycle/defined.js
```diff
@@ -0,0 +1,9 @@
+import { define } from '../../../src/index';
+
+describe('lifecycle/defined', () => {
+  it('should add the [defined] attribute when the element is upgraded', () => {
```
nadiam84: 'it should not have [defined] attribute when the element is not upgraded'

treshugart: Good point. Will add now.

treshugart: Good point. I'll add a test and push it to master once the release happens for 1.1.

Issue #723, opened 2016-08-11
Multiple versions of skate on a page has issues with reregistering components


lukebatchelor:
When multiple versions of skate are on a page attempting to re-register a component will throw an error as a component of that name already exists, (likely due to one instance having registered and then the second instance not having the name in its cache).

Possible fix is when generating the new name, check that it hasnt been registered yet.

treshugart:
For v0 you'll need to `try / catch`. It's also worth measuring performance to see if that degrades it too much.

lukebatchelor:
Would something like this fit both use cases?

```
var isRegistered = function(name) {
  return document.createElement(name).constructor !== HTMLElement;
}
```

treshugart:
The only thing with that is I think there are situations where an element can be an instance of `HTMLUnknownElement`.

treshugart:
I think that wrapping a try / catch around v0 would be faster, but if not you might be able to do (or a variant of):

```js
function isCustomElement(node) {
  const ctor = node.constructor;
  return ctor !== HTMLElement && ctor !== HTMLUnknownElement;
}
```

nadiam84:
Created a PR for this issue #729

nadiam84:
@lukebatchelor 

> Would something like this fit both use cases?

It doesn't work. Because of the CustomElements polyfill. They are changing the HTMLElement, so this `document.createElement(name).constructor !== HTMLElement;` will never be true.

Also, this approach will fire 'create' and 'attributechange' callbacks, which is not good and breaks some tests.

PR #729, opened 2016-08-15
fix: define for multiple instances

nadiam84:
When there are multiple skate instances on a page, some components could be defined in more than one instance. This PR ensures that even in this situation all the components will be registered.

### DIFF
--- a/src/api/define.js
+++ b/src/api/define.js
@@ -108,12 +108,23 @@ function generateUniqueName(name) {
   return `${name}-${rand}`;
 }
 
+function registerElement(uniqueName, Ctor) {
+  if (customElementsV1) {
+    window.customElements.define(uniqueName, Ctor, { extends: Ctor.extends });
+    return Ctor;
+  } else if (customElementsV0) {
+    return document.registerElement(uniqueName, Ctor);
+  } else {
+    throw new Error('Skate requires native custom element support or a polyfill.');
+  }
+}
+
 export default function (name, opts) {
   if (opts === undefined) {
     throw new Error(`You have to define options to register a component ${name}`);
   }
 
-  const uniqueName = generateUniqueName(name);
+  let uniqueName = generateUniqueName(name);
   const Ctor = typeof opts === 'object' ? Component.extend(opts) : opts;
 
   formatLinkedAttributes(Ctor);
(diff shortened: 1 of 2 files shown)


src/api/define.js
```diff
@@ -122,13 +133,16 @@ export default function (name, opts) {
...
   Ctor[$renderer] = createRenderer(Ctor);
 
-  if (customElementsV1) {
-    window.customElements.define(uniqueName, Ctor, { extends: Ctor.extends });
-  } else if (customElementsV0) {
-    return document.registerElement(uniqueName, Ctor);
-  } else {
-    throw new Error('Skate requires native custom element support or a polyfill.');
+  let res;
+  try {
```
treshugart: You don't need to `try` in v1; you can just check if it exists using `window.customElements.get()`. Admittedly this kinda messes up the clean logic you've laid out, though. Maybe you can split out registration into `registerV0` and `registerV1` or something?

treshugart: Also, it's a V8 implementation detail, but I don't think functions that contain a `try / catch` are optimised, so there'd possibly be a performance hit registering lots of elements.

src/api/define.js
```diff
@@ -122,13 +133,16 @@ export default function (name, opts) {
...
+    res = registerElement(uniqueName, Ctor);
+  } catch(e) {
+    uniqueName = generateUniqueName(name);
```
bengummer: this assumes that the 2nd call of `generateUniqueName(name)` will produce something unique. Probably safe, but would be safer to use something like @treshugart 's [idea/function](https://github.com/skatejs/skatejs/issues/723#issuecomment-239343459) until a unique name is found.

nadiam84: it will be unique, just look at the generateUniqueName function :)

bengummer: Thanks for explaining IRL :) it's a shame that it sounds like we can't use the idea/function linked above, because it's a bit confusing that the first call of `generateUniqueName(name)` doesn't actually return a unique name.

src/api/define.js
nadiam84: What if we just get rid of the 'registry' object and always return unique name no matter what? This will simplify logic a bit and eliminate try/catch block.
But it will be a breaking change.

treshugart: That would break those wanting to use standard HTML, which is kinda the foundation of web components.

src/api/define.js
```diff
@@ -108,27 +108,50 @@ function generateUniqueName(name) {
...
-
-  const uniqueName = generateUniqueName(name);
   const Ctor = typeof opts === 'object' ? Component.extend(opts) : opts;
-
   formatLinkedAttributes(Ctor);
 
-  Ctor[$name] = uniqueName;
-  Ctor[$props] = createInitProps(Ctor);
-  Ctor[$renderer] = createRenderer(Ctor);
+  let uniqueName = generateUniqueName(name);
```
treshugart: I don't think you need this, or the checks in the local registry anymore. You can just check / try-catch against `name` and only generate a unique name if necessary, it would seem.

treshugart: Can probably remove the local registry, too.

src/api/define.js
```diff
@@ -108,27 +108,50 @@ function generateUniqueName(name) {
...
+  let res;
+
+  prepareForRegistration(uniqueName, Ctor);
 
   if (customElementsV1) {
-    window.customElements.define(uniqueName, Ctor, { extends: Ctor.extends });
+    if (window.customElements.get(uniqueName)) {
```
treshugart: nit: would these checks be better contained in the respective `register*` functions?

treshugart:
lgtm

lukebatchelor:
LGTM (as we are adding the tests separately)

bengummer:
Added a test case - it mocks out the define function (see `test/unit/api/define.js`)  because it is really hard to get two versions of skate on the same page.

PR #732, opened 2016-08-15
fix(vdom): Fix skipping algorithm so that it skips immedate children …

### LINKED ISSUE #731: Skipping doesn't work on immediate children
The following:

```js
vdom.element('div', { skip: true }, () => {
  vdom.text('1');
  vdom.element('span', () => vdom.text('2'));
});
```

Will output:

```js
1
```

But it shouldn't output anything.

treshugart:
…and descendants.

Fixes #731

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -19,7 +19,6 @@ const fallbackToV0 = !shadowDomV1 && shadowDomV0;
 // executed.
 const stackChren = [];
 
-const $skipCurrentElement = '__skip';
 const $currentEventHandlers = '__events';
 const $stackCurrentHelperProps = '__props';
 
@@ -30,8 +29,8 @@ let stackCurrentHelper;
 // to pass the main elementOpen() function.
 let overrideArgs;
 
-// Whether or not to skip the current rendering tree.
-let skipCurrentTree = false;
+// The number of levels deep after skipping a tree.
+let skips = 0;
 
 // Adds or removes an event listener for an element.
 function applyEvent(elem, ename, newFunc) {
@@ -78,9 +77,9 @@ attributes.ref = function (elem, name, value) {
 attributes.skip = function (elem, name, value) {
   if (value) {
     skip();
-    elem[$skipCurrentElement] = true;
+    elem.__skip = true;
   } else {
-    delete elem[$skipCurrentElement];
+    delete elem.__skip;
   }
 };
 
@@ -149,6 +148,16 @@ function resolveTagName(tname) {
   return tname;
 }
 
+function elementOpenStart(...args) {
+  overrideArgs = args;
+}
+
+function elementOpenEnd() {
+  const node = newElementOpen(...overrideArgs);
+  overrideArgs = null;
+  return node;
+}
+
 function wrapIdomFunc(func, tnameFuncHandler = () => {}) {
   return function wrap(...args) {
     args[0] = resolveTagName(args[0]);
@@ -163,28 +172,21 @@ function wrapIdomFunc(func, tnameFuncHandler = () => {}) {
       // children, it will queue up for the next stack, if there is one.
       stackChren[stackChren.length - 1].push([wrap, args]);
     } else {
-      const isElementClosing = func === elementClose;
-      const isElementOpening = func === elementOpen;
-
-      if (isElementOpening) {
-        if (skipCurrentTree) {
-          return;
-        } else {
-          const elem = func(...args);
-          skipCurrentTree = !!elem[$skipCurrentElement];
-          return elem;
-        }
+      // Stat skipping if the element has been flagged as skipped.
+      if (!skips && currentElement().__skip) {
+        ++skips;
       }
-      
-      if (isElementClosing) {
-        let elem = currentElement();
 
-        if (elem[$skipCurrentElement]) {
-          skipCurrentTree = false;
-        } else {
-          elem = func(...args);
+      if (func === elementOpen) {
+        return skips ? ++skips : func(...args);
+      }
+
+      if (func === elementClose) {
+        if (skips) {
+          return --skips;
         }
 
+        const elem = func(...args);
         const ref = elem[$ref];
 
         // We delete so that it isn't called again for the same element. If the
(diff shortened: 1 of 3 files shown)


src/api/vdom.js
```diff
@@ -166,17 +166,25 @@ function wrapIdomFunc(func, tnameFuncHandler = () => {}) {
...
+        }
       }
+      
+      if (isElementClosing) {
+        let elem = currentElement();
 
-      const elem = func(...args);
+        if (elem[$skipCurrentElement]) {
+          skipCurrentTree = false;
+        } else {
```
bradleyayers: Shouldn't this be `else if (!skipCurrentTree) {`?

test/unit/vdom/skip.js
```diff
@@ -10,6 +10,7 @@ describe('vdom/skip', () => {
       },
       render() {
         vdom.element('div', { skip: true }, () => {
+          vdom.text('text');
```
bradleyayers: Why is this added?

treshugart: Yeah, but more importantly immediate children in general. That's what was broken. I should probably be checking `innerHTML` rather that `textContent` below to test both.

src/api/vdom.js
treshugart: Good point. I'm not sure why it would be passing then, though. I'll investigate.

test/unit/vdom/skip.js
nadiam84: `render() {
        vdom.element('div', { skip: true }, () => {
          vdom.element('span', () => {
            vdom.text('text');
          });
          vdom.element('span', () => {
            vdom.text('text');
          });
        });
      },`

render function like this will fail the test

src/api/vdom.js
```diff
@@ -166,17 +166,25 @@ function wrapIdomFunc(func, tnameFuncHandler = () => {}) {
...
+          return elem;
+        }
       }
+      
+      if (isElementClosing) {
+        let elem = currentElement();
 
-      const elem = func(...args);
+        if (elem[$skipCurrentElement]) {
+          skipCurrentTree = false;
```
bradleyayers: Shouldn't we call [`skipNode`](https://github.com/google/incremental-dom/blob/master/src/core.js#L452) here too?

treshugart: We're using the latest version. Skate's `skip` functionality is just an abstraction over Incremental DOM's which allows you to be more declarative in how you do it.

test/unit/vdom/skip.js
treshugart: Thanks @nadiam84 will add that to the test cases.

src/api/vdom.js
treshugart: K @bradleyayers updated the PR. Logic changed substantially because we now have to keep track of how many levels deep past the skipped element because `currentElement()` doesn't actually return the element we think it does. In IDOM master `currentPointer()` would be used.

test/unit/vdom/skip.js
treshugart: @nadiam84 I've updated the tests. Can you pls have a look when you have a chance? Thanks!

test/unit/vdom/skip.js
```diff
@@ -10,8 +10,14 @@ describe('vdom/skip', () => {
       },
       render() {
         vdom.element('div', { skip: true }, () => {
```
nadiam84: 'it should not affect siblings' (probably)?
```
render() {
        vdom.element('div', { skip: true }, () => {
          vdom.text('child')
        });
        vdom.element('div', () => {
          vdom.text('sibling')
        });
      },
```
```
expect(elem[symbols.shadowRoot].textContent).to.equal('sibling'),
```
fails

treshugart: Awesome. Thank you!

test/unit/vdom/skip.js
```diff
@@ -2,26 +2,69 @@ import { define, prop, props, symbols, vdom } from '../../../src/index';
...
+          elementClose('span');
+          elementOpen('div');
+            text('14 ');
+            elementOpen('span');
+              text('15');
+            elementClose('span');
+          elementClose('div');
+        elementClose('div');
       },
     });
```
nadiam84: That is one horrible looking test :D

```
render() {
        vdom.element('div', () => {
          vdom.element('div', {skip: true}, () => {
            vdom.text('1');
          })
        });
        vdom.element('div', () => {
          vdom.text('2');
        });
      },
```
```
expect(sr(elem).innerHTML).to.equal("<div><div></div></div><div>2</div>")
```
fails

treshugart: Fixed!

nadiam84: What about the situation 'a custom element inside a custom element'? Should `skip` affect another custom element's behavior?
What should happen here:
```
const Elem2 = define('x-test2', {
      render() {
        vdom.element('div', () => {
          vdom.text('2');
          vdom.element('slot');
        });
      },
    });
    const Elem = define('x-test', {
      render() {
        vdom.element('x-test2', {skip: true}, () => {
          vdom.element('div', () => {
            vdom.text('1');
          });
        });
      },
    });
```
Right now '2' and 'slot' are rendering despite `{skip: true}` for the x-test2.

treshugart: That's expected because of the shadow boundary. We expect shadow DOM and
custom elements to abide by the rules of normal DOM so I don't think we
need to make any special considerations here.

nadiam84:
lgtm

src/api/vdom.js
```diff
@@ -78,9 +77,9 @@ attributes.ref = function (elem, name, value) {
 attributes.skip = function (elem, name, value) {
   if (value) {
     skip();
-    elem[$skipCurrentElement] = true;
+    elem.__skip = true;
   } else {
-    delete elem[$skipCurrentElement];
+    delete elem.__skip;
```
bradleyayers: Wouldn't it be better to just remove the `__skip` on the way out of `elementClose`?

treshugart: Skip should remain `true` until it changes.

bradleyayers:
LGTM

treshugart:
lgtm (interesting it didn't pass this check with two maintainers doing it)

PR #734, opened 2016-08-17
Fix linked attribute going out of sync

bqx:
Setting a property to the same value twice and then trying to set the linked attribute will not update the property in native (Chrome + Chrome Canary).

This is because the second time the property setter is run, trying to set the linked attribute to the same value won't trigger the `attributeChangedCallback` callback.

### DIFF
--- a/src/lifecycle/props-init.js
+++ b/src/lifecycle/props-init.js
@@ -94,13 +94,16 @@ function createNativePropertyDefinition(name, opts) {
 
     // Link up the attribute.
     const attributeName = data(this, 'propertyLinks')[name];
-    if (attributeName && !propData.settingAttribute && valueHasChanged) {
+    if (attributeName && !propData.settingAttribute) {
       const serializedValue = opts.serialize(newValue);
       propData.syncingAttribute = true;
       if (shouldRemoveAttribute || empty(serializedValue)) {
         this.removeAttribute(attributeName);
       } else {
         this.setAttribute(attributeName, serializedValue);
+      }
+      if (!valueHasChanged && propData.syncingAttribute) {
+        propData.syncingAttribute = false;
       }
     }
 
--- a/src/lifecycle/props-init.js
+++ b/src/lifecycle/props-init.js
@@ -79,6 +79,7 @@ function createNativePropertyDefinition(name, opts) {
     propData.internalValue = newValue;
 
     const changeData = { name, newValue, oldValue };
+    const valueHasChanged = newValue !== oldValue;
 
     if (typeof opts.set === 'function') {
       opts.set(this, changeData);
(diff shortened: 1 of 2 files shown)


lukebatchelor:
lgtm

test/unit/lifecycle/properties.js
```diff
@@ -104,6 +104,39 @@ describe('lifecycle/property', () => {
         expect(elem.getAttribute('test-name')).to.equal('something else');
       });
 
+      it('setting the attribute updates the property value', done => {
```
nadiam84: maybe move this test to others like this? somewhere here: https://github.com/skatejs/skatejs/blob/master/test/unit/lifecycle/properties.js#L320

bqx: That makes sense, will do.

src/lifecycle/props-init.js
```diff
@@ -93,7 +94,7 @@ function createNativePropertyDefinition(name, opts) {
 
     // Link up the attribute.
     const attributeName = data(this, 'propertyLinks')[name];
-    if (attributeName && !propData.settingAttribute) {
+    if (attributeName && !propData.settingAttribute && valueHasChanged) {
```
nadiam84: When Skate syncing attributes and properties, it relies on `syncingAttribute` property of the `propData` (https://github.com/skatejs/skatejs/blob/master/src/api/component.js#L56)

When this situation (same property set) happens, `this.setAttribute(attributeName, serializedValue);` never fires, which means that `syncingAttribute` is never reset, which results in the problem you're fixing.
Consider this, maybe the better solution would be just reset the `syncingAttribute` value here:
https://github.com/skatejs/skatejs/blob/master/src/lifecycle/props-init.js#L102

something like this:
```
this.setAttribute(attributeName, serializedValue);
if (!valueHasChanged) {
    propData.syncingAttribute = false;
}
```

nadiam84: Or, better even, like this:
```
this.setAttribute(attributeName, serializedValue);
if (!valueHasChanged && propData.syncingAttribute) {
    propData.syncingAttribute = false;
}
```

src/lifecycle/props-init.js
```diff
@@ -79,6 +79,7 @@ function createNativePropertyDefinition(name, opts) {
     propData.internalValue = newValue;
 
     const changeData = { name, newValue, oldValue };
+    const valueHasChanged = newValue !== oldValue;
```
nadiam84: Is it possible for a newValue to be `null` and for a oldValue to be `undefined`? (or vice versa) Don't remember at which point they are coerced. Because the strict comparison wouldn't work in this case.

bqx: `newValue` is coerced before we do this equality check, so it should be fine.

nadiam84:
lgtm

test/unit/lifecycle/properties.js
bqx: For posterity: The issue has to do with the way the property `set` and the `attributeChangedCallback` callback interact, so it doesn't quite fit in either location - leaving it here should be OK.

Issue #730, opened 2016-08-15
Virtual element functions should accept numbers as text content


treshugart:
This doesn't work:

```js
vdom.element('span', 1);
```

PR #736, opened 2016-08-20
Numbers as text

nadiam84:
Issue #730 solved.
Some unrelated linting fixes.

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -284,7 +284,7 @@ export function element(tname, attrs, chren) {
   const atype = typeof attrs;
 
   // If attributes are a function, then they should be treated as children.
-  if (atype === 'function' || atype === 'string') {
+  if (atype === 'function' || atype === 'string' || atype === 'number') {
     chren = attrs;
   }
 
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -1,7 +1,6 @@
 import {
   applyProp,
   attributes,
-  currentElement,
   elementClose,
   elementOpen,
   elementVoid,
@@ -55,13 +54,13 @@ function applyEvent(elem, ename, newFunc) {
 }
 
 // Attributes that are not handled by Incremental DOM.
-attributes.key = attributes.statics = function () {};
+attributes.key = attributes.statics = () => {};
 
 // Attributes that *must* be set via a property on all elements.
 attributes.checked = attributes.className = attributes.disabled = attributes.value = applyProp;
 
 // V0 Shadow DOM to V1 normalisation.
-attributes.name = function (elem, name, value) {
+attributes.name = (elem, name, value) => {
   if (elem.tagName === 'CONTENT') {
     name = 'select';
     value = `[slot="${value}"]`;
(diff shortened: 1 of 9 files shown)


test/unit/vdom/elements.js
```diff
@@ -184,58 +184,62 @@ describe('vdom/elements', () => {
...
-
-    it('div > *span* > text', done => {
-      const Span = (props, chren) => vdom.element('span', chren);
-      const Div = () => vdom.element('div', () => vdom.element(Span, 'text'));
-      const Elem = element().skate({
-        render() {
-          vdom.element(Div);
-        },
+      it(`div > *span* > ${ch}`, done => {
+        const Span = (props, chren) => vdom.element('span', chren);
```
bengummer: this line is repeated from line 190 above

nadiam84: I just wrapped those tests into the funtion

test/unit/vdom/ref.js
```diff
@@ -76,9 +76,9 @@ describe('vdom/ref', () => {
   test('custom elements', define('x-test', {
     render() {
       vdom.element('slot');
-    }
+    },
   }));
-  test('function helpers', (props, chren) => {
-    vdom.element('div', props, chren);
+  test('function helpers', (lprops, chren) => {
+    vdom.element('div', lprops, chren);
```
bengummer: not obvious what the "L" in `lprops` means here. no big deal :)

bengummer:
lgtm

src/api/vdom.js
```diff
@@ -153,7 +152,7 @@ function elementOpenStart(...args) {
 }
 
 function elementOpenEnd() {
-  const node = newElementOpen(...overrideArgs);
+  const node = newElementOpen(...overrideArgs); // eslint-disable-line no-use-before-define
```
lukebatchelor: Where is ...overrideArgs coming from?

test/unit/vdom/ref.js
nadiam84: 'local' :D It's a linting fix, there is global 'props' already here

src/api/vdom.js
nadiam84: no idea, it'a a linting fix

test/unit/vdom/ref.js
treshugart: Airbnb is killing me.

src/api/vdom.js
lukebatchelor: I feel like this is a legitimate warning...  @treshugart ?

treshugart: It's declared [here](https://github.com/skatejs/skatejs/pull/736/files#diff-9793dbfbf12d7bba475ec4c3ded46adeL31) and it has a comment telling why it's necessary. They do a similar thing in their source. Airbnb shouldn't be complaining here.

treshugart: It could be complaining due to function hoisting, in which case doing `const elementOpenEnd = ...` should fix it. But I feel like we're having to jump through unnecessary hoops just to satisfy an overexcited linting rule.

test/unit/api/define.js
```diff
@@ -5,9 +5,9 @@ const { name: $name } = symbols;
 
 function mockDefine(name) {
   if (customElementsV1) {
-    window.customElements.define(name, function() {});
+    window.customElements.define(name, () => {});
```
treshugart: This is a perfect example of where Airbnb rules are just ludicrous. These functions require constructors (not arrow functions).

treshugart:
lgtm pending comment reviews, but ffs, I'm thinking that someone must have come up with a slightly more common sense airbnb configuration that we could use instead at some point.

Issue #733, opened 2016-08-16
Setting a property to the same value twice breaks setting the value via a linked attribute


bqx:
It looks like setting a property to the same value twice is setting the `syncingAttribute` flag to true, but not reaching the `attributeChangedCallback` where it sets the flag to false, because the attribute value is the same.

A fix would most likely involve checking that the new property value is not the same as the old one, and not setting the flag/trying to set the attribute if that is the case.

treshugart:
Fixed in #734. Thanks @bqx!

PR #739, opened 2016-08-23
fix: Fix for setting a linked attribute not updating the property

### LINKED ISSUE #738: Cannot set property via linked attribute
This happens when a property has a linked attribute, and then the value of the property is immediately set to the same value as its initial value once it is initialised (e.g. during the created callback, or immediately after instantiating an instance of the component):

This seems to be the same issue as #733 , except that it occurs when setting to the initial value again immediately.

Example:

```
define('my-component', {
  created(el) {
    el.myProp = 'initial value';
  },
  props: {
    myProp: {
      attribute: true
    }
  }
});
```

```
<my-component my-prop="initial value"></my-component>
```
```
document.querySelector('my-component').setAttribute('my-prop', 'new value'); // does nothing
```

bqx:
Followup to the fix for #733, check that the actual attribute value has changed after serialization

Fixes #738

---

The issue was that we were only checking that value had changed, not that the attribute value had changed (which was the root cause of the problem). We need to check that the new attribute value (i.e. `serializedValue`) is not new.

Also moved tests to a new section at the bottom of the file, and added one new test.

### DIFF
--- a/src/lifecycle/props-init.js
+++ b/src/lifecycle/props-init.js
@@ -79,7 +79,6 @@ function createNativePropertyDefinition(name, opts) {
     propData.internalValue = newValue;
 
     const changeData = { name, newValue, oldValue };
-    const valueHasChanged = newValue !== oldValue;
 
     if (typeof opts.set === 'function') {
       opts.set(this, changeData);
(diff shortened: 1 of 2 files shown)


src/lifecycle/props-init.js
```diff
@@ -96,13 +95,16 @@ function createNativePropertyDefinition(name, opts) {
     const attributeName = data(this, 'propertyLinks')[name];
     if (attributeName && !propData.settingAttribute) {
       const serializedValue = opts.serialize(newValue);
+      const currentAttrValue = this.getAttribute(attributeName);
+      const attributeChanged = !((empty(serializedValue) && empty(currentAttrValue)) ||
+                                (serializedValue === currentAttrValue));
       propData.syncingAttribute = true;
       if (shouldRemoveAttribute || empty(serializedValue)) {
```
treshugart: You can probably store the check to `empty(serializedValue)` since it's used more than once.

treshugart:
LGTM pending comments!

lukebatchelor:
Cool! LGTM

bqx:
LGTM

treshugart:
lgtm lgtm lgtm lgtm

PR #746, opened 2016-08-25
fix: play nice with other libraries touching the idom.attributes global

### LINKED ISSUE #745: Multiple skate versions are incompatible when idom.attributes mutation differs
There's a fundamental problem in living up to

> It works with multiple versions of itself on the page.

due to how skate hooks into incremental-dom. incremental-dom exposes an `attributes` object which can be used to hook into the lifecycle of DOM updates, and skate leverages this to do some internal state tracking.

Unfortunately skate currently assumes that it can safely mutate `attributes.___`, but it forgot that it'll actually be stepping on other versions of itself.

This happened in practice with v1.0.0-beta12 and v1.2.3, because back in v1.0.0-beta12 we had the code:

```
// Attributes that are not handled by Incremental DOM.
attributes.key = attributes.skip = attributes.statics = function () {};
```

but in v1.2.3 we *do* need `attributes.skip` so we have:

```
// Skip handler.
incrementalDom.attributes.skip = function (elem, name, value) {
  if (value) {
    elem[$skip] = true;
  } else {
    delete elem[$skip];
  }
};
```

Unfortunately for me v1.0.0-beta12 was being executed *after* v1.2.3 and clobbering the `attributes.skip` hook.

bradleyayers:
fixes #745

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -9,6 +9,7 @@ import {
 } from 'incremental-dom';
 import { name as $name, ref as $ref } from '../util/symbols';
 import { shadowDomV0, shadowDomV1 } from '../util/support';
+import propContext from '../util/prop-context';
 
 const applyDefault = attributes[symbols.default];
 const fallbackToV0 = !shadowDomV1 && shadowDomV0;
@@ -31,6 +32,8 @@ let overrideArgs;
 // The number of levels deep after skipping a tree.
 let skips = 0;
 
+const noop = () => {};
+
 // Adds or removes an event listener for an element.
 function applyEvent(elem, ename, newFunc) {
   let events = elem[$currentEventHandlers];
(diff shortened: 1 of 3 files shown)


src/api/vdom.js
```diff
@@ -258,7 +267,7 @@ const newElementOpenEnd = wrapIdomFunc(elementOpenEnd);
 
 // Standard open / closed overrides don't need to reproduce internal behaviour
 // because they are the ones referenced from *End and *Start.
-const newElementOpen = wrapIdomFunc(elementOpen, stackOpen);
+const newElementOpen = attributesContext(wrapIdomFunc(elementOpen, stackOpen));
```
treshugart: How can we be sure that this is the only function we need to wrap? Are there any performance implications with this do you think?

src/util/prop-context.js
```diff
@@ -0,0 +1,27 @@
+import assign from './assign';
+
+function enter(object, props) {
+  const saved = {};
+  Object.keys(props).forEach(key => {
```
treshugart: Could this be a `reduce()`?

treshugart:
lgtm pending comments

src/api/vdom.js
bradleyayers: Yes this is the only function. There might be performance implications, but I think at this point they're of less concern than the multi-version incompatibility, and something we could improve later.

src/util/prop-context.js
bradleyayers: Semantically `reduce` is used to build and return a *new* object based on collection, but in this case that isn't our goal. Instead we want to mutate and existing one (`object`).

We could certainly *exploit* the looping behaviour of `reduce` for this, but unless it brings a perf improvement I don't think it would be mislead the reader.

src/api/vdom.js
bradleyayers: Perhaps a useful upstream change to incremental-dom would be to allow the entire attributes object to be changed, rather than just attributes on it. This would allow context switches to be faster (single assignment operation) rather than being proportional to the number of attribute overrides.

bradleyayers:
lgtm

src/api/vdom.js
treshugart: The idea would be to pass `attributes` in to every patch and patch would look to that as the source of truth instead of the global.

src/util/prop-context.js
treshugart: Yeah, I can see that now.

Issue #742, opened 2016-08-24
Compatibility with webreflection/document-register-element


treshugart:
cc @webreflection in case you're interested

WebReflection:
what's the current "_incompatibility_" if I might ask?

is this about using alternatively my polyfill instead of the web components one?

treshugart:
> what's the current "incompatibility" if I might ask?

The one I've just fixed would be that Skate also uses v0 if available and had defined `createdCallback` et al in it's `Component` base class to be called in v0-land. Your fill defined this on the element that gets returned from `super()` and when I tried to call it, it was calling yours instead of mine. I've just moved the logic in to a separate function to get around it. There's some others and I'll annotate those as I go along.

> is this about using alternatively my polyfill instead of the web components one?

Quite possibly, but either way it'd be nice to support both so we can direct Skate users to polyfills that fit their needs. There's still a few things we would need to consider before including it with skatejs/web-components by default. Some things I haven't measured yet are performance and size.

WebReflection:
>  I've just moved the logic in to a separate function to get around it. 
that part changed already locally to indeed avoid overrides when already defined.

You should've filed a bug in my repo ;-) sorry for the inconvenience.

treshugart:
> You should've filed a bug in my repo ;-) sorry for the inconvenience.

I don't really consider it a bug with your repo, though. We should be being defensive about this anyways.

If I do find what I think are legitimate bugs, of course I'll file them with you repo.

WebReflection:
let me know if there's anything else I can help with

treshugart:
Sure, thing. Thanks for responding :)

treshugart:
@WebReflection in https://github.com/WebReflection/document-register-element/pull/67#issuecomment-242031456 you mentioned FF still had some issues? I'm currently seeing some issues where the call to `nodeNames.get()` returns `undefined` at https://github.com/WebReflection/document-register-element/blob/master/src/document-register-element.js#L896. I'm still debugging the issue (could be my fault), but maybe this is something you've already seen?

WebReflection:
please try its latest and let me know . FF Nightly breaks V0 incapable of accepting a string as second argument + other breaks. FF Desktop also has some glitch but right now I have no time to debug it.

WebReflection:
FYI I've apparently fixed every FireFox hiccup with `1.0.7`, you can test pretty much every known browser on this page and see green light: https://webreflection.github.io/document-register-element/test/

[...]

PR #757, opened 2016-08-31
#742 - add support for webreflection/document-register-element

treshugart:
WIP: 

- [ ] Wait for skatejs/web-components#7 to be merged.
- [ ] Update tests testing class extension to not run in IE 9 and 10 since it doesn't support static property extension. Consumers supporting these browsers should use `.extends()` instead.
- [ ] Add IE9 and 10 to Browserstack.

Still retains support for native and the webcomponents.js v1 polyfill. Adds support for IE 9 and 10 mostly through just consuming the `document-register-element1` polyfill.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Whenever you change the `name` property - or attribute - the component will re-r
     - [NPM](#npm)
     - [Script Tag](#script-tag)
   - [Dependencies](#dependencies)
+  - [Browser Support](#browser-support)
   - [Resources](#resources)
   - [Questions?](#questions)
   - [Terminology](#terminology)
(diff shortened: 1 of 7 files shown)


jpnelson:
lgtm

will we test with both versions of skatejs-web-components?

treshugart:
@jpnelson we probably should. I'll have a look at doing that. We really shouldn't *have* to, but document-register-element passes the element as an argument to the constructor in IE, I think, which requires us to write custom code for that. We could instead write a test for that integration point specifically, but might be better to just do both polyfills as you said.

treshugart:
Merging as it seems the Browserstack build has finally passed.

treshugart:
lgtm

PR #759, opened 2016-08-31
fix: Apply skate semantics to element attributes in a stack context.

### LINKED ISSUE #756: className="..." broken for 1.2.4
In 1.2.4 (came in via #746) the following logic seems to be broken:

```js
import { vdom } from 'skatejs';
import shadowStyles from './shadow.less';

export default (props, children) => (
  <a
    tabindex="-1"
    className={shadowStyles.locals.href}
    href={props.href}
  >
      {children()}
  </a>
);
```

this rendered like:
```html
<a class="_1Ay9htsj9kSvVouo0C0qZE" href="x" tabindex="-1">Carrot cake</a>
```
in 1.2.3, but renders like:

```html
<a classname="_1Ay9htsj9kSvVouo0C0qZE" href="x" tabindex="-1">Carrot cake</a>
```

in 1.2.4 (please note the `classname` attribute instead of `class`).

Spreading the props out on the element seems to fix it for some reason, so if the template function gets changed to:
```js
export default (props, children) => (
  <a
    {...props}
    tabindex="-1"
    className={shadowStyles.locals.href}
    href={props.href}
  >
      {children()}
  </a>
);
```
it correctly changes `className` to `class`.

bradleyayers:
Fixes #756

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -2,7 +2,7 @@ import {
   applyProp,
   attributes,
   elementClose,
-  elementOpen,
+  elementOpen as idomElementOpen,
   skip,
   symbols,
   text,
@@ -155,6 +155,10 @@ function resolveTagName(tname) {
   return tname;
 }
 
+// Incremental DOM's elementOpen is where the hooks in `attributes` are applied,
+// so it's the only function we need to execute in the context of our attributes.
+const elementOpen = attributesContext(idomElementOpen);
+
 function elementOpenStart(tag, key = null, statics = null) {
   overrideArgs = [tag, key, statics];
 }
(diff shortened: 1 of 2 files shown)


treshugart:
lgtm

bradleyayers:
Fundamentally the problem was that there was a case where we were using Incremental DOM's `elementOpen`, rather than a version that would execute in the context of our attribute hooks.

This was possible because we didn't actually wrap `elementOpen`, but instead wrapped the `newElementOpen` entry point, and left `elementOpen` alone:

    const newElementOpen = attributesContext(wrapIdomFunc(elementOpen, stackOpen));

`wrapIdomFunc` is given Incremental DOM's `elementOpen` function rather than one wrapped in `atributesContext`.

At the time I thought this was fine, because I thought it would always be executed synchronously during a call to `newElementOpen`, but this was false. When passing children to function helpers, `wrapIdomFunc` builds up a stack of API calls, and as part of that will stash the raw `elementOpen` function on the stack. It's executed some time later (inside `stackClose`), but critically *not* inside a function wrapped with `attributesContext`.

The fix here is to wrap `elementOpen` directly, and pass that around everywhere.

bradleyayers:
lgtm

joscha:
Nice work @bradleyayers!

PR #760, opened 2016-09-01
#687 - Allow symbols in props

### LINKED ISSUE #687: Allow usage of symbols in props
Currently you cannot - but should be able to - do:

```js
const sym = Symbol();
skate.define('x-test', {
  props: {
    [sym]: {}
  }
});
```

mcorbanini:
Add support for Symbols in props. Make use of `createSymbol` function which fallback returning the argument used as the Symbol's description if it is not supported in the current browser.

fixes #687


src/lifecycle/props-init.js
```diff
@@ -14,14 +14,18 @@ function getInitialValue(elem, name, opts) {
   return typeof opts.initial === 'function' ? opts.initial(elem, { name }) : opts.initial;
 }
 
+function getPropNamespace(name) {
```
treshugart: 👍 thanks :)

treshugart: You could abstract this further into a data function for namespaced props rather than having to repeat the `getPropNamespace()` call.

src/util/get-all-keys.js
```diff
@@ -0,0 +1,10 @@
+
+export default function keys(obj, enumerable = false) {
```
treshugart: I know it may exist in the code elsewhere, but I've been a fan of not falling into the [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap) and using an options object instead.

You could define this as:

```js
keys(obj, { enumerable = false });
```

src/util/get-all-keys.js
```diff
@@ -0,0 +1,10 @@
...
+  let listOfKeys = [
+    ...Object[enumerable ? 'getOwnPropertyNames' : 'keys'](obj),
```
treshugart: This will probably turn it into an `Array.from()` call and we currently don't import the polyfill. That said, the `Object` methods should return an `Array` anyways, right? Do we even need to spread here?

src/util/symbols.js
```diff
@@ -1,9 +1,13 @@
-export const connected = '____skate_connected';
-export const created = '____skate_created';
-export const name = '____skate_name';
-export const props = '____skate_props';
-export const ref = '____skate_ref';
-export const renderer = '____skate_renderer';
-export const rendering = '____skate_rendering';
-export const rendererDebounced = '____skate_rendererDebounced';
-export const shadowRoot = '____skate_shadowRoot';
+function createSymbol(name) {
```
treshugart: Nice!

treshugart: I'd be a fan of randomly generating a string here, even if it's done simply so you don't have to pass it in everywhere.

treshugart:
Nice work, @mcorbanini! Great to see this PR come in.

src/util/get-all-keys.js
zzarcon: Both `getOwnPropertyNames` and `keys` return an array as result, so I think @treshugart is right:

```javascript
const listOfKeys = Object[enumerable ? 'getOwnPropertyNames' : 'keys'](obj);
```
Also **const** is maybe more correct here? later we can just do

```javascript
listOfKeys.push(...Object.getOwnPropertySymbols(obj))
```

Instead of mutating the variable?

treshugart: `push()` mutates, too, so seems like one evil for another.

zzarcon: You are right, sorry I meant mutating the reference ^^. Which I think it's more aligned when using **const** isn't?

src/util/get-all-keys.js
mcorbanini: I agree. Will change that.

src/util/get-all-keys.js
mcorbanini: Thanks for the feedback! Yes, this was a leftover of a previous commit, since I was creating an array like ```[...array1, ...array2]```.
Now it doesn't makes much sense so I agree on just having 
```javascript 
const listOfKeys = Object[enumerable ? 'getOwnPropertyNames' : 'keys'](obj);
``` 

and afterwards, if it meets the condition, will push the symbols as @zzarcon suggest:
```javascript 
listOfKeys.push(...Object.getOwnPropertySymbols(obj))
```

src/util/symbols.js
mcorbanini: For sure!

src/util/get-all-keys.js
```diff
@@ -1,10 +1,9 @@
...
-export default function keys(obj, enumerable = false) {
-  let listOfKeys = [
-    ...Object[enumerable ? 'getOwnPropertyNames' : 'keys'](obj),
-  ];
+export default function keys(obj, { enumerable = false } = {}) {
+  const method = enumerable ? 'getOwnPropertyNames' : 'keys';
+  const listOfKeys = Object[method](obj);
   if (typeof Object.getOwnPropertySymbols === 'function') {
-    listOfKeys = listOfKeys.concat(Object.getOwnPropertySymbols(obj));
+    listOfKeys.push(...Object.getOwnPropertySymbols(obj));
```
treshugart: nit: using concat is declarative and doing something like:

```js
return typeof Object.getOwnPropertySymbols === 'function' ?
  listofKeys.concat(Object.getOwnPropertySymbols(obj)) :
  listOfKeys;
```

would be a more functional way of doing it.

mcorbanini: Yes you are right there, thanks!

src/util/symbols.js
mcorbanini: @treshugart Actually, do we need to add a description to a `Symbol`?  Is it mostly useful for debugging, but if we create a random string, and you are in a context where Symbol is not supported, it will be difficult to know. Maybe we can keep those since they are more descriptive ?

test/unit/api/props.js
```diff
@@ -5,6 +5,7 @@ import { props } from '../../../src/index';
 
 describe('api/props', () => {
   let elem;
+  const secret = typeof Symbol === 'function' ? Symbol('secret') : 'secret';
```
joscha: use `createSymbol` here?

mcorbanini: Yes, will need to add that to a `util` then.

src/util/createSymbol.js
```diff
@@ -0,0 +1,3 @@
+export default function createSymbol(description) {
```
treshugart: File name: `createSymbol` -> `create-symbol`

treshugart: Maybe should write a test for this?

treshugart:
Hmm, I just re-ran the build after updating master. 1.3.0 released fine fro, all tests passing in IE9+ but seems one of the tests here is failing in IE10? Maybe worth re-running after you've finished any work you have left.

treshugart:
lgtm once the build's green

test/unit/util/get-all-keys.js
```diff
@@ -0,0 +1,40 @@
...
+
+  it('should return symbol keys from an object', () => {
+    const secret = createSymbol('secret');
+    expect(keys({ foo: 1, [secret]: 2 }))
+      .to.have.lengthOf(2)
+      .and.to.deep.equal(['foo', secret]);
+  });
+
+  describe('when enumerable option is set', () => {
+    const obj = {};
```
bengummer: This could just be `let obj;` and then inside the `beforeEach` block it can be

```js
obj = {};
Object.defineProperties(obj, {
  foo: { enumerable: true, value: 1 },
  bar: { enumerable: false, value: 2 },
})
```

Probably no side effects the way it is, but safer + clearer to redefine each time.

test/unit/lifecycle/properties.js
```diff
@@ -41,26 +41,15 @@ describe('lifecycle/property', () => {
...
-
-    elem.removeAttribute('test1');
-    elem.removeAttribute('test2');
-
-    expect(elem.test1).to.equal('test1');
-    expect(elem.test2).to.equal('test2');
-
-    expect(elem.getAttribute('test1')).to.equal(null);
-    expect(elem.getAttribute('test2')).to.equal(null);
+    ['test1', 'test2'].forEach(value => {
```
bengummer: could add a Symbol prop in this test as well

bengummer:
2 optional comments, lgtm 🎉

src/util/get-all-keys.js
```diff
@@ -0,0 +1,7 @@
+
+export default function keys(obj = {}, { enumerable = false } = {}) {
+  const listOfKeys = Object[enumerable ? 'getOwnPropertyNames' : 'keys'](obj);
```
nadiam84: Shouldn't this be the other way around? 
```
!enumerable ? 'getOwnPropertyNames' : 'keys'
```

test/unit/util/get-all-keys.js
```diff
@@ -0,0 +1,40 @@
...
+    beforeEach(() =>
+      Object.defineProperties(obj, {
+        foo: { enumerable: true, value: 1 },
+        bar: { enumerable: false, value: 2 },
+      })
+    );
+
+    describe('to false', () => {
+      it('should not return enumerable keys', () =>
```
nadiam84: It tests the exact opposite of the description. If it should not return enumerable keys, then only the 'bar' should be returned.
Why btw you want to return non-enumerables only? Why not 'enumerables' and 'all'?

src/util/get-all-keys.js
mcorbanini: True, it's confusing because the first one returns non-enumerable and enumerable. And the second one only enumerable. Will change that for consistency.

test/unit/util/get-all-keys.js
mcorbanini: It's just a bad name for a flag in this case 👅 
Code using this function now, was using either one of those methods. Once I change the `enumerable` flag will change the test for sure 👍

test/unit/lifecycle/properties.js
mcorbanini: 👍

src/util/get-all-keys.js
treshugart: Might be good to invert the clause for readability so your mind doesn't have to invert the variable value.

mcorbanini: @treshugart @nadiam84 `getOwnPropertyNames` give you all props enum and non-enum. `keys` only give you enum. What do you guys think to just rename the boolean flag to `enumOnly` or `onlyEnumerable` (help with the name) and flip the return values (`enumerable ? 'keys' : 'getOwnPropertyNames'`) ?

treshugart: 👍

nadiam84: like `enumOnly`

test/unit/lifecycle/properties.js
mcorbanini: Already discussed about this, and we shouldn't be able to add a symbol `attribute: true` property. So it doesn't make sense to add a test here with symbol prop. Check #687

joscha:
LGTM

joscha:
LGTM

treshugart:
let's try this again: lgtm

joscha:
LGTM

joscha:
I've reran these tests numerous times now - BrowserStack seems to time out every single time - maybe some connection timeout needs to be increased?

lukebatchelor:
:shipit:

treshugart:
lgtm

joscha:
wohoo!

PR #767, opened 2016-09-05
Fix case where event is not emitted outside of the shadow dom in Chrome 53 (native v1 shadow dom)

craigpetchell:
This fails for Chrome (53.0.2785), passes for Firefox (48.0.0), so needs a fix to move forward.

**DO NOT MERGE** if you want a green build. :)

### DIFF
--- a/src/api/emit.js
+++ b/src/api/emit.js
@@ -10,11 +10,16 @@ const CustomEvent = ((Event) => {
 })(window.CustomEvent);
 
 function createCustomEvent(name, opts = {}) {
-  if (CustomEvent) {
-    return new CustomEvent(name, opts);
+  let e;
+  if (Event) {
+    e = new Event(name, opts);
+    if (opts.detail) {
+      Object.defineProperty(e, 'detail', { value: opts.detail });
+    }
+  } else {
+    e = document.createEvent('CustomEvent');
+    e.initCustomEvent(name, opts.bubbles, opts.cancelable, opts.detail);
   }
-  const e = document.createEvent('CustomEvent');
-  e.initCustomEvent(name, opts.bubbles, opts.cancelable, opts.detail);
   return e;
 }
 
@@ -24,6 +29,9 @@ export default function (elem, name, opts = {}) {
   }
   if (opts.cancelable === undefined) {
     opts.cancelable = true;
+  }
+  if (opts.composed === undefined) {
+    opts.composed = true;
   }
   return elem.dispatchEvent(createCustomEvent(name, opts));
 }
--- a/src/api/emit.js
+++ b/src/api/emit.js
@@ -13,7 +13,7 @@ function createCustomEvent(name, opts = {}) {
   let e;
   if (Event) {
     e = new Event(name, opts);
-    if (opts.detail) {
+    if ('detail' in opts) {
       Object.defineProperty(e, 'detail', { value: opts.detail });
     }
   } else {
--- a/src/api/emit.js
+++ b/src/api/emit.js
@@ -10,15 +10,18 @@ const CustomEvent = ((Event) => {
 })(window.CustomEvent);
 
 function createCustomEvent(name, opts = {}) {
+  const { detail } = opts;
+  delete opts.detail;
+
   let e;
   if (Event) {
     e = new Event(name, opts);
-    if ('detail' in opts) {
-      Object.defineProperty(e, 'detail', { value: opts.detail });
+    if (typeof detail !== 'undefined') {
+      Object.defineProperty(e, 'detail', { value: detail });
     }
   } else {
     e = document.createEvent('CustomEvent');
-    e.initCustomEvent(name, opts.bubbles, opts.cancelable, opts.detail);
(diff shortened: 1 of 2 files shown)


treshugart:
As discussed IRL, setting `{ composed: true }` works in native Shadow DOM v1. That's probably the fix we want as if intentionally didn't add it back and made the user specify it, it would be breaking current expectations (even if it's not our API).

treshugart:
Also, looks like the build is passing. We should probably add Chrome 53 to the list of browsers to make it fail.

craigpetchell:
And browser stack only seem to go up to stable 52 right now. :/

It looks like skate build is setup to run again the latest anyway, it just seems to only be 52.

treshugart:
Right ok, wasn't sure if they just hadn't bumped latest or something.

src/api/emit.js
```diff
@@ -10,11 +10,16 @@ const CustomEvent = ((Event) => {
 })(window.CustomEvent);
 
 function createCustomEvent(name, opts = {}) {
-  if (CustomEvent) {
-    return new CustomEvent(name, opts);
+  let e;
+  if (Event) {
+    e = new Event(name, opts);
+    if (opts.detail) {
```
treshugart: Might be okay to leave this out, or do `'detail' in opts` so that falsy values can still be passed.

treshugart:
lgtm

src/api/emit.js
```diff
@@ -10,11 +10,16 @@ const CustomEvent = ((Event) => {
 })(window.CustomEvent);
 
 function createCustomEvent(name, opts = {}) {
-  if (CustomEvent) {
-    return new CustomEvent(name, opts);
+  let e;
+  if (Event) {
+    e = new Event(name, opts);
```
bradleyayers: `opts` can contain more than just `bubbles` and `cancelable`, so I would probably pull those out and pass them explicitly.

src/api/emit.js
```diff
@@ -25,5 +30,8 @@ export default function (elem, name, opts = {}) {
   if (opts.cancelable === undefined) {
     opts.cancelable = true;
   }
+  if (opts.composed === undefined) {
```
bradleyayers: Where does `composed` come from?

treshugart: @bradleyayers Not sure it's officially documented yet. The only real evidence of agreement we have is https://github.com/w3c/webcomponents/issues/513 and the fact that it landed in v1 stable in both Blink and Webkit. They mention it in the [DOM standard](https://dom.spec.whatwg.org) but it's somewhat out of context. I also [asked Rob Dodson](https://twitter.com/treshugart/status/772656495167086593) about it and he gave a link to https://developers.google.com/web/fundamentals/primers/shadowdom/#customevents which is close to what we're doing here.

Though it would seem that it's nice to not have it bubble out by default, I would think this would be considered breaking for Skate, so maybe we can invert this in 2.0? If so, I'll create an issue for that.

bradleyayers: I am in favour of them not bubbling out by default.

craigpetchell: Definitely a breaking change, since they bubble outside of the "shadow dom" in the polyfill world. I suppose there should be a polyfill change to stop bubbling outside of the shadow dom at some point too.

src/api/emit.js
craigpetchell: @bradleyayers - I'm not totally sure what you're saying here. 

Are you saying anything in opts we know the constructor doesn't use (currently!?), explicitly add as a property on the event (like we do for detail).

Or only pass opts to the constructor we know it will take? Or some other variation.

src/api/emit.js
treshugart: I think it's worth fixing support for the 1.x releases and raising an issue and possibly going 2.0 soonish.

treshugart:
@craigpetchell I just raised #768 so that we can continue with your fix here and then change it for 2.x.

src/api/emit.js
bradleyayers: > Or only pass opts to the constructor we know it will take? 

This is what I was talking about.

craigpetchell: I can do that, my only concern would be we'd need to update this "whitelist", if the constructor does take new parameters in the future, where as (I assumed) it now just ignores the extras.

treshugart: I think not having to maintain a whitelist is reasonable here. If anything, we should see a red build if something isn't working right.

bradleyayers: Far enough. I suppose I would have just added a `const { detail } = ops; delete opts.detail` to avoid passing `detail`. Passing random data into external APIs is generally a bad idea.

treshugart: Ahh I see what you mean, that's a reasonable thing to do, I think.

craigpetchell: The spread operator makes this nice, but needs a babel transform to support it.

I'll follow Bradley's approach.

jpnelson:
lgtm

craigpetchell:
@jpnelson - I've covered everything off (afaik), so it just needs some release/merge magic.

PR #778, opened 2016-09-08
#770 - fix: Fix errors when passing falsy values as the attrs arg to vdom.el…

### LINKED ISSUE #770: Vdom attributes should support a null / falsy value and do nothing with it
This fails: http://www.webpackbin.com/EJrn9OOj-.

treshugart:
…ement().

Fixes #770


bradleyayers:
In the example http://www.webpackbin.com/NJ2NU7NhW I don't see `vdom.element` being used.

As such it's not clear to me what the use-case is. In general I'm not in favour of blurry APIs like this. If we we want support for `null`, then we should add support for `null`, and we should test for that, but we shouldn't be testing that `''` and `0` also work — I would say those should not be part of the API.

treshugart:
True, but the issue is for falsy values; `null` was just mentioned.

bradleyayers:
My fundamental issue is I don't understand the use-case for a falsey value, so it makes it hard for me to be in favour of this.

treshugart:
Since we handle other values explicitly for attributes and then set them as text, we can only worry about handling `null` here as you've suggested.

bradleyayers:
lgtm

nadiam84:
lgtm

PR #784, opened 2016-09-13
fix: skip ES2015 class tests that rely on statics inheritance in IE9 and 10

bradleyayers:
This fixes the test CI by skipping a few tests in IE9 and 10.

### DIFF
--- a/test/unit/lifecycle/properties.js
+++ b/test/unit/lifecycle/properties.js
@@ -28,7 +28,7 @@ describe('lifecycle/property', () => {
   describe('props declared as attributes with ES2015 classes are linked', () => {
     const skip = !classStaticsInheritance();
 
-    it('uses the same attribute and property name for lower-case names', (done) => {
+    it('uses the same attribute and property name for lower-case names', function(done) {
       if (skip) this.skip();
 
       const elem = new (element().skate(class extends Component {
@@ -44,7 +44,7 @@ describe('lifecycle/property', () => {
       );
     });
 
-    it('uses the same attribute and property name for dashed-names names', (done) => {
+    it('uses the same attribute and property name for dashed-names names', function(done) {
       if (skip) this.skip();
 
       const elem = new (element().skate(class extends Component {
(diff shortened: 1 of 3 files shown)


bengummer:
lgtm

test/unit/lifecycle/properties.js
```diff
@@ -25,7 +26,11 @@ describe('lifecycle/property', () => {
   });
 
   describe('props declared as attributes with ES2015 classes are linked', () => {
+    const skip = !classStaticsInheritance();
```
treshugart: You can just check if `Object.setPrototypeOf` exists but you're probably right in doing this as it's more descriptive.

treshugart:
lgtm

test/unit/lifecycle/properties.js
stevemao: `Object.setPrototypeOf` is es6, so it compiles class down to es6 code...

treshugart: @stevemao I'm not sure I follow. Are you saying `Object.setPrototypeOf` won't work?

bengummer:
I just re-ran the tests for this PR and they completed this time, but with failures https://travis-ci.org/skatejs/skatejs/builds/159560726

bradleyayers:
Thanks. Silly last-minute refactor to revert some changes I made that I later thought were unnecessary, but were indeed necessary. I've pushed changes, will see how CI goes this time.

test/unit/lifecycle/properties.js
stevemao: I'm saying `Object.setPrototypeOf` is ES6, class is also ES6. So babel compiles class to `Object.setPrototypeOf`, not some ES5 code?

treshugart: Is that relevant? `setPrototypeOf` won't exist in older browsers, so it's just a simpler way of checking.

stevemao: Not saying it won't work. Just surprised that it compiles ES6 code to ES6. Don't know how it can increase browser support.

treshugart: Oh, right, not sure. I do know many browsers shipped `setPrototypeOf` before the class syntax, though.

Issue #751, opened 2016-08-29
Some libs may return a function from render()


treshugart:
If so, we might think about checking the return value from `render()` and calling it if it's a function. If not and you use something like https://github.com/thejameskyle/incremental-dom-react-helper, you have to do:

```js
// Note the extra "()"
return (<span />)();
```

treshugart:
Closed as it was fixed in #777.

PR #786, opened 2016-09-13
#777 - feat(vdom): Add ability to create element functions using vdom.elemen…

### LINKED ISSUE #777: Add vdom utility for exporting element functions
This would be useful for cleaning up code that calls `vdom.element()` directly:

```js
vdom.element('p', () => {
  vdom.element('a', 'test');
  vdom.element('span', 'test');
});
```

And with this you could do:

```js
const [ a, p, span ] = vdom.elements('a', 'p', 'span');
p(() => {
  a('test');
  span('test');
});
```

This should also work with custom element constructors and stateless components:

```js
const CustomElement = define('x-test', {
  render() {
    vdom.element('slot');
  }
});
const CustomStateless = (props, chren) => (
  vdom.element('p', chren)
);

const [ customElement, customStateless ] = vdom.elements(CustomElement, CustomStateless);

customElement(() => (
  customStateless
));
```

The code to do this that I've used before is something like:

```js
const elements = (...names) => names.map(name => vdom.element.bind(null, name));
```

treshugart:
…ts().

Allows ability to create functions so you can just type p(), div() etc.

Resolves #777 

---

I've found this to be immensely helpful when using the vdom functions to compose a DOM structure and found myself typing it every time I was PoCing something.


README.md
```diff
@@ -1217,6 +1217,56 @@ const MyElement = (props, chren) => <div>Hello, {chren()}!</div>;
 
 
 
+#### `vdom.elements (...elements)`
+
+Returns an array of functions that create elements.
+
+```js
+const [ div, p ] = skate.vdom.elements('div', 'p');
+div(p.bind('test'));
```
bradleyayers: This example made me ponder for a while why it wasn't just `div(p('test'))`.

README.md
```diff
@@ -1217,6 +1217,56 @@ const MyElement = (props, chren) => <div>Hello, {chren()}!</div>;
...
+
+```js
+const [ div, myFunc, myComponent ] = skate.vdom.elements(
+  'div',
+  (props, chren) => skate.vdom.element('p', props, chren),
+  define('my-component', {})
+);
+div(() => {
+  myFunc(() => {
+    xTest('test');
```
bradleyayers: Why wouldn't this be `xTest.bind('test')` like the previous example?

README.md
treshugart: Do you think it should be:

```js
div(() => {
  p('test');
});
```

...instead?

README.md
treshugart: That's a good point. I did that and having more than one level seemed to read easier.

README.md
bradleyayers: I'd probably prefer something like `vdom.element(div(p('test')));`

treshugart: I like that you suggested that. I've thought about that in the past but never did it because you have to return a function, but just realised that essentially that's what we have to do anyways by passing a function that it executes as children.

Though, it's not clear how would siblings work then because you'd currently have to do:

```js
div(() => {
  p('test');
  p('test');
});
```

Maybe:

```js
div(p('test'), p('test'));
```

But for performance, wouldn't it be better to have a single function like in the first example? I like the syntax, I'm just not sure yet if the perf implications are worth it.

If we did do this, then we could implement https://github.com/skatejs/skatejs/issues/751 and then just do:

```js
render() {
  return div(p('test'));
}
```

Instead of having to call `vdom.element()`.

bradleyayers:
```
div(p('test'), p('test'));
```

Yes.

> But for performance, wouldn't it be better to have a single function like in the first example?

I'm skeptical it would matter enough to warrant a poorer API. I'd want to see numbers on the perf implications before deciding against it.

treshugart: > I'm skeptical it would matter enough to warrant a poorer API. I'd want to see numbers on the perf implications before deciding against it.

My real concern is memory and negating the benefits of IncrementalDOM. That said, you're right: innocent until proven guilty.

test/unit/vdom/elements.js
```diff
@@ -288,4 +288,61 @@ describe('vdom/elements', () => {
...
+    it('should return an array of functions that create corresponding elements', done => {
+      const [a, b, c] = elements('a', 'b', 'c');
+      fixture(new (define('x-test', {
+        render() {
+          a();
+          b();
+          c();
+        },
+        rendered({ shadowRoot }) {
+          const [elA, elB, elC] = [].slice.call(shadowRoot.children);
```
nadiam84: Why not use spread syntax here?

nadiam84:
lgtm

test/unit/vdom/elements.js
treshugart: We don't include the `Array.from()` polyfill, I don't believe.

bradleyayers:
I would have called this `vdom.builder()`, what's the rationale behind `create` as the name?

treshugart:
@bradleyayers renamed to builder()

src/lifecycle/render.js
```diff
@@ -47,7 +47,15 @@ export default function (Ctor) {
     }
 
     if (shouldRender) {
-      patchInner(sr, render, elem);
+      patchInner(sr, () => {
+        const possibleFn = render(elem);
+        if (typeof possibleFn === 'function') {
+          possibleFn(elem);
+        } else if (Array.isArray(possibleFn)) {
+          possibleFn.forEach(fn => fn());
```
zzarcon: Should we also check if `fn` is a function here?

treshugart:
Might be a good idea to also:

```js
export const h = vdom.builder();
```

...from the main entry so that you can simply import it and start writing JSX.

bradleyayers:
If you're own the fence, then leave out `export const h = vdom.builder();` as it's easy to add later. I think the `vdom.builder()` API is good.

bradleyayers:
lgtm

treshugart:
Cool, going to merge this in then. I created https://github.com/skatejs/skatejs/issues/806 to clean up the rest of the vdom module as this formalises, mostly how to build things.

treshugart:
@bradleyayers I ended up adding `{ h }` as an export because Snabbdom and Yolk also do it. Raised an issue to bring Skate up to par with hyperscript id / class syntax: https://github.com/skatejs/skatejs/issues/807.

PR #795, opened 2016-09-14
#769 - feat: Add ability to get non unique and unique ids for an element.

### LINKED ISSUE #769: Add way to get the non-unique name of a unique component
Currently we generate a unique name for a component name conflicts. If something needs to access the original name (non-unique) of the component, there currently is no way to do that.

treshugart:
Resolves #769 

I'm not attached to this implementation. If there are better ideas, I'm all ears. Currently this adds two properties to the contructor:

- `id`
- `uniqueId` - makes the `name` symbol redundant

This allows us to get both the component name and the unique tag name if there was a conflict. This specifically was designed around helping with theming in AtlasKit, but surely has multiple uses outside of that. I think @nadiam84 had one?

I also proposed a way to do this with the native V1 CE API: https://github.com/w3c/webcomponents/issues/566


nadiam84:
What is the purpose of the `id` here other than make the `name` symbol redundant? Why can't you use 'uniqueId' everywhere and just drop the `id` completely?

treshugart:
Consistency. Could use `uniqueName` as a symbol but then that'd be a breaking change because `name` is already the unique one. What would be a good name for the non-unique-name? Maybe `canonicalName`?

nadiam84:
or `originalName`

src/api/define.js
```diff
@@ -91,44 +94,26 @@ function createInitProps(Ctor) {
...
 }
 
 function registerV0Element(name, Ctor) {
-  let res;
-  let uniqueName;
   try {
     prepareForRegistration(name, Ctor);
-    res = document.registerElement(name, Ctor);
+    return document.registerElement(Ctor.uniqueId, Ctor);
   } catch (e) {
```
jpnelson: Is there some way to make sure that the error was due to being non unique?

treshugart: Yeah, you can test the error message Blink gives you. This should be fairly innocuous as Blink isn't changing the old implementation. Should we do this as part of this PR? This is remnant from older changes.

treshugart:
:( Closing because too many conflicts and things changing in this area. Easier to just re-implement based on this PR.

PR #801, opened 2016-09-16
fix: Avoid symbols, to enable supporting multiple versions of skate.

### LINKED ISSUE #800: Skate 1.4.0 is backwards incompatible with other versions of skate
The new 1.4.0 release includes the use of native Symbols (rather than just strings) for things like `____skate_name`. This *seems* reasonable, but unfortunately introduces incompatibility with other versions of skate when using the vdom APIs.

Consider the following:

```javascript
// library using skate 1.3.2
const ComponentA = define('component-a', …);
```

```javascript
// library using skate 1.4.0
const ComponentB = define('component-b', {
  render() {
    // This line is critical, take note.
    vdom.elementVoid(ComponentA);
  }
});
```

On the critical line we pass the `ComponentA` constructor (which is a skate component) to skate.

Skate needs to be able to detect if what it's been passed is a component, and if it is, extract the tag name to pass onto Incremental DOM.

If skate erroneously determines that it's *not* a component constructor, it calls it like a regular function. Constructors will throw an error if they're not used with `new`, so it's critical that skate correctly detects if it's dealing with a constructor.

The way skate detects if it's been given a constructor, is by [stashing the name of the component on the constructor](https://github.com/skatejs/skatejs/blob/master/src/api/define.js#L106) and then [checking for it later](https://github.com/skatejs/skatejs/blob/master/src/api/vdom.js#L145).

bradleyayers:
Revert "Merge pull request #760 from skatejs/allow-symbols-in-props"

This reverts commit fe41a7622a51bd537fcfebd51bd509e27ab89210, reversing
changes made to e9ee2f090cfdd4f1f5aa5ce83273c1ee8309b4f6.

This resolves #800


bradleyayers:
lgtm

stevemao:
lgtm

PR #803, opened 2016-09-19
#687 - Allow symbols in props

mcorbanini:


test/unit/util/get-all-keys.js
```diff
@@ -0,0 +1,41 @@
+import keys from '../../../src/util/get-all-keys';
+import createSymbol from '../../../src/util/create-symbol';
+
+describe('utils/get-all-keys:', () => {
+  it('should return normal keys from an object', () =>
+    expect(keys({ foo: 1, bar: 2 }))
+      .to.have.lengthOf(2)
```
bradleyayers: Seems a bit redundant given the following assertion.

test/unit/util/get-all-keys.js
```diff
@@ -0,0 +1,41 @@
...
+      .and.to.deep.equal(['foo', 'bar'])
+  );
+
+  it('should return symbol keys from an object', () => {
+    const secret = createSymbol('secret');
+    expect(keys({ foo: 1, [secret]: 2 }))
+      .to.have.lengthOf(2)
```
bradleyayers: Seems a bit redundant given the following assertion.

test/unit/util/get-all-keys.js
```diff
@@ -0,0 +1,41 @@
...
+      .and.to.deep.equal(['foo', secret]);
+  });
+
+  describe('when enumOnly option is set', () => {
```
bradleyayers: Some tests that assert behaviour when using symbols would be good here.

test/unit/util/get-all-keys.js
```diff
@@ -0,0 +1,41 @@
...
+      obj = {};
+      Object.defineProperties(obj, {
+        foo: { enumerable: true, value: 1 },
+        bar: { enumerable: false, value: 2 },
+      });
+    });
+
+    describe('to false (default)', () => {
+      it('should return all keys', () =>
+        expect(keys(obj, { enumOnly: false })).to.have.lengthOf(2)
```
bradleyayers: `.to.have.lengthOf(2)` is a bit redundant here.

test/unit/util/get-all-keys.js
```diff
@@ -0,0 +1,41 @@
...
+          .and.to.deep.equal(['foo', 'bar'])
+      );
+    });
+    describe('to true', () => {
+      it('should return enumerable keys only', () =>
+        expect(keys(obj, { enumOnly: true })).to.have.lengthOf(1)
```
bradleyayers: `.to.have.lengthOf(1)` is a bit redundant here.

bradleyayers:
lgtm

src/lifecycle/props-init.js
```diff
@@ -15,14 +15,19 @@ function getInitialValue(elem, name, opts) {
   return typeof opts.initial === 'function' ? opts.initial(elem, { name }) : opts.initial;
 }
 
+function getNamespacedPropData(elem, name) {
```
mcorbanini: Will go with `getPropData()` if you think so to be consistent with the helper functions defined in this context using the `getSomething` nomenclature.

PR #808, opened 2016-09-22
Try to improve React dependencies handling

zzarcon:
Right now we are including React dependencies (react core and react dom) using script tags. I tried to improve that introducing Bower for it... it might seem not a good idea to introduce a new package manager but I think it bring us some benefits in this case:

* **Semver**: We are currently installing react dependencies with npm but we are not using them at all, this way we can keep the same version target and actually require them which, makes it easier to change in the future.
* **Tests speed**: We don't need to wait for those scripts to load when running the testsuite
* **Being able to run tests offline**: I think this a nice to have thing and it's always a good idea to don't depend on internet connection for the dev environment

###  TODO

* Automatice dependencies installation (bower install) 

I will totally understand if you close this PR since the fact of adding Bower doesn't sounds good this days, but imo I think it will do the job in this case :)

PD: I just realised i didn't commit using `commitizen`, will revert the commit and use the proper command if your are happy with the changes.

Thanks!!


treshugart:
Can we use NPM?

zzarcon:
Couldn't manage to find the react and react-dom production versions of it :(

zzarcon:
Actually this probably the way to go, https://facebook.github.io/react/downloads.html#npm

> By default, React will be in development mode. To use React in production mode, set the environment variable NODE_ENV to production (using envify or webpack's DefinePlugin). A minifier that performs dead-code elimination such as UglifyJS is recommended to completely remove the extra code present in development mode.

treshugart:
Hey @zzarcon just following up on this. Did you want to implement the NPM way with the prod builds?

zzarcon:
Using npm now, take a look :)

treshugart:
Going to merge now and when https://github.com/skatejs/build/pull/403 is in the perf numbers will be using production react (as before, but now with us consuming it via NPM).

PR #817, opened 2016-09-26
Fixes #813 V1 component constructor

### LINKED ISSUE #813: Tests fails on Chrome 55 (Canary)
Hi! I wanted to run the testsuite in a WebComponents V1 environment, so I decided to run them in Chrome 55, currently in canary and turns out that they fail. Apparently the issue appears when running the [Component constructor](https://github.com/skatejs/skatejs/blob/master/src/api/component.js#L76-L80), which is only used when V1 is available.

```javascript
const elem = HTMLElement.call(this, self);
```
```
Uncaught TypeError: Failed to construct 'HTMLElement': Please use the 'new' operator,
this DOM object constructor cannot be called as a function
```

Apparently is not allowed to create an element in such way. So my question is, how are we actually testing that the library works with V1? maybe Im just using a buggy version of Chrome and this is fine :)

If that's not the case, maybe we can use something like `document.createElement` or create a dummy function which has the `HTMLElement` prototype? 

Let me know if this makes sense to you, thanks!!

zzarcon:
This uses `Reflect.constructor` in order to make the inheritance of `HTMLElement` work properly in V1. 

Also we had to move some attr initialisation to the attach callback in order to avoid DOM exceptions, more info here https://www.w3.org/TR/custom-elements/#custom-element-conformance

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -10,6 +10,8 @@ import data from '../util/data';
 import debounce from '../util/debounce';
 import getOwnPropertyDescriptors from '../util/get-own-property-descriptors';
 
+const setElementAsDefined = elem => elem.setAttribute('defined', '');
+
 function callConstructor(elem) {
   const elemData = data(elem);
   const readyCallbacks = elemData.readyCallbacks;
@@ -30,8 +32,6 @@ function callConstructor(elem) {
   if (created) {
     created(elem);
   }
-
-  elem.setAttribute('defined', '');
 
   if (readyCallbacks) {
     readyCallbacks.forEach(cb => cb(elem));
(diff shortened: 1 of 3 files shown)


zzarcon:
@treshugart take a look, test are green now. The only thing I don't like how it looks is that I had to add a timeout to `lifecycle/defined` test in order to make it work on Firefox 😥.

In the it was not necessary to modify this https://github.com/skatejs/skatejs/blob/master/src/lifecycle/props-init.js#L112. Let me know if I'm missing something

treshugart:
I'll have a look in a bit. @bengummer might be interested in reviewing this as well.

treshugart:
We should ensure https://github.com/skatejs/skatejs/pull/817/files#diff-d3ff8469a13e890ec2a5c73b90df75abR29 doesn't set attributes if props are linked.

bengummer:
this PR is continued in #824 (just so we could get the Saucelabs tests to run) so this PR can probably be closed

PR #821, opened 2016-09-28
fix: Fix issue causing false to not remove attributes from virtual el…

### LINKED ISSUE #819: Attribute not removed from normal dom element when passed false
Given the component:

```js
define('x-test', {
  props: {
    test: prop.boolean(),
  },
  render(e) {
    vdom.element('div', { test: e.test });
  },
})
```

The rendered `<div />` will still have the `test` attribute even when it's passed `false`. It should be removed.

treshugart:
…ements.

fixes #819

### DIFF
--- a/test/unit/vdom/properties.js
+++ b/test/unit/vdom/properties.js
@@ -38,13 +38,21 @@ describe('vdom/properties', () => {
 
   it('false should remove the attribute', done => {
     const elem = new (define('x-test', {
-      render() {
-        vdom.element('div', { test: false });
+      props: {
+        test: prop.boolean(),
+      },
+      render(e) {
+        vdom.element('div', { test: e.test });
       },
     }));
     fixture(elem);
+    let div;
     afterMutations(
-      () => expect(elem[symbols.shadowRoot].firstChild.hasAttribute('test')).to.equal(false),
+      () => (div = elem[symbols.shadowRoot].firstChild),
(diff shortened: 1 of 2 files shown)


test/unit/vdom/properties.js
```diff
@@ -38,13 +38,21 @@ describe('vdom/properties', () => {
 
   it('false should remove the attribute', done => {
```
nadiam84: `undefined` and `null` should ... ? ;)

lukebatchelor: Agreed that we should explicitly have an answer to this, but not sure what the expected behaviour is/should be...

treshugart: We never handled those before. The reasoning was to try and make as little bespoke behaviour on top of HTML as possible. Meaning `setAttribute('attr', false)` will set `attr="false"` (stringify). It does the same will `null` and `undefined`. Changing that so that `false` would remove the attribute (which is what was originally done here and this is a bug) made it so that we keep the bespoke footprint small. Technically adding `null` / `undefined` would be a breaking change.

We can do this, but I'd rather do it as a separate issue to make the change deliberate if we think that's a good idea. Thoughts?

treshugart:
Merging now as I've raised #832 for possibly adding `null` / `undefined` in a breaking change.

PR #822, opened 2016-09-28
Fixes #820 updated Component.updated function to check for Symbols

### LINKED ISSUE #820: Symbol props don't cause a re-render when they change
When a props name is hidden with a (native) symbol it doesnt seem to be getting picked up in the Component.updated check. The values are being passed in by skate.props but don't seem to be checked.

Will raise a PR today or tomorrow

lukebatchelor:
I had to skip the precommit hook because it just wasnt letting me commit. 
All the linting passes but it still complains about extraneous packages (even after pruning...)

I figured CI would pick up on any *actual* issues

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -8,6 +8,7 @@ import {
 import { customElementsV0 } from '../util/support';
 import data from '../util/data';
 import debounce from '../util/debounce';
+import getAllKeys from '../util/get-all-keys';
 import getOwnPropertyDescriptors from '../util/get-own-property-descriptors';
 
 function callConstructor(elem) {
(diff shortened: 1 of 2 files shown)


src/api/component.js
```diff
@@ -115,19 +116,23 @@ Component.extend = function extend(definition = {}, Base = this) {
...
-  for (const name in prev) {
-    if (prev[name] !== elem[name]) {
-      return true;
+  let wasUpdated = false;
+  // use get all keys so that we check Symbols as well as regular props
+  // using a for loop so we can break early
+  const allKeys = getAllKeys(prev);
+  for (let i = 0; i < allKeys.length; i += 1) {
+    if (prev[allKeys[i]] !== elem[allKeys[i]]) {
+      wasUpdated = true;
```
zzarcon: Why not `return true` here as it was before?

lukebatchelor: 😵 
I had it in a forEach before and returning true was returning from the forEach function. Completely missed that. Haha

src/api/component.js
```diff
@@ -115,19 +116,21 @@ Component.extend = function extend(definition = {}, Base = this) {
...
 Component.updated = function updated(elem, prev) {
   if (!prev) {
     return true;
   }
-
-  // eslint-disable-next-line no-restricted-syntax
-  for (const name in prev) {
-    if (prev[name] !== elem[name]) {
+  // use get all keys so that we check Symbols as well as regular props
+  // using a for loop so we can break early
```
treshugart: In some perf testing I've done, `for` is also faster in performance critical code than `Array` methods. Especially in older browsers.

src/api/component.js
```diff
@@ -115,19 +116,21 @@ Component.extend = function extend(definition = {}, Base = this) {
...
+  const allKeys = getAllKeys(prev);
+  for (let i = 0; i < allKeys.length; i += 1) {
```
treshugart: Might also be good to save `.length` as a `const` (`allKeysLen`, or something).

src/api/component.js
```diff
@@ -115,19 +116,21 @@ Component.extend = function extend(definition = {}, Base = this) {
 // Skate
 //
 // This is a default implementation that does strict equality copmarison on
-// prevoius props and next props. It synchronously renders on the first prop
+// previous props and next props. It synchronously renders on the first prop
```
bqx: 👌

lukebatchelor: Boyscout 😎

lukebatchelor:
LGTM

jpnelson:
LGTM

treshugart:
lgtm

PR #824, opened 2016-09-29
Fixes #813 V1 component constructor

### LINKED ISSUE #813: Tests fails on Chrome 55 (Canary)

bengummer:
copied commits from https://github.com/skatejs/skatejs/pull/817/commits and then made some changes to get tests passing on IE 11/10/9 and Safari 9

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -232,8 +232,9 @@ Component.prototype = Object.create(HTMLElement.prototype, {
         } else {
           // Sync up the property.
           const propOpts = this.constructor.props[propertyName];
-          // console.log(propOpts);
+          // console.log('propOpts', propOpts);
           propData.settingAttribute = true;
+          // console.log('setting prop to ', (newValue !== null && propOpts.deserialize ? propOpts.deserialize(newValue) : newValue));
           this[propertyName] = newValue !== null && propOpts.deserialize ? propOpts.deserialize(newValue) : newValue;
         }
       }
@@ -248,6 +249,7 @@ Component.prototype = Object.create(HTMLElement.prototype, {
   createdCallback: {
     configurable: true,
     value() {
+      // console.log('createdCallback');
       callConstructor(this);
     },
   },
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -27,7 +27,7 @@ function callConstructor(elem) {
   // Set up a renderer that is debounced for property sets to call directly.
   elem[$rendererDebounced] = debounce(Ctor[$renderer]);
 
-  if (props) {
+  if (props && Ctor[$props]) {
     Ctor[$props](elem);
   }
 
@@ -139,6 +139,9 @@ Component.prototype = Object.create(HTMLElement.prototype, {
   connectedCallback: {
     configurable: true,
     value() {
+      // will eventually need to callsomething like this here:
+      // Ctor[$props](elem);
+      console.log('connectedCallback')
       setElementAsDefined(this);
       callConnected(this);
     },
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -12,8 +12,6 @@ import getAllKeys from '../util/get-all-keys';
 import getOwnPropertyDescriptors from '../util/get-own-property-descriptors';
 import syncPropToAttr from '../util/sync-prop-to-attr';
 
-const setElementAsDefined = elem => elem.setAttribute('defined', '');
-
 function callConstructor(elem) {
   const elemData = data(elem);
   const readyCallbacks = elemData.readyCallbacks;
@@ -59,7 +57,17 @@ function callConstructor(elem) {
   }
 }
 
+function syncPropsToAttrs(elem) {
+  const props = elem.constructor.props;
+  Object.keys(props).forEach((propName) => {
+    const prop = props[propName];
+    // console.log(`Sync ${propName}`);
+    syncPropToAttr(elem, prop, propName, true);
+  });
+}
+
 function callConnected(elem) {
+  syncPropsToAttrs(elem);
   const Ctor = elem.constructor;
   const { attached } = Ctor;
   const render = Ctor[$renderer];
@@ -73,6 +81,8 @@ function callConnected(elem) {
   if (typeof attached === 'function') {
     attached(elem);
   }
+
+  elem.setAttribute('defined', '');
 }
 
 function callDisconnected(elem) {
@@ -147,22 +157,13 @@ Component.updated = function updated(elem, prev) {
   return false;
 };
 
-function syncPropsToAttrs(elem) {
-  const props = elem.constructor.props;
-  Object.keys(props).forEach((propName) => {
-    const prop = props[propName];
-    syncPropToAttr(elem, prop, propName, true);
-  });
-}
-
 Component.prototype = Object.create(HTMLElement.prototype, {
   // v1
   connectedCallback: {
     configurable: true,
     value() {
-      syncPropsToAttrs(this);
+      // console.log('connectedCallback', window.navigator.userAgent);
       callConnected(this);
-      setElementAsDefined(this);
     },
   },
 
@@ -178,14 +179,17 @@ Component.prototype = Object.create(HTMLElement.prototype, {
   attributeChangedCallback: {
     configurable: true,
     value(name, oldValue, newValue) {
+      // console.log(`attrChanged name = '${name}' newValue = ${JSON.stringify(newValue)}`);
       const { attributeChanged, observedAttributes } = this.constructor;
       const propertyName = data(this, 'attributeLinks')[name];
 
       // In V0 we have to ensure the attribute is being observed.
       if (customElementsV0 && observedAttributes.indexOf(name) === -1) {
+        // console.log('exiting !v0');
         return;
       }
 
+      // console.log('propertyName', propertyName);
       if (propertyName) {
         const propData = data(this, `api/property/${propertyName}`);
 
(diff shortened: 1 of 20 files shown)


src/api/component.js
```diff
@@ -74,10 +74,11 @@ function callDisconnected(elem) {
 }
 
 // v1
-function Component(self) {
-  const elem = HTMLElement.call(this, self);
-  callConstructor(elem);
-  return elem;
+function Component(...args) {
+  const hasReflect = window.hasOwnProperty('Reflect'); // eslint-disable-line no-prototype-builtins
+  const elm = (hasReflect && Reflect.construct(HTMLElement, args, this.constructor)) || HTMLElement.call(this, args[0]);
```
bengummer: doing `Reflect && Reflect.construct(...)` here didn't work on IE

bengummer:
PR based on the awesome work by @zzarcon in https://github.com/skatejs/skatejs/pull/817/files#diff-d3ff8469a13e890ec2a5c73b90df75abR29

Needed to create this branch in the skatejs/skatejs repo so that Saucelabs browser tests would run

src/api/component.js
```diff
@@ -74,10 +74,11 @@ function callDisconnected(elem) {
 }
 
 // v1
-function Component(self) {
-  const elem = HTMLElement.call(this, self);
-  callConstructor(elem);
-  return elem;
+function Component(...args) {
+  const hasReflect = window.hasOwnProperty('Reflect'); // eslint-disable-line no-prototype-builtins
```
treshugart: This should probably live at the top level as we probably don't need a lazy check for it. You can also do `'Reflect' in window`.

src/api/component.js
```diff
@@ -189,6 +191,7 @@ Component.prototype = Object.create(HTMLElement.prototype, {
   createdCallback: {
     configurable: true,
     value() {
+      setElementAsDefined(this);
```
treshugart: I knew this was moved to `connectedCallback()`, but why do we need this here? Shouldn't it be causing tests in Canary to fail?

bengummer: you're right, just didn't get picked up in CI - fixing now

src/api/component.js
```diff
@@ -11,6 +11,9 @@ import debounce from '../util/debounce';
 import getAllKeys from '../util/get-all-keys';
 import getOwnPropertyDescriptors from '../util/get-own-property-descriptors';
 
+const setElementAsDefined = elem => elem.setAttribute('defined', '');
+const hasReflect = 'Reflect' in window;
```
zzarcon: Should we move this to [support](https://github.com/skatejs/skatejs/blob/master/src/util/support.js) then?

src/api/component.js
zzarcon: I thought this needs to be called here for V0 support, otherwise it will never be called since it has been extracted from `callConstructor`? Maybe Im wrong, but tests works for me on Chrome.

zzarcon:
@bengummer thanks a lot for taking over my initial work 🎉 🚀 !

I think I had something wrong configured locally and the tests were not working fine on CI, also didn't know that this was breaking Safari and IE, I thought there the constructor was not called since V1 and maybe V0 was not available at all hehe.

Anyways this looks good to me, I pretty much just did what @treshugart told me, so the job is mostly from him!

src/api/component.js
bengummer: good idea

treshugart:
@zzarcon don't be silly. You jumped in the deep end and swam, even if I was standing lifeguard :). Nice work both of you!

treshugart:
I think the only last thing on here is to ensure we're releasing a breaking change with this (unless there's evidence to suggest against it). The reason I say this is that moving attribute linking to when the element is connected might break expectations for some, even if we're still ensuring props are initialised.

bengummer:
fyi @treshugart and @zzarcon: i've tried to defer the v1 attribute syncing until the `connectedCallback` happens and got most of the way, but for some reason the `connectedCallback` wasn't being triggered when I ran the tests in Chrome 55 so i couldn't get the tests to pass.

i've just added a commit with my work in progress trying to defer the attr sync. I'm away for the next 5 or so days so feel free to pick up from here otherwise i will continue when i'm back.

thanks for your help @zzarcon ! and for the guidance @treshugart :)

zzarcon:
@treshugart I don't have that much experience with the lib but I also think this is a breaking change and will be better to do bump the version, even if maybe no users will be affected right now 😿

src/api/component.js
```diff
@@ -74,10 +75,10 @@ function callDisconnected(elem) {
 }
 
 // v1
-function Component(self) {
-  const elem = HTMLElement.call(this, self);
-  callConstructor(elem);
-  return elem;
+function Component(...args) {
+  const elm = (hasReflect && Reflect.construct(HTMLElement, args, this.constructor)) || HTMLElement.call(this, args[0]);
```
bradleyayers: Why not use a proper ternary here?

bradleyayers: is `elm` a new convention? I thought `elem` was the standard in skatejs.

src/api/component.js
```diff
@@ -138,6 +139,10 @@ Component.prototype = Object.create(HTMLElement.prototype, {
   connectedCallback: {
     configurable: true,
     value() {
+      // will eventually need to callsomething like this here:
+      // Ctor[$props](elem);
+      console.log('connectedCallback')
```
bradleyayers: Stray console.log?

test/unit/api/component.js
```diff
@@ -52,7 +52,7 @@ describe('api/Component', () => {
   });
 
   describe('updated function', () => {
-    it('should return true if a prop changes', () => {
+    it.only('should return true if a prop changes', () => {
```
bradleyayers: Stray `.only`.

test/unit/lifecycle/defined.js
```diff
@@ -5,9 +5,16 @@ describe('lifecycle/defined', () => {
...
+  it('should add the [defined] attribute when the element is upgraded', (done) => {
     const Elem = define('x-test', {});
     const elem = new Elem();
-    expect(elem.hasAttribute('defined')).to.equal(true);
+
+    document.body.appendChild(elem);
+
+    setTimeout(() => {
+      expect(elem.hasAttribute('defined')).to.equal(true);
+      document.body.removeChild(elem);
```
mcorbanini: This should be covered in `afterEach` right?

test/unit/lifecycle/defined.js
```diff
@@ -5,9 +5,16 @@ describe('lifecycle/defined', () => {
...
 
-  it('should add the [defined] attribute when the element is upgraded', () => {
+  it('should add the [defined] attribute when the element is upgraded', (done) => {
     const Elem = define('x-test', {});
     const elem = new Elem();
-    expect(elem.hasAttribute('defined')).to.equal(true);
+
+    document.body.appendChild(elem);
+
+    setTimeout(() => {
```
mcorbanini: Why not using `afterMutations` ?

zzarcon: Oh good point, didn't know about that utility

src/api/component.js
treshugart: If it works in Chrome it should be fine. We'll do another breaking change soon to remove v0 support, so as long as the tests pass for it I'm okay with whatever works for now as long as we can get v1 fixed up.

src/api/component.js
treshugart: +1 for `elem`

treshugart:
Ok, I just made a bunch of updates. Things that still need to happen:

1. On `connected`, sync attributes (can probably just re-set props to their same values).
2. Write test for 1 (probably should do this first, but I listed 1 first for context).
3. Update prop tests so that they test the attribute values after connected and after mutations.

test/unit/lifecycle/defined.js
treshugart: I don't think there's an after each in those tests.

test/unit/lifecycle/defined.js
treshugart: done

test/unit/api/component.js
treshugart: done

src/api/component.js
treshugart: removed

src/api/component.js
treshugart: changed to `elem`

treshugart: updated to use a ternary

src/api/component.js
treshugart: done

bengummer:
This is now ready for PR again, we've got the attributes being synced from props on elem connect. Also have tests for this (see new `delayed-attributes.js` tests). There was actually quite a bit of refactoring needed for the existing tests, a lot of `afterMutations` as well as some issues on IE10/IE11/FF.

treshugart:
Totes approving this! I think this introduces some ambiguity about whether an attribute or property *should* take precedence due to the delay of syncing but I don't know there's much we can do about that until we come up with an answer for #838.

PR #831, opened 2016-09-30
fix(vdom): Fix issue causing multiple children (and as attrs) to not …

### LINKED ISSUE #830: Need to handle arrays for when specifying vdom children
We need to support something like:

```js
h('div', 
  h('span'),
  [h('span'), h('span')],
  [h('span'), h('span'), h('span')]
);
```

So this can work in JSX:

```js
<div>
  <span />
  {[0, 1].map(() => <span />)}
  {[0, 1, 2].map(() => <span />)}
</div>
```

treshugart:
…be rendered.

fixes #830

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -321,7 +321,7 @@ export function element(tname, attrs, ...chren) {
       ch();
     } else if (ctype === 'string' || ctype === 'number') {
       newText(ch);
-    } else if (ctype === 'object') {
+    } else if (Array.isArray(ch)) {
       ch.forEach(sch => sch());
     }
   });
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -293,7 +293,7 @@ export function element(tname, attrs, ...chren) {
 
   // If attributes are a function, then they should be treated as children.
   if (atype === 'function' || atype === 'string' || atype === 'number') {
-    chren = [attrs];
+    chren.unshift(attrs);
   }
 
   // Ensure the attributes are an object. Null is considered an object so we
@@ -321,6 +321,8 @@ export function element(tname, attrs, ...chren) {
       ch();
     } else if (ctype === 'string' || ctype === 'number') {
       newText(ch);
+    } else if (ctype === 'object') {
+      ch.forEach(sch => sch());
     }
   });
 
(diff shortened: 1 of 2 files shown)


src/api/vdom.js
```diff
@@ -321,6 +321,8 @@ export function element(tname, attrs, ...chren) {
       ch();
     } else if (ctype === 'string' || ctype === 'number') {
       newText(ch);
+    } else if (ctype === 'object') {
```
zzarcon: Is this safe enough to check if the type it's an Array? what about `Array.isArray`?

test/unit/vdom/builder.js
```diff
@@ -79,5 +79,32 @@ describe('create()', () => {
...
+        rendered({ shadowRoot }) {
+          expect(shadowRoot.innerHTML).to.equal(
+            '<div>' +
+              '<span id="1"></span>' +
+              '<span id="2"></span>' +
+              '<span id="3"></span>' +
+              '<span id="4"></span>' +
+              '<span id="5"></span>' +
+              '<span id="6"></span>' +
+            '</div>'
```
zzarcon: Maybe is better to use a template literal for this?

```javascript
`<div>
  <span id="1"></span>
  <span id="2"></span>
  <span id="3"></span>
  <span id="4"></span>
  <span id="5"></span>
  <span id="6"></span>
</div>`
```

treshugart: Then the whitespace in the literal would fail the test (vdom doesn't generate whitespace).

Hotell:
Hey folks! any update on this?

it is really inconvenient to call `jsx` transforms manually 

```jsx
const PleaseFixMe = () => {
  return [1,2,3].map( itm => <b>{itm}</b>() )
}
```

src/api/vdom.js
Hotell: `Array.isArray` is the right way IMHO, `'object'` can be `null`,`Object`, `Array`

treshugart:
@Hotell just restarting the builds now. I need this too :p

src/api/vdom.js
treshugart: Yeah, not sure why I used `'object'`. Possibly because I was testing object literal traversal at some point? Will change.

Issue #829, opened 2016-09-30
skate does not differentiate between multiple Symbol props with the same description


bqx:
e.g.

```
const { define, vdom } = skate;

const prop1 = Symbol();
const prop2 = Symbol();

define('x-el', {
  props: {
    [prop1]: {},
    [prop2]: {}
  }
});
```

See this jsbin for an example: http://jsbin.com/xequbegedu/edit?js,console,output

The workaround is to create Symbols with different descriptors, e.g. the above example would work with:

```
const prop1 = Symbol('prop1');
const prop2 = Symbol('prop2');
```

PR #834, opened 2016-10-04
Fix multiple symbol props

bqx:
Use the prop names as keys instead of prefixing with a string, so that skate can differentiate between multiple symbol props. Previously, using two Symbols with the same descriptor would generate something like:

```
{ 
  'api/property/Symbol()': { ... }
}
```

since we were just converting the Symbol to a string.

Also, put the props inside an object with key `props` to prevent clashing with any other skate data.

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -163,7 +163,7 @@ Component.prototype = Object.create(HTMLElement.prototype, {
       }
 
       if (propertyName) {
-        const propData = data(this, `api/property/${propertyName}`);
+        const propData = data(this, 'props')[propertyName];
 
         // This ensures a property set doesn't cause the attribute changed
         // handler to run again once we set this flag. This only ever has a
--- a/test/unit/api/props-with-symbol.js
+++ b/test/unit/api/props-with-symbol.js
@@ -1,22 +1,27 @@
 import afterMutations from '../../lib/after-mutations';
+import hasSymbol from '../../lib/after-mutations';
+import createSymbol from '../../../src/util/create-symbol';
 import element from '../../lib/element';
 import fixture from '../../lib/fixture';
 import { props } from '../../../src/index';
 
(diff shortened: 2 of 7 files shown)


src/api/component.js
```diff
@@ -163,7 +163,7 @@ Component.prototype = Object.create(HTMLElement.prototype, {
       }
 
       if (propertyName) {
-        const propData = data(this, `api/property/${propertyName}`);
```
bqx: Yep, previously we stored the skate data like this:

```
{
  'api/property/myProperty': { ... },
  'api/property/myOtherProperty': { ... },
  'attributeLinks': { ... },
  'propertyLinks': { ... },
}
```

Now we're storing it like this:

```
{
  'props': {
    'myProperty': { ... },
    'myOtherProperty': { ... },
  },
  'attributeLinks': { ... },
  'propertyLinks': { ... },
}
```

test/unit/api/props-with-symbol.js
```diff
@@ -0,0 +1,73 @@
...
+  const secret2 = createSymbol('secret');
+
+  beforeEach(done => {
+    elem = new (element().skate({
+      props: {
+        [secret]: {
+          initial: 'secretKey',
+        },
+        [secret2]: {
+          initial: 'secretKey2',
```
mcorbanini: Do we have a test that actually check [this issue](https://github.com/skatejs/skatejs/issues/829) being solved?
I mean, adding 2+ props with symbol keys without description and trying to get their values?

bqx: We test here for two symbols with the same descriptor - this should be sufficient to test that we use the actual Symbols as keys instead of stringifying them (in which case they would be indistinguishable)

treshugart:
Restarting build. If it's green, I'll merge so it can be released.

PR #843, opened 2016-10-10
WIP for spiking the decoupling of Incremental DOM.

treshugart:
We seem to be having to wrap a lot of Incremental DOM's functionality for:
- intuitive `skip` behaviour
- simpler builder API
- functions as tag names
- `ref`

It doesn't mean Incremental DOM isn't a good solution, but we've wrapped a lot around it, some of it possibly negating (some of) the benefits around it's great memory overhead. This is just an exploration of what a virtual DOM abstraction layer might look like at the Skate level and how vdom adpaters can be written and included as separate packages which would allow things like Preact, React, Hyperscript, Virtual DOM, SkateJS-DOM-Diff, MorphoDOM, Incremental DOM, etc, to be used as rendering layers within Skate while Skate provides a single vdom API that integrates well with JSX and a clean hyperscript style builder API. Meaning rendering layers (adapters) can be switched out without the consumer needing to change their `render()` function.

I was able to break out Incremental DOM while maintaining compatibility with the current API. As such, iDOM is still included as the default renderer in the core, but it's been separated out from the virtual DOM layer and into it's own renderer which allows other renderers to be easily specified, and if done with tree shaking, should tree shake out iDOM.


treshugart:
#848 does a lot of the work at exposing `renderer()` all this will need to do is remove iDOM specific stuff in `vdom` and implement the adapter pattern for other vdom libs.

treshugart:
This has diverged from #872 which does a lot of work to enable this in a different way, so closing.

Issue #845, opened 2016-10-10
Multiple instances of incremental dom break component extension


treshugart:
If you have a component that renders and it's extended by another component, you get Incremental DOM errors:

```js
// Contains it's own SkateJS / Incremental DOM
import FirstComponent from 'first-component';

// Imports its own version of Incremental DOM
import { define } from 'skatejs';

define('second-component', class extends FirstComponent {});
```

treshugart:
A way to get around this would be to define a `renderer` option on the constructor that is used in the `render` lifecycle. When extending, the extended gets both `render()` and `renderer()` from the parent, thus getting the parent's version if iDOM.

PR #848, opened 2016-10-11
#845 - Fix multiple Incremental DOM instance issues when extending from another component that has a separate instance

treshugart:
Add a renderer() callback to customise the renderer.

This allows a child component to inherit the render() and renderer() from a parent component so that multiple versions of Incremental DOM aren't brought in (only the one from the parent). While this isn't bullet proof, it's a good boyscout refactor to contain things more closely to where they're used in the base Componnet class and covers the major use case of inheriting from a parent where you're not redefining the renderer or render methods. The added benefit is that renderer() can now be overridden, which is desirable as we move towards a more decoupled rendering pipeline (even if we continue to support Incremental DOM).


src/api/props.js
```diff
@@ -13,9 +13,7 @@ function get(elem) {
 
 function set(elem, newProps) {
   assign(elem, newProps);
-  if (elem.constructor.render) {
-    elem.constructor[$renderer](elem);
-  }
+  elem.constructor[$renderer](elem);
```
bradleyayers: In my upcoming PR I actually changed this to:

```js
function set(elem, newProps) {
  assign(elem, newProps);

  // `set()` may be called on a non-skate or non-initialised component, which
  // means we can't be sure that `render()` exists.
  const render = elem.constructor[$renderer];
  if (render) {
    render(elem);
  }
}
```

We should probably add a test to lock in the behaviour that `props` can be called on a non-initialised or non-skate element.

treshugart: Yeah, I'll change it back to what it was.

src/api/component.js
```diff
@@ -26,16 +51,16 @@ function callConstructor(elem) {
   elem[$created] = true;
 
   // Set up a renderer that is debounced for property sets to call directly.
-  elem[$rendererDebounced] = debounce(Ctor[$renderer]);
+  elem[$rendererDebounced] = debounce(constructor[$renderer].bind(constructor));
 
   // Set up property lifecycle.
-  if (props && Ctor[$props]) {
-    Ctor[$props](elem);
+  if (constructor.props && constructor[$props]) {
```
bradleyayers: I realise this is a boyscout, but why are we even checking `constructor.props`?

bradleyayers:
Missing tests.

src/api/component.js
treshugart: It was checking `props` before (which came from `constructor`.

src/api/component.js
```diff
@@ -1,22 +1,47 @@
...
 import getOwnPropertyDescriptors from '../util/get-own-property-descriptors';
+import getSetProps from './props';
 import syncPropToAttr from '../util/sync-prop-to-attr';
 
+// Abstracts shadow root across v1, v0 and no support.
+// Once v1 is supported everywhere, we can call elem.attachShadow() directly
+// and remove this function.
+function attachShadow(elem) {
+  if (shadowDomV1) {
+    return elem.attachShadow({ mode: 'open' });
```
nadiam84: Why always `open`?

src/api/component.js
```diff
@@ -66,32 +91,27 @@ function syncPropsToAttrs(elem) {
...
 function callConnected(elem) {
-  const Ctor = elem.constructor;
-  const { attached } = Ctor;
-  const render = Ctor[$renderer];
+  const { constructor } = elem;
 
   syncPropsToAttrs(elem);
 
   elem[$connected] = true;
+  constructor[$renderer](elem);
```
nadiam84: You don't need to check `type of === function` any more?

src/api/component.js
treshugart: It was always open, this code was just brought into it's own function.

src/api/component.js
treshugart: It's not public API and it's pre-defined on the class itself.

PR #849, opened 2016-10-11
fix: assign helper function fix for symbols

mcorbanini:


src/util/assign.js
```diff
@@ -1,5 +1,8 @@
-const assign = Object.assign;
-export default assign ? assign.bind(Object) : (obj, ...args) => {
-  args.forEach(arg => Object.keys(arg).forEach(name => obj[name] = arg[name])); // eslint-disable-line no-return-assign
+import keys from './get-all-keys';
+
+// We are not using Object.assign if it is defined since it will cause problems when Symbol is polyfilled.
+// Apparently Object.assign (or any polyfill for this method) does not copy non-native Symbols.
```
treshugart: I think we should raise an issue with the polyfill for this, even if we do this for now. Polyfills should behave consistently with native where possible and I can't think of anything that would prevent this.

PR #851, opened 2016-10-11
Use microtasks for debounce scheduling if available, resolves #846.

### LINKED ISSUE #846: FOUC when nesting components
When rendering a skate component *within* another skate component, the nested component is vulnerable to a FOUC.

This can occur when the nested component has a prop that directly affects the way the component is rendered. Consider the following:

```js
define('x-parent', {
  render(elem) {
    return h('x-child', { active: true });
  }
});

define('x-child', {
  render(elem) {
    return h('p', { class: elem.active ? 'active' : '' });
  },
  
  props: {
    active: {}
  }
});
```

The problem occurs when rendering `<x-parent />`.

Sequence of events:

1. `<x-parent />` is added to the DOM, and its `connectedCallback` synchronously causes `render()` to be invoked.
2. `<x-child />` is synchronously added to the DOM via incremental-dom, (but importantly *without* an `active` attribute).
3. `x-child`'s `render()` function is synchronously invoked (again via `connectedCallback`) causing `<p class=""></p>` (at this time `active` is `false`) to be added to the DOM.
4. Incremental DOM now applies `active=true` attribute to `x-child`, which uses skate's overridden behaviour to actually apply the attribute as property (`node.active = true`).
5. The `active` property setter is invoked, [which schedules a `render()` *on the next tick*](https://github.com/skatejs/skatejs/blob/master/src/lifecycle/props-init.js#L73).
6. *One tick passes.*
7. `x-child`'s `render()` function is invoked, and adds `active` to the `<p>` element.

Here's a demo: http://www.webpackbin.com/N1Rm2yEAZ

There's a simple way to solve this — don't debounce `render()` when a property changes. However this would introduce a performance problem when setting multiple properties, as it would cause multiple renders:

```js
elem.foo = true;  // triggers render()
elem.bar = true;  // triggers render()
…
```

The whole point of the debouncing is ensure only *one* invocation of `render()` occurs regardless of the number of properties that are set.

bradleyayers:
Copying across @zzarcon's [concern](https://github.com/skatejs/skatejs/pull/850#issuecomment-252840327):

> This approach seems quite good to me, the only thing Im wondering its if we should disconnect the `MutationObserver` observer when the component gets disconnected? 

> Not sure if right now we will leak something, or maybe we just don't need to care about it.

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -68,7 +68,7 @@ function syncPropsToAttrs(elem) {
 function callConnected(elem) {
   const Ctor = elem.constructor;
   const { attached } = Ctor;
-  const render = Ctor[$renderer];
+  const render = elem[$rendererDebounced];
 
   syncPropsToAttrs(elem);
 
--- a/src/api/props.js
+++ b/src/api/props.js
@@ -13,7 +13,7 @@ function get(elem) {
 
 function set(elem, newProps) {
   assign(elem, newProps);
-  if (elem.constructor.render) {
+  if (elem.constructor[$renderer]) {
     elem.constructor[$renderer](elem);
   }
 }
--- a/src/util/debounce.js
+++ b/src/util/debounce.js
@@ -7,15 +7,40 @@ for (let i = 0; i < longerTimeoutBrowsers.length; i += 1) {
   }
 }
 
-export default function (fn) {
+function microTaskDebounce(cbFunc) {
+  let called = false;
+  let i = 0;
+  let cbArgs = [];
+  const elem = document.createElement('span');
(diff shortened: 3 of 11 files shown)


bradleyayers:
Will MutationObserver leak here? I don't remember if JS garbage collection is tracing or reference counting based? Either way I'm probably happy to tackle it in a separate issue, there's a high degree of urgency for this specific issue.

zzarcon:
@bradleyayers fine for me to do it in another pr if necessary, I'd like to run some benchmarks on it to see if leaks memory, will share the results afterwards.

PR #858, opened 2016-10-12
Allow components to extend other Elements than HTMLElement

### LINKED ISSUE #804: Ability to extend elements other than HTMLElement
Right now, `Component` extends directly from `HTMLElement`. We should keep this as the default, but also allow the consumer to specify a class that `Component` should extend.

zzarcon:
As @treshugart suggested, now we can extend other components than `HTMLElement`: 

Extending `HTMLInputElement` 
```javascript
define('my-element', class extends component(HTMLInputElement) {});
````

Default extend to `HTMLElement`
```javascript
define('my-element', class extends component() {});
```

We will need to add some docs in order to explain users how to use this, thoughts? @bradleyayers @treshugart 

Closes #804

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -320,4 +320,8 @@ Component.prototype = Object.create(HTMLElement.prototype, {
   }
 });
 
+export const component = (superClass = HTMLElement) => {
+  return Component.extend({}, superClass);
+};
+
 export default Component;
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@ import * as prop from './api/prop';
 import * as symbols from './api/symbols';
 import * as vdom from './api/vdom';
 import Component from './api/component';
+import { component } from './api/component';
 import define from './api/define';
 import emit from './api/emit';
 import link from './api/link';
@@ -12,6 +13,7 @@ const h = vdom.builder();
 
 export {
   Component,
+  component,
   define,
   emit,
   h,
--- a/test/unit/registration.js
+++ b/test/unit/registration.js
@@ -1,7 +1,8 @@
 /* eslint-env jasmine, mocha */
 
-import { define } from '../../src/index';
+import { define } from '../../src';
 import helperElement from '../lib/element';
+import { component } from '../../src';
 
 const { HTMLElement } = window;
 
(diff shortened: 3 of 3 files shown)


treshugart:
@zzarcon Maybe updating https://github.com/skatejs/skatejs#customised-built-in-elements to include how to use this. I think this is an edge-case mostly, so I think keeping `Component` as the main extension point would be useful. If that's the case, maybe it'd be better to have `Component.create()` instead of just `component()` so that it's less confusing when reading. What do you think?

src/api/component.js
```diff
@@ -320,4 +320,8 @@ Component.prototype = Object.create(HTMLElement.prototype, {
   }
 });
 
+export const component = (superClass = HTMLElement) => {
+  return Component.extend({}, superClass);
```
treshugart: I'm not sure this would actually work, would it? It would extend the intended built-in, but because you've overwritten `this`, pointing to `Component` in the `extend()` function, it won't get Skate behaviour, would it?

zzarcon:
@treshugart I personally like more the `Component.create()` instead of `component` since it can be confusing for some people. But since it has conflicts with the inheritance we should stick with `component` :\

About if the component will still work, I still have to add some tests for ensure that works as expected and also to ensure the default inheritance, just pushed this for now but I was not 100% sure if this was what we were expecting :)

treshugart:
@zzarcon I had some chats with some people on Polymer - and it was also mentioned in the conference - but they're not supporting customised built-ins in Polymer 2, and they don't seem to feel like customised built-ins should be encouraged if Apple isn't going to reach consensus on it. I tend to agree, even though built-ins would be very useful. It also simplifies our API. What do you think?

zzarcon:
Im not that aware about Polymer 2 and Apple direction on this, but I do think it can be useful for users of skate. Which is also true is that makes out Api complicated 😔 

That said, we can think about how important is this feature for us, maybe right now is not something critical and we can wait until other users request the feature. So Im fine closing this for now and maybe opening in it in some months? what do you think.

treshugart:
Yeah, I think we should close it since the path forward is a little muddy. I've also spoken to some people about custom attributes. There's a some different opinions on that, but it seems like something that some are thinking about proposing as a standard. That could easily replace something like the `is` attribute and built-ins.

PR #862, opened 2016-10-16
fix: Remove redundant code to reduce bloat.

### LINKED ISSUE #861: Remove redundant debounce code that uses setTimeout()
We require a mutation observer polyfill be on the page if you're using skate in a browser that doesn't have custom elements natively. All current polyfills require this, so we shouldn't have to work around it.

treshugart:
fixes #861

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     "react": "15.x",
     "react-dom": "15.x",
     "skatejs-build": "9.x",
-    "skatejs-web-components": "1.x"
+    "skatejs-web-components": "1.x",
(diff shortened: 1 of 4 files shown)


treshugart:
Looks like we'll need to place a hard dependency on `MutationObserver` being present. We could rollback and only detect `Trident` but #863 will need mutation observers anyway if it gets accepted.

treshugart:
Going to close this as it seems we need to add the element to the document for the mutation observer polyfill to pick it up. Will do another PR to only check `Trident` for the `setTimeout()` version.

PR #866, opened 2016-10-16
#863 - Add slot() prop type for reflecting slotted nodes.

### LINKED ISSUE #863: Add slot() prop type
### Problem

As a developer, I want a prop to update whenever a slot is assigned nodes so that my component is re-rendered. This prop should contain all of the assigned nodes as an array for the corresponding slot.

Slots have a `slotchange` event that fires when elements are slotted, but it doesn't seem to fire on initial slotting, for example if you render your shadow root with slots _after_ the host element contains children.
### Solution

We have a few options:
1. add a ref with a `setTimeout()` - because you cant trigger a re-render in `render()` - to init the slotted nodes
2. create a custom function helper for a slot that does the above
3. create a custom prop that uses a `MutationObserver` to listen for changes on the host and updates the prop when it changes

Options 1 and 2 are not ideal because they use `setTimeout()` and all the baggage that comes with it. Option 3 isn't perfect because it doesn't use `slotchange` at all. It completely subverts it and uses a `MutationObserver` instead (which _does_ have the same timing).

Option 3 could be altered to not use mutation observers, but then you'd have to declare a `slot()` prop along with a `slotchange` handler. This means declaring code in two separate spots. I did try doing this, but it's more to manage and worry about over just declaring a slot prop and being done with it.
### Spike

The slot prop accepts a `slot` option that corresponds to the `name` of the `<slot />` element the prop should represent. If not given, it defaults to `null` which means it should be used for the default slot.
### Potential issues

Performance with several mutation observers. However, they don't pose a performance issue if you are not modifying anything that they're not supposed to be listening to. The logic within them should also have little overhead. The only consideration here is filtering the array of `children`, but that's quick unless you're doing hundreds of thousands of them.
### Todo if accepted

If this solution is accepted, we should notify the W3C of our workaround and why `slotchange` didn't work for us.

treshugart:
Fixes #863.


src/util/create-symbol.js
```diff
@@ -1,3 +1,4 @@
+let counter = 0;
 export default function createSymbol (description) {
-  return typeof Symbol === 'function' ? Symbol(description) : description;
+  return typeof Symbol === 'function' ? Symbol(description) : `____skate_symbol_${description || (++counter)}`;
```
treshugart: Boyscout. Wanted to not have to specify a descriptor like native symbol.

src/util/symbols.js
```diff
@@ -1,10 +1,11 @@
-export const connected = '____skate_connected';
-export const created = '____skate_created';
+import createSymbol from './create-symbol';
+export const connected = createSymbol();
+export const created = createSymbol();
 export const name = '____skate_name';
```
treshugart: Left this the same as there's a test that references this.

src/util/symbols.js
```diff
@@ -1,10 +1,11 @@
...
-export const props = '____skate_props';
-export const ref = '____skate_ref';
-export const renderer = '____skate_renderer';
-export const rendering = '____skate_rendering';
-export const rendererDebounced = '____skate_rendererDebounced';
-export const shadowRoot = '____skate_shadowRoot';
-export const updated = '____skate_updated';
+export const props = createSymbol();
```
treshugart: Boyscouts because of the change to `createSymbol()`.

src/util/symbols.js
treshugart: Left this the same as there's a test that references this.

test/boot.js
```diff
@@ -2,6 +2,10 @@
 
 import helperFixture from './lib/fixture';
 
+if (!window.MutationObserver) {
+  require('webcomponents.js/MutationObserver');
```
treshugart: Required to make the new prop type tests pass in IE. It will only error on the user if they try and use `slot()` without `MutationObserver`. This isn't a breaking change, but maybe we want to log if `MutationObserver` isn't available? Might be best to do this at some global level, or throw an error in the spot that requires it?

test/unit/api/properties.js
```diff
@@ -162,6 +163,55 @@ describe('api/prop', () => {
...
+        () => expect(elem.test2).to.deep.equal([]),
+        () => elem.appendChild(div1),
+        () => elem.appendChild(div2),
+        () => expect(elem.test1).to.deep.equal([div1]),
+        () => expect(elem.test2).to.deep.equal([div2]),
+        done
+      );
+    });
+  });
+
```
nadiam84:
```
    let setCallOne = sinon.spy();
    let setCallTwo = sinon.spy();
    beforeEach(() => {
      elem = new (define('x-test', {
        props: {
          test1: prop.slot({
            set: setCallOne,
          }),
          test2: prop.slot({ name: 'test',
            set: setCallTwo,
          })
        },
      }))();
      div1 = document.createElement('div');
      div2 = document.createElement('div');
      div2.setAttribute('slot', 'test');
    });
```

```
it('set should work', done => {
      fixture(elem);
      afterMutations(
        () => {
          expect(setCallOne).not.to.be.called;
          expect(setCallTwo).not.to.be.called;
        },
        () => elem.appendChild(div1),
        () => elem.appendChild(div2),
        () => {
          expect(setCallOne).to.be.called;
          expect(setCallTwo).to.be.called;
        },
        done
      );
    });
```

This fails, setters on the props doesn't work.

treshugart: That passes for me. I copied and updated indentation. This looks right, correct?

```js
describe('nadia breaking my tests again!', () => {
  let setCallOne = sinon.spy();
  let setCallTwo = sinon.spy();
  beforeEach(() => {
    elem = new (define('x-test', {
      props: {
        test1: prop.slot({
          set: setCallOne,
        }),
        test2: prop.slot({ name: 'test',
          set: setCallTwo,
        })
      },
    }))();
    div1 = document.createElement('div');
    div2 = document.createElement('div');
    div2.setAttribute('slot', 'test');
  });
  it('set should work', done => {
    fixture(elem);
    afterMutations(
      () => {
        expect(setCallOne).not.to.be.called;
        expect(setCallTwo).not.to.be.called;
      },
      () => elem.appendChild(div1),
      () => elem.appendChild(div2),
      () => {
        expect(setCallOne).to.be.called;
        expect(setCallTwo).to.be.called;
      },
      done
    );
  });
});
```

treshugart: Did you mean to assert that the setters *should* be called?

src/util/symbols.js
bradleyayers: I think this will break cross-version compatibility, because our idom integration looks at the `elem[$props]` to determine if it should set as attribute as a property doesn't it?

README.md
```diff
@@ -1010,6 +1010,17 @@ skate.prop.boolean({
 
 Generally built-in properties will only return a definition containing `coerce`, `deserialize` and `serialize` options. They may also define a `deafult`, such as with the `boolean` property.
 
+Be aware that if you pass options into a property that you might be overriding default behaviour. With this in mind, it is up to you to make sure you callback or properly set anything that you override:
+
+```js
+const slot = skate.prop.slot();
+const prop = Object.assign(slot, {
```
bradleyayers: It's not clear to me if this means that this would be a problem:

```js
const slot = skate.prop.slot({ set: () => {} });
```

treshugart: > Be aware that if you pass options into a property that you might be overriding default behaviour. With this in mind, it is up to you to make sure you callback or properly set anything that you override:

treshugart: And it's just JS. I don't think it's worthwhile to document the implementation details for each built-in. In practice, we've found overriding to be an edge case and I think is something that we'd likely deprecate in the future.

src/api/prop.js
```diff
@@ -11,27 +12,99 @@ export function create (def) {
...
   deserialize: alwaysUndefinedIfNotANumberOrNumber,
   serialize: alwaysUndefinedIfNotANumberOrNumber
 });
 
+
+
+// slot
+
+const { MutationObserver } = window;
+const attrSlot = 'slot';
```
bradleyayers: I would have inlined this, to avoid any confusion that this variable is used anywhere else.

treshugart: Lol, I anticipated someone pointing out a string literal and saying to use a `const` in trying to avoid less PR churn :DDD

src/util/symbols.js
treshugart: We do this instead: https://github.com/skatejs/skatejs/blob/master/src/api/vdom.js#L130. Only [`name` and `ref`](https://github.com/skatejs/skatejs/blob/master/src/api/vdom.js#L12) are used in that file and `ref` is used internally, and probably shouldn't actually be used outside as it should be expected that function helpers should be written with a specific version in mind. Cross-compat as far as API is concerned is limited to `name` so that you can use constructors for other versions because there's no spec'd way to do that.

treshugart: Meant to comment below, but I'll keep this and reference it.

src/util/symbols.js
treshugart: See https://github.com/skatejs/skatejs/pull/866#discussion_r83575988

README.md
bradleyayers: I think where our mindset might differ is that I treat the value of `skate.prop.slot()` **as a black box**. I definitely wouldn't expect `Object.assign(skate.prop.slot(), …)` to have any meaningful or predictable behaviour.

I get the sense that you're treating the value of `skate.prop.slot()` **not** a black box, and instead `skate.prop.slot()` is really just syntax sugar.

Do you think that's a fair assessment?

src/api/prop.js
bradleyayers: I think I was thrown off by the names of the variables. I read the cost top to bottom and stumbled on all of them. Perhaps I didn't have enough context to the problem-being-solved before reading it.

treshugart: Happy to remove or leave.

README.md
treshugart: Yeah, syntactic sugar for returning a prop definition. It's a function to ensure it's not mutated causing side effects on other components. The argument to it is syntactic sugar for `Object.assign(skate.slot.prop(), { name: 'something' });`.

test/unit/api/properties.js
andreawyss: Can you please add a few comments into the code of this PR? 
I have a hard time understanding how the code works. Thank you.

treshugart:
I'm going to close this as we've lived without it for this long. I'd like to see what happens with #995 and then revisit this with a fresh mind.

PR #867, opened 2016-10-16
#710 - remove support for v0

### LINKED ISSUE #710: Drop support for v0
This allows us to remove any code related to Custom Elements / Shadow DOM V0, dropping a bit of weight and code to normalise behaviour for Custom Elements.

treshugart:
Fixes #710.

Since v1 is support in Chrome, there's no need to support v0 even if Opera hasn't updated yet. Until they pick up the new Blink (if they haven't already) people wanting to support opera can use the polyfills.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ const skate = window.skate;
 
 ## Dependencies
 
-Skate doesn't require you provide any external dependencies, but recommends you provide some web component polyfills depending on what browsers you require support for.
+Skate doesn't require you provide any external dependencies, but recommends you provide some web component polyfills depending on what browsers you require support for. **Skate requires both Custom Elements and Shadow DOM v1.**
 
 To get up and running quickly with our recommended configuration, we've created a single package called [`skatejs-web-components`](https://github.com/skatejs/web-components) where all you have to do is load it before Skate.
 
@@ -215,8 +215,6 @@ Or you can use script tags:
 ```
 
 If you want finer grained control about which polyfills you use, you'll have to BYO Custom Element and Shadow DOM polyfills.
-
-*Skate will work without Shadow DOM support, but you won't be able to compose components together due to the lack of DOM encapsulation that Shadow DOM gives you.*
 
 
 
--- a/README.md
+++ b/README.md
@@ -214,9 +214,9 @@ Or you can use script tags:
 <script src="https://unpkg.com/skatejs/dist/index-with-deps.min.js"></script>
 ```
 
-If you want finer grained control about which polyfills you use, you'll have to BYO Custom Element and Shadow DOM V1 polyfills. Skate will work without Shadow DOM support, but you won't be able to compose components together due to the lack of DOM encapsulation that Shadow DOM gives you.
+If you want finer grained control about which polyfills you use, you'll have to BYO Custom Element and Shadow DOM polyfills.
 
-**Skate works with the native V0 APIs, however, the V0 polyfills have several bugs that are inconsistent with the native APIs, so it's not recommended you try and use the V0 polyfills.**
+*Skate will work without Shadow DOM support, but you won't be able to compose components together due to the lack of DOM encapsulation that Shadow DOM gives you.*
 
 
 
@@ -251,9 +251,7 @@ If you have any questions about Skate you can use one of these:
 
 Let's define some terms used in these docs:
 
-- v0 - the original Blink implementation of web components - when the spec was still contentious.
-- v1 - the non-contentious - modern-day - specs.
-- polyfill, polyfilled, polyfill-land - not v0 or v1; no native custom element support at all.
+- polyfill, polyfilled, polyfill-land - no native web component support.
 - upgrade, upgraded, upgrading - when an element is initialised as a custom element.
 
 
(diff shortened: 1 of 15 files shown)


src/api/component.js
```diff
@@ -24,16 +21,10 @@ import syncPropToAttr from '../util/sync-prop-to-attr';
...
+// Uses Shadow DOM only if it's available. Once all browsers support it
+// natively then we can remove this function and call it directly.
 function attachShadow (elem) {
-  if (shadowDomV1) {
-    return elem.attachShadow({ mode: 'open' });
-  } else if (shadowDomV0) {
-    return elem.createShadowRoot();
-  }
-  return elem;
+  return elem.attachShadow ? elem.attachShadow({ mode: 'open' }) : elem;
```
zzarcon: Shouldn't `attachShadow` be always available now?

zzarcon: Oh that's true, forgot my comment then! ;)

nadiam84:
`test/unit/api/define.js` this file also has references to the v0 custom elements.

treshugart:
@nadiam84 I removed those references. Thanks!

README.md
```diff
@@ -214,9 +214,9 @@ Or you can use script tags:
 <script src="https://unpkg.com/skatejs/dist/index-with-deps.min.js"></script>
 ```
 
-If you want finer grained control about which polyfills you use, you'll have to BYO Custom Element and Shadow DOM V1 polyfills. Skate will work without Shadow DOM support, but you won't be able to compose components together due to the lack of DOM encapsulation that Shadow DOM gives you.
+If you want finer grained control about which polyfills you use, you'll have to BYO Custom Element and Shadow DOM polyfills.
 
-**Skate works with the native V0 APIs, however, the V0 polyfills have several bugs that are inconsistent with the native APIs, so it's not recommended you try and use the V0 polyfills.**
+*Skate will work without Shadow DOM support, but you won't be able to compose components together due to the lack of DOM encapsulation that Shadow DOM gives you.*
```
bradleyayers: I wouldn't document this, as it makes it API. I would say

> Skate requires Custom Elements v1 and Shadow DOM v1 support. Use the skatejs-web-components library to polyfill browsers lacking native support.

src/api/define.js
```diff
@@ -107,42 +106,22 @@ function prepareForRegistration (name, Ctor) {
...
     throw new Error(`You have to define options to register a component ${name}`);
   }
   const Ctor = typeof opts === 'object' ? Component.extend(opts) : opts;
   formatLinkedAttributes(Ctor);
 
-  if (customElementsV1) {
-    return registerV1Element(name, Ctor);
-  } else if (customElementsV0) {
-    return registerV0Element(name, Ctor);
+  if (!window.customElements) {
```
bradleyayers: I'd move this up to the top where the other assertions are made.

test/unit/vdom/shadow-dom.js
```diff
@@ -1,34 +1,23 @@
...
-      expect(elem[symbols.shadowRoot]).not.to.equal(elem);
-    });
-  }
-
-  if (shadowDomV1) {
-    it('should work for attachShadow()', () => {
-      const elem = new Elem();
+  it('should work for attachShadow()', () => {
+    const elem = new Elem();
+    if (elem.attachShadow) {
```
bradleyayers: I'd assume existence of `attachShadow`.

test/unit/vdom/shadow-dom.js
```diff
@@ -1,34 +1,23 @@
...
       expect(elem[symbols.shadowRoot]).not.to.equal(elem);
```
bradleyayers: Nuke `symbols.shadowRoot`

src/api/component.js
```diff
@@ -257,18 +236,13 @@ Component.prototype = Object.create(HTMLElement.prototype, {
     }
   },
 
-  // v0 and v1
+  // v1
```
treshugart: Ahh yep, should probably convert that to `// spec` maybe? It's nice to differentiate in the code what is Skate and what is spec'd. What should we do when v2 comes? Will possibly want to differentiate between those, in that case this is future-proof :D

src/api/define.js
```diff
@@ -107,42 +106,22 @@ function prepareForRegistration (name, Ctor) {
...
+    throw new Error('Skate requires native custom element support or a polyfill.');
   }
 
-  throw new Error('Skate requires native custom element support or a polyfill.');
+  let uniqueName = name;
+  if (window.customElements.get(name)) {
+    uniqueName = generateUniqueName(name);
+  }
+  prepareForRegistration(uniqueName, Ctor);
```
nadiam84: No need in this function anymore, you can just move those two lines here.

src/api/define.js
treshugart: Do you think it's worth keeping here to enable async loading of polyfill / Skate?

test/unit/vdom/elements.js
```diff
@@ -136,49 +135,6 @@ describe('vdom/elements', () => {
...
-        function assertSlotElement () {
-          expect(ch1.tagName).to.equal('SLOT', 'vdom');
-          expect(ch1.getAttribute('name')).to.equal('test', 'vdom');
-          expect(ch2.tagName).to.equal('SLOT', 'vdom.element(slot)');
-          expect(ch2.getAttribute('name')).to.equal('test', 'vdom.element(slot)');
-        }
-
-        if (shadowDomV1) {
-          assertSlotElement();
-        } else if (shadowDomV0) {
```
treshugart: We've removed support for no shadow DOM, so the only thing we need to test now is SD v1.

src/api/define.js
treshugart: I might keep this for now if that's okay? I'm shuffling this around in #873 anyways and this is less churn to merge.

src/api/define.js
bradleyayers: I meant moving it to line 113 (not the top of the file).

test/unit/vdom/elements.js
bradleyayers: Isn't that what line 168 was doing?

PR #869, opened 2016-10-16
test: Add tests to expose bugs in properties handling.

### LINKED ISSUE #796: prop `set` callback is not called when using MutationObserver custom element polyfill
Given a component that has a prop with a `set` callback, it's not called for the initial value when a custom element polyfill (that uses MutationObserver) is used.

``` javascript
define('x-test', {
  props: {
    foo: {
      set: (elem, data) => {
      // not called
      },
    },
   },
});
```

When this element is written to the DOM using Incremental DOM and properties are set synchronously, skate doesn't have a chance to jump in with its `Object.defineProperty()` calls to register the `set` callback.

This means that the `set` callback is never called the first time a property is set in environments like Firefox.

treshugart:
cc @bradleyayers just opened this to see what the build status was like.

Intended to fix #796.


treshugart:
@bradleyayers updated to fix the problem and to add a test for the same possible problem in native (since the spec says it's possible)

test/unit/lifecycle/properties.js
```diff
@@ -455,6 +455,80 @@ describe('lifecycle/property', () => {
...
+                  expect(data.newValue).to.equal('bar');
+                  expect(data.oldValue).to.equal(null);
+                  ++calls;
+                }
+              }
+            }
+          });
+
+          afterMutations(
+            () => fixture(`<${safe} foo="bar" />`),
```
bengummer: how does the `safe` tagName here get related to the `skate(...)` call above? seems like magic :)

test/unit/vdom/properties.js
```diff
@@ -74,6 +75,22 @@ describe('vdom/properties', () => {
...
+  it('should set props as properties instead of attributes on ctor references', () => {
+    let fixture = document.createElement('div');
+    const Elem = define('x-test', {
+      props: {
+        foo: {}
+      }
+    });
+
+    IncrementalDOM.patch(fixture, () => {
+      vdom.element(Elem, { foo: 'bar' });
```
bradleyayers: I'd throw in another attribute here that *isn't* declared as a prop, and then assert that it *wasn't* set as a prop, assuming that's the desired behaviour.

test/unit/lifecycle/properties.js
treshugart: It is magic. #891 is for refactoring that out and making this a bit simpler using just `define()`.

test/unit/vdom/properties.js
treshugart: Done.

PR #871, opened 2016-10-17
fix(util): fix mobile safari 9.0 - mo + microtask debounce

jpnelson:
native MutationObserver feature detection was incorrect for mobile safari 9.0 . Also, microtasks

were not debounced correctly

### DIFF
--- a/src/util/native.js
+++ b/src/util/native.js
@@ -1,1 +1,7 @@
-export default fn => (fn || '').toString().indexOf(['native code']) > -1;
+const nativeHints = [
+  'native code',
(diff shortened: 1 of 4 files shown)


src/util/native.js
```diff
@@ -1 +1,7 @@
-export default fn => (fn || '').toString().indexOf(['native code']) > -1;
+const nativeHints = [
+  'native code',
+  '[object MutationObserverConstructor]' // for mobile safari iOS 9.0
+];
+export default fn => nativeHints.map(
```
bradleyayers: Wouldn't `.some` be better here?

Issue #872, opened 2016-10-19
Closer to the platform


treshugart:
Skate components should be definable via `customElements.define()` and `skate.define()` should become a syntactic sugar for anonymising components, and that's all. This piggy backs off #867 which implements #710.

zzarcon:
This means that we will deprecated `skate.define` at some point and encourage the users to just pass skate components to `customElements.define`?

treshugart:
@zzarcon we'll see. I need to spike how we could do unique tag names without `define()`.

treshugart:
So it turns out `define()` won't be deprecated. It gives you the ability to anonymise components, or use Webpack's HMR.

PR #873, opened 2016-10-19
#872 - closer to the platform

### LINKED ISSUE #875: Map out and test API points that enable cross-version compatibility
There have been some issues in the past with multiple versions breaking each other. While these are far and few between, they've generally arisen because the API points that versions need to stay compatible are not properly tested.

We need to map these out and explicitly test that they exist and behave as normal. Some of these are (not exhaustive, remove this comment when it is):
- `WebComponent[skate.symbols.name]`

treshugart:
Brings in changes from #710 since there's is a bit of overlap.

Fixes #875 and #891.


src/api/component.js
```diff
@@ -194,81 +133,66 @@ Component.updated = function updated (elem, prev) {
...
-// Calls the user-defined updated() lifecycle callback.
-Component[$updated] = function _updated (elem) {
-  if (typeof this.updated === 'function') {
-    const prev = elem[$props];
-    elem[$props] = getSetProps(elem);
-    return this.updated(elem, prev);
-  }
-  return true;
+// DEPRECATED
+Component.rendered = function _rendered () {};
```
zzarcon: Should we log any kind of warning message for deprecated methods?

src/api/component.js
```diff
@@ -293,31 +217,78 @@ Component.prototype = Object.create(HTMLElement.prototype, {
         attributeChanged(this, { name, newValue, oldValue });
       }
     }
-  },
+  }),
+
+  // Skate
+  //
+  // This is a default implementation that does strict equality copmarison on
```
zzarcon: `comparison` 😄

src/api/component.js
```diff
@@ -3,44 +3,45 @@ import {
...
+function syncPropsToAttrs (elem) {
+  const props = elem.constructor.props;
+  Object.keys(props).forEach((propName) => {
+    const prop = props[propName];
+    syncPropToAttr(elem, prop, propName, true);
+  });
 }
 
-function callConstructor (elem) {
+// v1
```
treshugart: Ah yep, totes.

src/api/component.js
treshugart: That's a good point. I think that's a more involved discussion in terms of how do we do it / what do we use. It'd have to be ignored in node prod etc. Happy to issue a second pr for this along side it (just so it doesn't block this going in). Any ideas?

src/api/component.js
treshugart: :(

README.md
```diff
@@ -319,86 +321,116 @@ import { define, vdom } from 'skatejs';
...
-  props: { ...props }
+customElements.define('my-component', class extends skate.Component {
+  static get props () {
+    return {};
+  }
 });
 ```
 
-Custom properties, when set, queue a `render()`. This happens after a `setTimeout()` so that you only trigger a single render for a series of property sets.
+Custom properties, when set, queue a `renderCallback()`. Property sets queue a render, so if you are setting multiple properties, only a single render will happen.
```
bradleyayers: I think this could be worded more simply:

> Declare properties that (when mutated) should cause the component re-render.

README.md
```diff
@@ -687,62 +743,64 @@ elem.arr.push('something');
...
+      skate.emit(this, 'name-changed', { detail: prev });
     }
   }
 });
 ```
 
 
 
-#### `render`
+#### `renderCallback`
```
bradleyayers: A few thoughts:

- It would be nice to more concretely describe what the return value is expected to be. At the moment it's a bit fuzzy and forces the reader to derive it from the examples.
- It would be useful to mention that in older versions of skate this was called `render`, else I'm sure there'll be some confusion when people try to figure out how `render` differs from `renderCallback`.

README.md
```diff
@@ -687,62 +743,64 @@ elem.arr.push('something');
...
 
-Function that is called to render the element. This is called when the element is first created and on subsequent prop updates if the property `updated()` callback returns `true`.
+Function that is called to render the element. This is called when the element is first created and on subsequent prop updates if `updatedCallback()` callback returns `true`.
```
bradleyayers: Rather than regurgitate the lifecycle of a component, I'd probably just say:

> Function that is responsible for rendering the Shadow DOM of the component.

And perhaps at the bottom of the section have a:

> See also:
>     - [`updatedCallback()`](...)
>     - [`renderedCallback()`](...)

README.md
treshugart: Yeah, I still have to comb the readme for the legacy methods and denote the new ones. Part of the legacy API for `renderCallback()` was that you could just put Incremental DOM in there (and you still can). It's worth noting ideal and the possibilities, though. I'll note that for when I update it.

README.md
treshugart: I like that much more. Thanks!

README.md
treshugart: Updated.

README.md
treshugart: Updated (but only for this instance) for now.

README.md
```diff
@@ -64,26 +66,29 @@ Whenever you change the `name` property - or attribute - the component will re-r
...
-        - [`default`](#default)
-        - [`deserialize`](#deserialize)
-        - [`get`](#get)
-        - [`initial`](#initial)
-        - [`serialize`](#serialize)
-        - [`set`](#set)
-      - [`created`](#created)
-      - [`updated`](#updated)
+    - [Using the Platform](#using-the-platform)
+    - [`constructor` - supercededs `static created()`](#constructor---supercededs-static-created)
```
nadiam84: `supercededs` is a really weird word. My English is not as good as I hoped or it really just means `supersedes`?

nadiam84: Ha, google says there is such word. My life will never be the same.

treshugart: lol :( yep I did

treshugart: Spell check didn't do it's thang!

README.md
```diff
@@ -588,18 +655,20 @@ The parameters passed to the function are:
...
 A function that is called whenever the property is set. This is also called when the property is first initialised.
 
 ```js
-skate.define('my-component', {
-  props: {
-    myProp: {
-      set (elem, data) {
-        // do something
+customElements.define('my-component', class extends skate.Component {
+  static get props () {
```
nadiam84: maybe `set` here?

README.md
```diff
@@ -588,18 +655,20 @@ The parameters passed to the function are:
...
+    return {
+      myProp: {
+        set (elem, data) {
+          // do something
```
nadiam84: Could you please mention this issue (https://github.com/skatejs/skatejs/pull/866#pullrequestreview-4409414) here, if this is still the case?

README.md
```diff
@@ -1268,7 +1281,7 @@ When `vdom.builder()` is called with arguments, it returns an array of functions
 
 ```js
 const [ div, p ] = skate.vdom.builder('div', 'p');
-define('my-component', {
+customElements.define('my-component', class extends skate.Component {
   render() {
```
nadiam84: should this be a `renderCallback` as well?

src/api/component.js
```diff
@@ -3,21 +3,83 @@ import {
...
+    descriptors[descriptorName] = props[descriptorName](descriptorName);
+    return descriptors;
+  }, {});
+}
+
+function createInitProps (Ctor) {
+  const props = ensurePropertyDefinitions(Ctor);
+
+  return (elem) => {
+    if (!props) {
```
nadiam84: When exactly `props` can be falsy? `ensurePropertyDefinitions` will always return at least an empty object (since it's an initial value), isn't it?

nadiam84:
Lots of changes and no new tests at all are really scary

src/api/component.js
treshugart: This is legacy code, I'd rather not touch it.

treshugart:
@nadiam84 I didn't want to rewrite any of the tests initially because I wanted to make sure the refactors would be backward compatible. Will rewrite / add appropriate tests prior to merging.

README.md
```diff
@@ -317,86 +324,128 @@ import { define, vdom } from 'skatejs';
...
+If you're using IE9 and 10, transpiling class extension is limited to instance methods and properties. Statics don't get added to the chain. However, Skate supports calling `extend()` on a component which you want to extend. This means that in older browsers, you can do something like:
 
-const MyComponent = define('my-component', class extends Component {});
+```js
+const MyComponent1 = skate.Component.extend();
+const MyComponent2 = MyComponent1.extend();
 ```
 
-A simpler way, especially for ES5 users, would be to just pass an object literal.
+Recently, we've deprecated several old methods in favour of aligning closer to the native APIs. These methods still work but will be removed in a future version. Not all methods / properties are listed here, only the ones that have been deprecated and what they're superceded by.
```
treshugart: Thought I fixed that, but I think that was a search and replace of the misspelling. It should be supersede.

README.md
treshugart: I'm not sure what to do here. The line you've commented on says `// do something` and that PR comment is about clearing up if prop functions are a black box or syntactic sugar.

README.md
treshugart: Why `set`? `props` only needs a getter.

README.md
treshugart: Yep, thank you!

README.md
```diff
@@ -1650,20 +1651,16 @@ The nice part about thinking this way is that you get both a declarative and imp
 
 ### Naming Collisions
 
-You may write a component that you change in a backward incompatible way. In order for your users to upgrade, they'd have to do so all at once instead of incrementally if you haven't given the new one a different name. You could rename your component so it can co-exist with the old one, or you can create a function that allows your users to define a name for your component:
+You may write a component that you change in a backward incompatible way. In order for your users to upgrade, they'd have to do so all at once instead of incrementally if you haven't given the new one a different name. One option is to choose a different name for your component, but that may not be idea. You could also use `skate.define()` to ensure the name is unique. An ideal solution would be to only export your constructor and let the consumer register it.
```
andreawyss: idea -> ideal

README.md
```diff
@@ -618,35 +687,35 @@ When the property is initialised, `oldValue` will always be `undefined` and `new
...
+#### `updatedCallback` - supersedes `static updated()`
 
-Called before `render()` after `props` are updated. If it returns falsy, `render()` is not called. If it returns truthy, `render()` is called.
+Called before `renderCallback()` after `props` are updated. If it returns falsy, `renderCallback()` is not called. If it returns truthy, `renderCallback()` is called.
 
 ```js
-skate.define('x-component', {
-  updated(elem, prevProps) {
+customElements.define('x-component', class extends skate.Component {
+  updatedCallback (previousState) {
```
treshugart: I'm not sure why I did that TBH because you're right.

README.md
```diff
@@ -269,26 +268,34 @@ The following is a simple counter that increments the count for every second tha
...
+      // for the count as part of the HTML.
+      count: skate.prop.number({ attribute: true })
+    };
+  }
+  connectedCallback () {
+    // Ensure we call the parent.
+    super.connectedCallback();
+
     // We use a symbol so we don't pollute the element's namespace.
     elem[sym] = setInterval(() => ++elem.count, 1000);
```
treshugart: Yep, it does, I just missed it here. Thanks!

README.md
```diff
@@ -1435,9 +1449,9 @@ skate.h('button', { onclick: onClick });
...
 
 ```js
-skate.define('x-element', {
-  created(elem) {
-    elem.addEventListener('change', elem.handleChange);
+customElements.define('my-element', class extends skate.Component {
+  constructor () {
+    this.addEventListener('change', this.handleChange);
   },
   prototype: {
```
treshugart: I also need to add a `super()` call.

README.md
treshugart: Thanks!

vvakame:
any update? 👀 
I' really looking forward to the merge of this PR 😋

treshugart:
I'm hoping to get this polished off tonight and do a feature release with
it.

vascofernandes:
I like this changes. Very good.

treshugart:
OMGOMG it's finally green!!

src/api/define.js
```diff
@@ -1,146 +1,36 @@
...
 
   if (!customElements) {
     throw new Error('Skate requires native custom element support or a polyfill.');
   }
 
-  let uniqueName = name;
-  if (customElements.get(name)) {
-    uniqueName = generateUniqueName(name);
+  // Support passing an anonymous definition.
+  if (!opts) {
```
nadiam84: To check, just in case: if I pass just a name here (like an idiot, yeah), it will fail. Should it do that? Or maybe some sort of warning would be better?
```
define('x-weird-custom-element');
```

src/api/vdom.js
```diff
@@ -140,17 +140,19 @@ const attributesContext = propContext(attributes, {
...
 function resolveTagName (tname) {
-  // If the tag name is a function, a Skate constructor or a standard function
-  // is supported.
-  //
-  // - If a Skate constructor, the tag name is extracted from that.
-  // - If a standard function, it is used as a helper.
-  if (typeof tname === 'function') {
-    return tname[$name] || tname;
+  if (!tname) {
+    return tname;
```
nadiam84: Is there some benefit of doing this instead of just returning `undefined` or `null`? Seems a bit weird.

treshugart: Agree. Possibly an artefact of the refactor and trying to keep it mostly the same.

Issue #876, opened 2016-10-21
Apply prop attributes as properties in a polyfill environment


bradleyayers:
A problem exists when composing skate component in an environment that asynchronously upgrades custom elements (e.g. via a `MutationObserver` based polyfill). This happens when a parent component tries to pass complex data onto a child component, that hasn't yet been upgraded.

The parent checks the constructor of the child element for a `props` attribute, and uses this information to decide whether or not to set as attribute as a property. When set as a property, it's possible to pass non-primitive values onto the child, but this only occurs if the attribute is declared as a prop.

The consequence is that it's currently impossible to pass functions onto child components on the initial render. On subsequent renders the issue does not occur, because the child has already been upgraded.

A few ideas of how to solve this:

1. Don't rely on checking the constructor of an element, instead look up its tag name in the custom elements register. (i.e. `customElements.get(elem.tagName).props` rather than `elem.constructor.props`).
2. Synchronously upgrade custom elements before checking `elem.constructor.props` (this could be done in the idom hooks).

PR #883, opened 2016-10-25
#876 - fix polyfilled component props / attrs that haven't been upgraded yet

treshugart:


test/unit/vdom/properties.js
```diff
@@ -117,4 +117,63 @@ describe('vdom/properties', () => {
...
+        const Elem1 = define('x-test', {
+          render (elem) {
+            return h(Elem2, { fooBar: true });
+          }
+        });
+        const Elem2 = define('x-test', {
+          observedAttributes: ['fooBar'],
+          render(elem) {
+            expect(elem.fooBar).to.be.equal(undefined);
+            expect(elem.getAttribute('foobar')).to.equal('true');
```
nadiam84: A bit inconsistent with everything else? Why it wasn't transform into `foo-bar`?

treshugart: I thought about that. I ended up just making it be a slight variant of the test above it. Come to think of it, i'll test both plus do the empty case all in the same test.

treshugart: I ended up refactoring the tests into a single one because we couldn't used `observedAttributes`.

test/unit/vdom/properties.js
```diff
@@ -117,4 +117,36 @@ describe('vdom/properties', () => {
...
+    const Elem2 = define('x-test', {
+      props: {
+        foo: {}
+      },
+      render(elem) {
+        expect(elem.foo).to.be.equal(true);
+        expect(elem.bar).to.be.equal(undefined);
+        expect(elem.baz).to.equal(true);
+        expect(elem.hasAttribute('foo')).to.equal(false);
+        expect(elem.getAttribute('bar')).to.equal('true');
```
nadiam84: Don't understand why this is 'true'. Where does it come from?

bengummer: as far as i understand, vdom sets as prop if `elem.hasOwnProperty(name) === true` otherwise it sets as attr

test/unit/vdom/properties.js
```diff
@@ -117,4 +117,36 @@ describe('vdom/properties', () => {
       );
     });
   });
+
+  it('#876 - Apply properties and attributes in a polyfill environment', (done) => {
+    const Elem1 = define('x-test', {
+      render (elem) {
+        return h(Elem2, { foo: true, bar: true, baz: true });
```
nadiam84: Up until now I was sure that camelCased props transform into attributes with dashes. I.e. 'testProp' becomes 'test-prop'.
But in this test 'testProp' is transforming into 'testprop'. Why?

test/unit/vdom/properties.js
treshugart: `setAttribute()` does a `toString()` on anything you give it

test/unit/vdom/properties.js
treshugart: The `*Attribute()` methods `toLowerCase()` the first argument.

treshugart: The camelCasing / dash-casing is handled by the property lifecycles when `{ attribute: true }` is specified.

PR #894, opened 2016-10-31
fix: Fix issue causing the importing of Skate into Node to fail.

### LINKED ISSUE #893: You should be able to include Skate in node
In apps that SSR and include Skate as a dependency, it shouldn't throw errors even if it won't create custom elements without a DOM impl.

treshugart:
It shouldn't fail to import / require() even though it may not be able to create custom elements.

fixes #893

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
   ],
   "dependencies": {
     "incremental-dom": "0.4.1",
-    "skatejs-build": "9.x"
+    "skatejs-build": "9.x",
+    "window-or-global": "^1.0.1"
   },
   "devDependencies": {
     "birdpoo": "0.x",
@@ -35,7 +36,7 @@
     "skatejs-web-components": "2.0.1"
   },
   "scripts": {
-    "test": "sk-tests",
+    "test": "node dist/ && sk-tests",
     "prepublish": "sk-bundle"
   },
   "config": {
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,5 +1,6 @@
 module.exports = require('skatejs-build/rollup.config');
 module.exports.globals = {
   'incremental-dom': 'IncrementalDOM',
-  'regex-native-function': 'isNativeRegex'
+  'regex-native-function': 'isNativeRegex',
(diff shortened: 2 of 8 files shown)


rollup.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = require('skatejs-build/rollup.config');
 module.exports.globals = {
   'incremental-dom': 'IncrementalDOM',
-  'regex-native-function': 'isNativeRegex'
```
treshugart: Boyscout fix as we don't use this anymore.

src/api/component.js
```diff
@@ -14,8 +14,10 @@ import getAllKeys from '../util/get-all-keys';
 import getOwnPropertyDescriptors from '../util/get-own-property-descriptors';
 import getSetProps from './props';
 import syncPropToAttr from '../util/sync-prop-to-attr';
+import root from 'window-or-global';
 
-const { HTMLElement } = window;
+const { HTMLElement } = root;
+const htmlElementPrototype = HTMLElement ? HTMLElement.prototype : {};
```
treshugart: If `HTMLElement` isn't available, this would cause errors below. Instead, we just create a prototype from the base `Object` even if it's not used.

package.json
```diff
@@ -35,7 +36,7 @@
     "skatejs-web-components": "2.0.1"
   },
   "scripts": {
-    "test": "sk-tests",
+    "test": "node dist/ && sk-tests",
```
treshugart: Fail early with running in node as this takes less time. Node will throw non-zero here, so this should be sufficient.

package.json
```diff
@@ -24,7 +24,8 @@
   ],
   "dependencies": {
     "incremental-dom": "0.4.1",
-    "skatejs-build": "9.x"
+    "skatejs-build": "9.x",
```
nadiam84: It's 10.x btw;)

bengummer:
does this mean server side rendering is coming?!

treshugart:
@bengummer not yet, but it means that it can at least be imported into SSR apps without causing issues.

PR #902, opened 2016-11-03
fix: Fix issues integrating with the polymer / webcomponents polyfills.

### LINKED ISSUE #901: Integrate skatejs-web-components with Polymer / Webcomponent polyfills
This removes the usage of the skatejs-named-slots polyfill and pulls in the polyfills as stated in 
https://github.com/skatejs/web-components/issues/26 and implemented in https://github.com/skatejs/web-components/pull/27.

This is essentially only for the purposes of testing, but initial tests uncovered a minor issue with the ShadyDOM polyfill, so there are a couple code changes that have to be made. More explained in the PR.

This should be treated as a patch release to fix issues integrating with those polyfills as well as simplifying event handling code.

treshugart:
Simplified event handling code and updated tests to be more clear.

fixes #901

### DIFF
--- a/karma.browsers.js
+++ b/karma.browsers.js
@@ -1,0 +1,1 @@
+module.exports = require('skatejs-build/karma.browsers');
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "react": "15",
     "react-dom": "15",
     "skatejs-build": "12",
-    "skatejs-web-components": "2"
+    "skatejs-web-components": "3"
   },
   "scripts": {
     "test": "node dist/ && sk-tests",
--- a/package.json
+++ b/package.json
@@ -27,11 +27,11 @@
     "window-or-global": "1.0.1"
   },
   "devDependencies": {
-    "birdpoo": "0",
-    "react": "15",
-    "react-dom": "15",
-    "skatejs-build": "12",
-    "skatejs-web-components": "3"
+    "birdpoo": "0.x",
+    "react": "15.x",
+    "react-dom": "15.x",
+    "skatejs-build": "12.x",
+    "skatejs-web-components": "3.x"
   },
   "scripts": {
     "test": "node dist/ && sk-tests",
--- a/package.json
+++ b/package.json
@@ -24,15 +24,14 @@
   ],
   "dependencies": {
     "incremental-dom": "0.4.1",
-    "window-or-global": "^1.0.1"
+    "window-or-global": "1.0.1"
   },
   "devDependencies": {
-    "birdpoo": "0.x",
-    "document-register-element": "1.x",
-    "react": "15.x",
-    "react-dom": "15.x",
-    "skatejs-build": "9.x",
-    "skatejs-web-components": "2.0.2"
+    "birdpoo": "0",
+    "react": "15",
+    "react-dom": "15",
+    "skatejs-build": "12",
+    "skatejs-web-components": "2"
   },
   "scripts": {
     "test": "node dist/ && sk-tests",
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -45,7 +45,7 @@ function applyEvent (elem, ename, newFunc) {
   }
 
   // Undefined indicates that there is no listener yet.
-  if (events[ename] === undefined) {
+  if (typeof events[ename] === 'undefined') {
     // We only add a single listener once. Originally this was a workaround for
     // the Webcomponents ShadyDOM polyfill not removing listeners, but it's
     // also a simpler model for binding / unbinding events because you only
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -59,7 +59,7 @@ function applyEvent (elem, ename, newFunc) {
   }
 
   // Not undefined indicates that we have set a listener, so default to null.
-  events[ename] = newFunc || null;
+  events[ename] = typeof newFunc === 'function' ? newFunc : null;
 }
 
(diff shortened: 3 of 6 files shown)


src/api/vdom.js
```diff
@@ -44,17 +44,22 @@ function applyEvent (elem, ename, newFunc) {
...
   }
 
-  const oldFunc = events[ename];
-
-  // Remove old listener so they don't double up.
-  if (oldFunc) {
-    elem.removeEventListener(ename, oldFunc);
+  // Undefined indicates that there is no listener yet.
+  if (events[ename] === undefined) {
+    // We only add a single listener once. Originally this was a workaround for
```
treshugart: The original issue is this one https://github.com/webcomponents/shadydom/issues/24. In that test we were adding the same callback more than once, thus they didn't get removed.

karma.browsers.js
```diff
@@ -0,0 +1 @@
+module.exports = require('skatejs-build/karma.browsers');
```
treshugart: Added because of build lib update.

package.json
```diff
@@ -24,15 +24,14 @@
   ],
   "dependencies": {
     "incremental-dom": "0.4.1",
-    "window-or-global": "^1.0.1"
+    "window-or-global": "1.0.1"
```
treshugart: Pinned this to ensure consistency in dependencies.

package.json
```diff
@@ -24,15 +24,14 @@
...
   },
   "devDependencies": {
-    "birdpoo": "0.x",
-    "document-register-element": "1.x",
-    "react": "15.x",
-    "react-dom": "15.x",
-    "skatejs-build": "9.x",
-    "skatejs-web-components": "2.0.2"
+    "birdpoo": "0",
```
treshugart: Boyscout: I realised you could just do this. Removes redundancy in the `.x` suffix.

src/api/vdom.js
```diff
@@ -44,17 +44,22 @@ function applyEvent (elem, ename, newFunc) {
     events = elem[$currentEventHandlers] = {};
   }
 
-  const oldFunc = events[ename];
-
-  // Remove old listener so they don't double up.
-  if (oldFunc) {
-    elem.removeEventListener(ename, oldFunc);
+  // Undefined indicates that there is no listener yet.
```
treshugart: The core change is this.

test/unit/vdom/events.js
```diff
@@ -29,31 +26,11 @@ describe('vdom/events (on*)', () => {
...
-        expect(shadowDiv.textContent).to.equal('1');
-        emit(shadowDiv, 'event');
-        expect(el._test).to.equal(2);
-
-        props(el, { test: el.test + 1 });
-        expect(shadowDiv.textContent).to.equal('2');
-        emit(shadowDiv, 'event');
-        expect(el._test).to.equal(3);
-      },
+      () => expect(el.shadowRoot.textContent).to.equal('0'),
```
treshugart: This essentially tests the same thing in a simpler manner. However the previous test didn't take into account things that may happen after mutations very well, this ensures that anything that might happen in a microtask, should happen, and that things are more direclty tested.

test/unit/vdom/events.js
```diff
@@ -108,16 +85,33 @@ describe('vdom/events (on*)', () => {
...
       }
     }))();
     fixture(myel);
+    let div
     afterMutations(
-      () => {}, // .render()
-      () => {
-        emit(myel.shadowRoot.firstChild, 'test');
-      },
+      () => (div = myel.shadowRoot.firstChild),
```
treshugart: I added several more assertions here since we rely on `null` and `undefined` explicitly in the event binding logic and we need to make sure that if an event is set to one of those that it doesn't screw it up. Also passed falsy / truthy values to ensure that we only allow functions.

package.json
nadiam84: I remember this discussion, that `1.x` is confusing for some people, everyone is used to `^1.0.0`. Don't you think this would be even more confusing?

lukebatchelor:
LGTM 👍

src/api/vdom.js
```diff
@@ -44,17 +44,22 @@ function applyEvent (elem, ename, newFunc) {
...
+  if (typeof events[ename] === 'undefined') {
```
nadiam84: Sure this shouldn't be `!==`?

nadiam84: never mind...

package.json
treshugart: I find specifying the whole thing confusing, TBH. I much prefer this, but maybe that's just me?

src/api/vdom.js
treshugart: No, if it's `undefined`, that means there's no listener yet, so we must bind one in order to trigger anything that gets bound via a property. Later we set the handler in the `events` map.

This is a workaround for https://github.com/webcomponents/shadydom/issues/24 but I like the logic much better anyways. There'd be pretty much zero overhead for adding a single noop listener, and optimising for the common use case, generally (from experience) once you add a listener, you're going to keep it there.

package.json
nadiam84: I prefer `x` :D But there were this discussion and complains about skate's complexity ;) But it's your choice.

src/api/vdom.js
nadiam84: feels like it's breaking change...

treshugart: It's definitely not a breaking change. The API stays the same; nothing changes for the user.

Issue #754, opened 2016-08-29
create a standalone typing module and flowtype module


stevemao:
I think it's worth to create them at some point since these two are quite popular.

treshugart:
I've had this in the back of my mind for awhile, I'm glad you brought it up :) I'm assuming this would take refactoring things into flowtype, correct? I guess we wouldn't have to do this completely up front but I haven't done that much investigation of what this would take.

stevemao:
Oh, I'm not saying to refactor this module. I just want the consumers of this module to use this one without defining types by themselves.

treshugart:
Ahh ok. I'd like to experiment using flowtype because it'd probably resolve #596 without having to do the checking with a library.

bradleyayers:
I've been working on a TypeScript declaration:

```ts
declare module "skatejs" {
  export const vdom: any;

  type PropData = { [prop: string]: any };
  export const props: (elem: any, data?: PropData) => void;
  export const props: (elem: any) => PropData;

  export const emit: (elem: any, eventName: string, eventOptions = {}) => boolean;

  export function define(name: string): <C>(component: C) => C;

  export function define<Proto, Props>(name: string, definition: {
    prototype: Proto;
    props: Props;
    created?(elem: Component & Proto & Props): any;
    updated?(elem: Component & Proto & Props, prevProps: { [name: string]: any }): boolean | undefined;
    render?(elem: Component & Proto & Props): () => any | undefined;
    rendered?(elem: Component & Proto & Props): any;
    attached?(elem: Component & Proto & Props): any;
    detached?(elem: Component & Proto & Props): any;
    attributeChanged?(elem: Component & Proto & Props, data: { name: string, oldValue: any, newValue: any }): any;
    observedAttributes?: string[];
  }): { new(...args: any[]): Component & Proto & Props };

  export const symbols: {
    shadowRoot: string | symbol;
    name: string | symbol;
  };

  export const ready: (elem: HTMLElement, callback: (...args: any[]) => any) => void;

  type HOut = () => any;
  type HChildren = string | HTMLElement | null | HOut | HChildren[];
  type HBuilder = (tag: string | Component, attrs?: { [name: string]: string }, ...children: HChildren[]) => HOut;
  export const h: HBuilder;

  export const builder: (...args: string[]) => HBuilder[];

  type PropOptions = {
    attribute?: boolean | string;
    coerce?: (value: any) => any;
[...]

PR #906, opened 2016-11-07
fix: Add TypeScript definitions.

bradleyayers:
This aims to resolve the TypeScript portion of #754, by introducing a first-stab at definitions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   },
   "main": "dist/index.js",
   "jsnext:main": "src/index.js",
+  "types": "src/index.d.ts",
   "keywords": [
     "components",
     "custom",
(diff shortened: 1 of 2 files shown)


bradleyayers:
Testing is tricky, but possible. The DefinitelyTyped repo has some nice infrastructure setup to test `.d.ts` files against example `.ts` files, e.g.

- [`react-global-tests.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/react/react-global-tests.ts)
- [`react-global.d.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/react/react-global.d.ts)

Something along these lines would make sense, but I don't have time at the moment to work on this.

PR #909, opened 2016-11-10
#908 - Fix Safari TP and @webcomponents/custom-element polyfill

### LINKED ISSUE #908: Fix Safari TP
This is because of a bug in document-register-element where it's always polyfilling and doesn't seem to find the constructor properly to initialise. I didn't have time to investigate it, so I switched to using the webcomponents polyfill instead, which got us past that but now there's two more errors:

- The `@webcomponents/custom-element` polyfill is calling `attributeChangedCallback()` twice
- Safari Tech Preview's `customElements.get()` name parameter is case sensitive

This PR is for moving to the new polyfill and fixing the issues associated with the move.

treshugart:
fixes #908

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,17 +25,14 @@
   ],
   "dependencies": {
     "incremental-dom": "0.4.1",
-    "skatejs-build": "^12.0.1",
     "window-or-global": "1.0.1"
   },
   "devDependencies": {
-    "@webcomponents/custom-elements": "https://github.com/webcomponents/custom-elements/archive/fc022e8.tar.gz",
-    "@webcomponents/shadycss": "https://github.com/webcomponents/shadycss/archive/19a46c9.tar.gz",
-    "@webcomponents/shadydom": "https://github.com/webcomponents/shadydom/archive/c162774.tar.gz",
     "birdpoo": "0.x",
     "react": "15.x",
     "react-dom": "15.x",
-    "skatejs-build": "12.x"
+    "skatejs-build": "12.x",
+    "skatejs-web-components": "3.x"
   },
   "scripts": {
     "test": "node dist/ && sk-tests",
--- a/package.json
+++ b/package.json
@@ -32,7 +32,9 @@
     "react": "15.x",
     "react-dom": "15.x",
     "skatejs-build": "12.x",
-    "skatejs-web-components": "3.x"
+    "@webcomponents/custom-elements": "https://github.com/webcomponents/custom-elements/archive/fc022e8.tar.gz",
+    "@webcomponents/shadycss": "https://github.com/webcomponents/shadycss/archive/19a46c9.tar.gz",
+    "@webcomponents/shadydom": "https://github.com/webcomponents/shadydom/archive/c162774.tar.gz"
   },
   "scripts": {
     "test": "node dist/ && sk-tests",
--- a/package.json
+++ b/package.json
@@ -25,16 +25,17 @@
   ],
   "dependencies": {
     "incremental-dom": "0.4.1",
+    "skatejs-build": "^12.0.1",
     "window-or-global": "1.0.1"
   },
   "devDependencies": {
+    "@webcomponents/custom-elements": "https://github.com/webcomponents/custom-elements/archive/fc022e8.tar.gz",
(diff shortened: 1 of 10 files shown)


treshugart:
It did because of https://github.com/w3c/webcomponents/issues/604.

src/api/vdom.js
```diff
@@ -89,7 +89,7 @@ const attributesContext = propContext(attributes, {
 
   // Default attribute applicator.
   [symbols.default] (elem, name, value) {
-    const { props, prototype } = customElements.get(elem.tagName) || {
+    const { props, prototype } = customElements.get(elem.tagName.toLowerCase()) || {
```
treshugart: @bengummer this is what Safari TP would fail on before.

src/api/component.js
```diff
@@ -19,6 +20,22 @@ import root from 'window-or-global';
...
 
+
+
+// Prevent double-calling with polyfill.
+
+const $prevName = createSymbol();
+const $prevOldValue = createSymbol();
+const $prevNewValue = createSymbol();
+
+function preventDoubleCalling (elem, name, oldValue, newValue) {
```
treshugart: This fixes the polyfill.

src/api/component.js
```diff
@@ -237,6 +254,16 @@ Component.prototype = Object.create(htmlElementPrototype, {
   attributeChangedCallback: {
     configurable: true,
     value (name, oldValue, newValue) {
+      // Polyfill calls this twice.
+      if (preventDoubleCalling(this, name, oldValue, newValue)) {
+        return;
+      }
+
+      // Set data so we can prevent double calling if the polyfill.
+      this[$prevName] = name;
```
treshugart: The other part of fixing the polyfill.

src/api/vdom.js
```diff
@@ -89,7 +89,7 @@ const attributesContext = propContext(attributes, {
 
   // Default attribute applicator.
   [symbols.default] (elem, name, value) {
-    const { props, prototype } = customElements.get(elem.tagName) || {
+    const { props, prototype } = customElements.get(elem.localName) || {
```
treshugart: Fun fact, seems `localName` returns lowercase. We should be using it anyways, but it's a bonus it's lowercase.

package.json
```diff
@@ -32,7 +32,9 @@
     "react": "15.x",
     "react-dom": "15.x",
     "skatejs-build": "12.x",
-    "skatejs-web-components": "3.x"
+    "@webcomponents/custom-elements": "https://github.com/webcomponents/custom-elements/archive/fc022e8.tar.gz",
+    "@webcomponents/shadycss": "https://github.com/webcomponents/shadycss/archive/19a46c9.tar.gz",
+    "@webcomponents/shadydom": "https://github.com/webcomponents/shadydom/archive/c162774.tar.gz"
```
jpnelson: would we keep skatejs/web-components up to date with these?

src/api/component.js
```diff
@@ -19,6 +20,22 @@ import root from 'window-or-global';
...
+
+
+// Prevent double-calling with polyfill.
+
+const $prevName = createSymbol();
+const $prevOldValue = createSymbol();
+const $prevNewValue = createSymbol();
+
+function preventDoubleCalling (elem, name, oldValue, newValue) {
+  return name === elem[$prevName]
```
justinfagnani: Can you send me a repro of the original issue?

treshugart: I've created a basic reproduction of what we're doing internally: http://www.webpackbin.com/EyVC4-Agz.

I think we should just continue being defensive, but this does raise some questions. In that bin:

- Chrome CE v1 text content is 1
- Safari CE v1 text content is 1772 (or whatever, call stack exceeded)
- In polyfill text content is 2

src/api/vdom.js
jpnelson: should we be normalising it here then? Or is `customElements.get` not case sensitive?

treshugart: It's case-sensitive, but `localName` should return the exact name of the element how it was specified in HTML, as I understand. More info here https://github.com/w3c/webcomponents/issues/604.

treshugart:
I'll update the skatejs-web-components stuff later. That PR is taking a bit longer because we need o figure out a way to consume the native-shim in both native and non-native browsers. Conditional requiring is difficult.

PR #926, opened 2016-11-20
Refactor with flow

andreawyss:
@treshugart 
This PR is for previewing my changes. I still need to work on making all the tests pass. 

- Added Flow Type annotations to some files. 
- Added some descriptions to help better understand the code.
- Did some refactoring.
- Cached the property definitions to prevent calling static props multiple times.
- Simpler logic for syncing changes between prop to attribute and attribute to prop.

I left a few question for you in the code look for "todo:"
I'm looking forward for you comments and feedback.

Thank You


treshugart:
Whoah, cool! Thanks @andreawyss. I'm wondering how hard you think it might be to integrate with the changes in #872. Quite a bit changed in the `Component` class there. Thoughts?

src/util/cached-prop-defs.js
```diff
@@ -0,0 +1,83 @@
+//@flow
+import assign from './assign';
+import dashCase from './dash-case';
+import getAllKeys from './get-all-keys';
+
+const CACHED_PROP_DEFS:string = '____skate_propDefs';
```
bradleyayers: Is it necessary to specify `:string` here, is it not inferred from the assignment?

andreawyss: Yes it is not necessary. It is already inferred. 
It just look better if you are used to a typed language. 
It will not make any difference on the generated code since all type annotations are stripped during the build by babel-plugin-transform-flow-strip-types.

bradleyayers: I'd recommend pursuing TypeScript instead of Flow, one of the reasons being these kind of type inferences are made automatically.

andreawyss: And I'm only using Flow internally to help me refactor the code.
For skatejs right now Flow is a better choice.
The syntax is very similar. If you know TypeScript you can learn Flow in a few days.
I used TypeScript for 2 years but now for functional programming I prefer Flow. No extra compiler needed, just annotate your JS or ES6 code. It is closer to the platform.

Hotell: I'm sorry, this comment will be maybe off topic, but I have to react on this one :D

> For skatejs right now Flow is a better choice.

Who is judge and jury of that? you @andreawyss ? :)

> I used TypeScript for 2 years but now for functional programming I prefer Flow.

Why a sudden change of heart? Typescript is super powerful for functional stuff, maybe you didn't understood something ? ( TS is not strictly OOP at all )

> No extra compiler needed, just annotate your JS or ES6 code. It is closer to the platform.

No Extra compiler needed? Typescript is both compiler and type checker, so you don't need babel anymore. 
What is closer to the platform? With Flow you are not writing valid javascript ( by .js extension ), Flow is not even written with the platform language :)

Also Typescript: 
- runs flawlessly in hybrid mode ( js, ts files + no babel needed )
- has very strong and powerful type inference
- mature and large ecosystem and works with all IDE's and Editors
- it gives library consumers extra benefit by providing up to date typings for free ( no more manualy crating `.d.ts` files

Don't get me wrong, I say better any type checker than none, but probably the SkateJs team is in charge to properly consider what's the best for them.

andreawyss: I apologize.
I agree that the SkateJs team is in charge to properly consider what's the best for them. 
I really don't care as long as we have a type checker in place.
I assumed, maybe wrongly, that Flow was an easier and less disrupting choice for the Team because SkateJs is currently built using Babel.

TypeScript adds a level of abstractions. Flow only adds annotations.

andreawyss:
I will submit smaller PRs for each that are more focused

Issue #918, opened 2016-11-16
Add FlowType and Comments to the code


andreawyss:
Adding FlowType and Comments to the code will it make it easier for other to understand the code and contribute to it.

Hotell:
honestly I would prefer migration to Typescript. Mature ecosystem etc. But definitely at least adding JSDoc would make the code more understandable

treshugart:
I'd prefer to stay in JS proper, but don't oppose definitions for those JS
supersets. Easier to maintain in the long run. Imagine if Skate was written
in CoffeeScript, for example.

I'm on the fence about JSDoc. I feel that if the code isn't self
documenting, a redactor can help that. Doc comments become out of sync very
easily and I feel should only be used as a last resort. Clean Code has some
good info about this.

I'm definitely open to PRs to address readability and flowtype support.

Hotell:
You right about JSDoc and stuff. About Typescript that is whole different story. I don't know why people tend to think that Typescript is different language. It's just javascript, but ofc it's your call in this one :) 

P.S.: also gradual migration by hybrid .js/.ts files is super simple with TS

pyros2097:
Yeah I generally prefer typescript and all my projects are written in typescript and that is mainly for safety and those random undefined errors when you forget the return the promise in js.
But one good thing about js is that I can directly import the code into my html file and test it out. I don't think I even need webpack in this case and everything works out of the box and very easy to get started right away. And maybe in the way future we won't need anymore webpacks, module bundlers and babel and react and anyone can just start coding apps directly.

treshugart:
It feels like TypeScript is to Flow as Webpack is to Gulp. It's definitely something I really like about TypeScript: less build config. I'm open to a (possibly gradual) rewrite to Flow / TS. I don't have a preference right now, but definitely see a need for type checking as the project team grows. My biggest question is can we support both no matter which one we choose without much, if any, extra effort?

pyros2097:
[...]

PR #931, opened 2016-11-22
#931 - Use Flow for Development

andreawyss:
This PR is to propose that we use Flow to help the team be more productive and make the code more expressive for others to understand and start to contribute more.
This is mostly for the benefit to the developers that maintain the SkateJs library and at least at this point, it is NOT for the users of the library.

One alternative could be TypeScript.
The choice is really between ES6+Flow   vs.   TypeScript 

TypeScript adds a level of abstraction. 
Flow only adds annotations to the existing ES6 code.

Syntactically they both looks quite similar.

I vote for Flow, because I think it is beneficial for SkateJs to stay with ES6 + Babel.
It is best to stay closer to the platform when dealing with buggy Polyfills or buggy browser native implementations.

I let the Team vote…

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,14 @@
+[ignore]
+.*/node\_modules/esformatter/test/.*
+.*/node\_modules/config\-chain/test/.*
+.*/node\_modules/flow\-remove\-types/test/.*
+.*/node\_modules/npm/test/.*
+.*/node\_modules/npmconf/test/.*
+
+[include]
+
+[libs]
+flow/
+
+[options]
+unsafe.enable_getters_and_setters=true
--- a/flow/interfaces.js
+++ b/flow/interfaces.js
@@ -1,0 +1,26 @@
+declare interface IGetData {
+  name:string|Symbol;
+  internalValue:any;
+}
+
+declare interface ISetData {
+  name:string|Symbol;
+  newValue:any;
+  oldValue:any;
+}
+
+/**
+ * Property Configuration
+ * The options that can be used when configuring a property with skate
+ * Note: every member of this interface is optional!
+ */
+declare interface IPropConfig {
+  attribute?: boolean|string;
+  coerce?: (v:any) => any;
+  default?: any; //|(elem:any, data:any) => any;
+  deserialize?: (v:?string) => any;
+  get?: (elem:any, data:IGetData) => any;
+  initial?: any; //|(elem:any, data:any) => any;
+  serialize?: (v:any) => ?string;
+  set?: (elem:any, data:ISetData) => void;
+}
--- a/package.json
+++ b/package.json
@@ -29,9 +29,10 @@
   },
   "devDependencies": {
     "birdpoo": "0.x",
+    "flow": "0.2.3",
     "react": "15.4.0",
     "react-dom": "15.4.0",
-    "skatejs-build": "12.x",
+    "skatejs-build": "12.1.x",
     "skatejs-web-components": "5.x"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -28,11 +28,12 @@
     "window-or-global": "1.0.1"
   },
   "devDependencies": {
+    "babel-plugin-transform-flow-strip-types": "^6.18.0",
     "birdpoo": "0.x",
     "flow": "0.2.3",
     "react": "15.4.0",
     "react-dom": "15.4.0",
-    "skatejs-build": "12.1.x",
+    "skatejs-build": "12.1.0",
     "skatejs-web-components": "5.x"
   },
   "scripts": {
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,15 +1,14 @@
-const rollupBabel = require('rollup-plugin-babel');
-const presetEs2015 = require('babel-preset-es2015-rollup');
-
-const babel = rollupBabel({
-  presets: presetEs2015,
-  plugins: ['transform-flow-strip-types']
-});
-
 module.exports = require('skatejs-build/rollup.config');
 module.exports.globals = {
   'incremental-dom': 'IncrementalDOM',
   'window-or-global': 'windowOrGlobal'
 };
+
+// This should be part of the default skatejs/build
+const rollupBabel = require('rollup-plugin-babel');
+const presetEs2015 = require('babel-preset-es2015-rollup');
+const babel = rollupBabel({
+  presets: presetEs2015,
+  plugins: ['transform-flow-strip-types']
+});
 module.exports.plugins[0] = babel;
-console.log('>>> rollup', babel);
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,7 +1,15 @@
+const rollupBabel = require('rollup-plugin-babel');
+const presetEs2015 = require('babel-preset-es2015-rollup');
+
+const babel = rollupBabel({
+  presets: presetEs2015,
+  plugins: ['transform-flow-strip-types']
+});
+
 module.exports = require('skatejs-build/rollup.config');
 module.exports.globals = {
   'incremental-dom': 'IncrementalDOM',
   'window-or-global': 'windowOrGlobal'
 };
-
-console.log('>>> rollup', module.exports.plugins[0]);
+module.exports.plugins[0] = babel;
+console.log('>>> rollup', babel);
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -3,3 +3,5 @@ module.exports.globals = {
   'incremental-dom': 'IncrementalDOM',
   'window-or-global': 'windowOrGlobal'
 };
+
+console.log('>>> rollup', module.exports.plugins[0]);
--- a/src/api/prop.js
+++ b/src/api/prop.js
@@ -1,38 +1,46 @@
+//@flow
 import assign from '../util/assign';
 import empty from '../util/empty';
 
 const alwaysUndefinedIfNotANumberOrNumber = val => (isNaN(val) ? undefined : Number(val));
 const alwaysUndefinedIfEmptyOrString = val => (empty(val) ? undefined : String(val));
 
-export function create (def) {
-  return (...args) => {
-    args.unshift({}, def);
+/**
+ * Returns a function to create Property Configurations based on the
+ * given template Property Configuration that provides some default options.
+ */
+export function create(template:IPropConfig) {
+  //todo: where this is called will more than on argument?
+  return (...args:IPropConfig[]) => {
+    args.unshift({}, template);
     return assign(...args);
   };
 }
 
-export const array = create({
+export const array:IPropConfig = create({
   coerce: val => (Array.isArray(val) ? val : [val]),
   default: () => [],
-  deserialize: JSON.parse,
+  // Note: JSON.parse(undefined) throws Error
+  // this was detected by flow
+  deserialize: val => (val === undefined ? undefined : JSON.parse(String(val))),
   serialize: JSON.stringify
 });
 
-export const boolean = create({
+export const boolean:IPropConfig = create({
   coerce: value => !!value,
   default: false,
   deserialize: value => !(value === null),
   serialize: value => (value ? '' : undefined)
 });
 
-export const number = create({
+export const number:IPropConfig = create({
   default: 0,
   coerce: alwaysUndefinedIfNotANumberOrNumber,
   deserialize: alwaysUndefinedIfNotANumberOrNumber,
   serialize: alwaysUndefinedIfNotANumberOrNumber
 });
 
-export const string = create({
+export const string:IPropConfig = create({
   default: '',
   coerce: alwaysUndefinedIfEmptyOrString,
   deserialize: alwaysUndefinedIfEmptyOrString,
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,4 +1,4 @@
 module.exports = require('skatejs-build/webpack.config');
 console.log('>>> webpack plugins before', module.exports.module.loaders[2].query.plugins);
-module.exports.module.loaders[2].query.plugins = ['transform-flow-strip-types', 'transform-class-properties'];
+module.exports.module.loaders[2].query.plugins = ['transform-class-properties', 'transform-flow-strip-types'];
 console.log('>>> webpack plugins after', module.exports.module.loaders[2].query.plugins);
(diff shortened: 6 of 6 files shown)


DanielRosenwasser:
I work on the TypeScript team, and for what it's worth, TypeScript is also mostly just ECMAScript with types. In fact, ever since the iteration speed of TC39 picked up, our focus has been on the type system and tooling aspects, rather than on creating any new constructs. We heavily strive to align with ES.

I'll also mention that TypeScript tends to be a lot more flexible in the type system in terms of what you can do (e.g. being able to reopen existing types, function overload signatures, type-checked tagged templates, more granularity for nullable types, etc.).

I think regardless of what your team picks, I'd appreciate getting an idea of what affects your decision. I'd also totally be open to answering any questions you have about TS.

treshugart:
> I think regardless of what your team picks, I'd appreciate getting an idea of what affects your decision. I'd also totally be open to answering any questions you have about TS.

Thanks @DanielRosenwasser, we definitely might have to take you up on that offer :)

andreawyss:
@DanielRosenwasser In September we could not use TypeScript for Web Component development because of this issue:  https://github.com/Microsoft/TypeScript/issues/7574 (now is resolved).
Before ES6 became widely used TypeScript was a life saver for me. I still like it a lot!
But with a big community for Babel plugins and because ES6 will eventually run in future browsers I'm favoring ES6. 
Also with Webpack one can set up a project where some files are ES6 and other are TypeScript but working that way is not simple because one can only use "import" within the same language and then one must use "require" when crossing languages.

DanielRosenwasser:
Thanks for the response @andreawyss!

I actually remember you from your comment on https://github.com/Microsoft/TypeScript/pull/10762 now! I definitely get that the `super()` call issue was a problem for Custom Elements. I can say that 2.1 will be out very soon.

I also agree that the Babel plugin ecosystem has grown pretty nicely - in the 2.2 timeframe, we're looking into opening our new emit pipeline for custom transforms to enable the same sort of development.

> because ES6 will eventually run in future browsers I'm favoring ES6.

But as soon as you pick either TypeScript or Flow, this isn't necessarily true. As soon as you have a type annotation, a type declaration, etc., that needs to get stripped out of the file in some respect. In Babel, this is through a preset, and in TypeScript this is using the `ESNext` target. 

I don't think this is a terrible thing though. With either approach, you can opt out. I know that with TS, we've always worked hard to ensure that people never felt locked in. And with either technology, you will get some benefits. Our teams are working hard to demonstrate the value of typed JS in the ecosystem.

> Also with Webpack one can set up a project where some files are ES6 and other are TypeScript but working that way is not simple because one can only use "import" within the same language and then one must use "require" when crossing languages.

If I understand you correctly, what you mean is that TypeScript won't understand an import to a `.js` file. 

You can actually fix this by using the `--allowJs` flag which allows you to mix TS & JS code. You'll still get the experience you expect in TS files, and if your input JS files use [JSDoc annotations](https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript), TypeScript can even give you accurate type info, completions, and errors based on that.

If you've used VS Code with JavaScript files, you've already seen a bit of what that's like, since the experience is powered by TypeScript itself.

cdata:
These are some TypeScript features that are not *just* type annotations on type of ES6, are not part of the latest accepted ECMAScript standard, and/or are in some way specialized in TypeScript:

 - ~~Modules~~ Namespaces
 - Decorators
 - Class fields (including use of `private` keyword)
 - Mixins

This list is possibly incomplete or misinformed (I'll amend it with any offered corrections). I don't know much about Flow so I can't offer comparable list for that language.

DanielRosenwasser:
That is true, but decorators & class fields are both [stage 2 ECMAScript proposals right now](https://github.com/tc39/proposals/tree/c449e9b0a1a98c4f334d843f244715100f4806de), and class fields are required to use classes in either type system. Accessibility modifiers (e.g. `public`, `private`) are simply a type-level thing that have no emit impact, and you can feel free to not use them.

While namespaces are mostly subsumed by ES modules, namespaces also act as containers of types and can be used to describe some complex patterns. You might not use it in a new project, but they're *incredibly* useful for describing existing JS libraries.

I'm not sure what you're referring to with respect to mixins. We have intersection types but those are just type annotations.

vvakame:
👍 for TS. I ❤️ TypeScript.

cdata:
@DanielRosenwasser 

[Private class fields](https://github.com/tc39/proposal-private-fields#a-brief-introduction) are currently proposed to use a different syntax.

I'm not 100% sure about the finer details, but I understand that decorators in TypeScript are not implemented as described in the current TC39 draft.

> I'm not sure what you're referring to with respect to mixins. We have intersection types but those are just type annotations.

https://www.typescriptlang.org/docs/handbook/mixins.html

cdata:
Also, I completely understand the rationale for namespaces. But, let's be fair: ECMAScript it ain't.

justinfagnani:
@cdata The examples on the mixins page are just common JS idioms and how to type them in TypeScript.

DanielRosenwasser:
Thanks for the input @cdata. I'll try to address your points.

The private fields proposal doesn't impact our implementation of accessibility. One is design-time, the other is run-time. It's not ideal conceptually, but they can coexist, and if you're aiming for readable ES5 emit and care mainly about intent by consumers, TypeScript `private`s are probably more useful.

If you're referencing the ability to implement an interface, that's entirely a type-level thing that's helpful for enforcing the shape of a class. It doesn't have any emit impact at all, and is actually something Flow doesn't currently support.

You're right that the decorators implementation is based on an older proposal, but this isn't a problem unique to TypeScript anyway. Luckily, decorator authors can write them in a way that can feature-detect and avoid imposing breaking changes on consumers.

cdata:
@justinfagnani Thanks for the clarification. I've amended my comment.

cdata:
@DanielRosenwasser 

Thanks for the helpful info related to these TypeScript features. It's valuable to understand the areas where TypeScript is not just ECMAScript with types.

vvakame:
I submitted TypeScript edition on #935

andreawyss:
I have been using TypeScript for more then 2 years and I really like it a lot.
Before that I spent about 10 years in ActionScript and TypeScript is like a superset of ActionScript.
Now for the last 3 month I took a deep dive into ES6 + Flow and I discovered a new development experience.

The syntax is very similar and one can achieve about the same results with both. (I miss "private" from TypeScript)
The real difference relies in the type of potential issues/warnings that are reported by the 2 code checkers.
Flow is really more than a type checker; It is a code linter on steroids!

E.g. in the sample code on this PR Flow found this problem:

In the IPropConfig interface the optional deserialize method is declared as:
     deserialize?: (v:?string) => any;

in the skate props.array strategy it was implemented as:
     deserialize: JSON.parse,

Flow gave me an error because: JSON.parse(undefined) throws an error!
Therefore, I had to make the following change to satisfy Flow:
     deserialize: val => (val === undefined ? undefined : JSON.parse(String(val))),

TypeScript on the other hand didn’t report this kind of issue. 
I’m sure that one can find scenarios where the opposite is true.
During these 3 month with Flow, I have been pleasantly surprised at the kind of potential runtime issue that it warns me about. It is very good at detecting potential problems with undefined and null values.

DanielRosenwasser:
Yes, I think that you'll also find situations where only TS catches certain errors. A pretty easy example is with [something like this](https://flowtype.org/try/#0PTAEAEDMBsHsHcBQiCmAPADrATgF1JAK4B2AxrgJazEGywAUaAXKAM67YXEDmANKAE8WAIzrQUAQ2IBKUAG9kAX2QA3CdlATQAXnmgAdIch1QigNyIJ+iawAm1uw9v1pZoA), whereas 2.1 will report such an issue.

Have either of you run into situations where Flow is more lax than you'd anticipated? How do you typically ensure that it's not overlooking something?

Hotell:
this is not probably related but maybe @DanielRosenwasser can help brainstorm on this on #936

andreawyss:
Yes, I find cases where Flow is more relaxed then TypeScript. 

I just submitted PR #938 for a 1-to-1 comparison with this PR #931:
#938 only converts one file (/src/api/prop.js) to prop.ts.
#931(this one) only adds Flow annotations to that same file.

One can see that TypeScript detected a different issue on the same file.

In the IPropConfig interface the optional serialize method is declared as:
TypeScript: serialize?: (v:any) => string|null|undefined;
Flow: serialize?: (v:any) => ?string;

in the skate prop.number strategy it was implemented as:
serialize: alwaysUndefinedIfNotANumberOrNumber

I had to make the following change to satisfy TypeScript:
serialize: alwaysUndefinedIfNotANumberOrString

In this case this issue would not have cause a runtime error since what gets returned from serialize is then used and automatically converted to a string in element.setAttribute( ‘name’, serializedValue );
On the other hand, JSON.parse(undefined) will definitely throw a runtime error.

andreawyss:
@DanielRosenwasser 
For most libraries I would not have a strong preference between TypeScript vs. ES+Flow+Babel. 
As long as a type checker and linter are in place the team will be more productive. 

The reasons that for the SkateJs library I’m voting for ES+Flow+Babel are:
1) SkateJs is already currently using Babel and some ES features.
2) I don't want to add an abstraction(TS) but just decorate(Flow) the existing code.
3) SkateJs must handle early implementations of WebComponents from both Polyfills and Browsers(native). To do this sometimes one needs to write them with low level JS Object manipulations.

For SkateJs it is beneficial to stay in JS proper and have the most control on how some ES or TS code gets transpiled.
Currently it is easier to achieve this in the Babel ecosystem than not in TypeScript ecosystem.
With TypeScript 2.2 this will slowly change.
I also agree that it is possible to keep some files in JS and not convert everything to TypeScript. The dev environment gets a bit more complex with separate checker, linters that may not be aware of each other codebase without some extra work (defining some types both in TS and Flow).

When I first submitted issue #918 - Add FlowType and Comments to the code
@Treshugart stated:
_I'd prefer to stay in JS proper, but don't oppose definitions for those JS supersets. Easier to maintain in the long run. Imagine if Skate was written in CoffeeScript, for example. 
…_

I really value @Treshugart and @justinfagnani opinions because they have a lot of experience working with early implementations of WebComponents from both Polyfills and Browsers(native).

justinfagnani:
@andreawyss what version of TypeScript were you using with that JSON.parse() warning? TypeScript 2.0 with strict null checks would find that error as well. I run TS with strictNullChecks, noImplicitAny, noUnusedLocals, and noUnusedParameters. It catches a lot of things.

I don't really view TypeScript as adding any abstractions, so I wonder what you're referring to. In general, TS doesn't add any runtime semantics, with the possible exception of class properties which have a high degree of alignment with the ECMA proposal, and decorators which are experimental and have to be explicitly enabled. I only use JS modules and imports and don't explicitly use namespaces, except where I come across them in third party type declarations.

TypeScript's type system is by now insanely powerful and suited to describing JS. It's recently added data-flow sensitive type checking, mapped types, the keyof operator, and custom elements APIs to it's DOM lib, which allows us to do things like this:

```typescript
class MySlider extends HTMLElement {
  slide() {}
}
customElements.define('my-slider', MySlider);

// Add 'my-slider' globally to the list of valid tag names and associate it with the right type
interface HTMLElementTagNameMap {
  'my-slider': MySlider;
}

document.createElement('my-slider').slide() // this typechecks!
document. getElementsByTagName('my-slider')[0].slide() // so does this!
```

Also, even though decorators are experimental, they add enough value for ergonomics, IMO, that we're going to ship some TS decorators for Polymer 2.0: https://github.com/Polymer/polymer/pull/3954

They're completely optional, and you could implement them so your JS interface doesn't change at all, but users could do this:

```typescript
import {skate, attribute} from '../skate/decorators.js';

@define('x-hello')
class HelloElement extends skate.Component {

  @attribute name: string;

  renderCallback () {
    return skate.h('div', `Hello, ${this.name}`);
  }
}
```

(some may not like that, but I do :) )

My experience with @DanielRosenwasser and @mhegazy has been awesome and they've been extremely responsive with adding features we've needed for awesome custom elements support, including https://github.com/Microsoft/TSJS-lib-generator/pull/150 and https://github.com/Microsoft/TypeScript/issues/7574

I really can't recommend it highly enough.

treshugart:
@andreawyss 

> SkateJs must handle early implementations of WebComponents from both Polyfills and Browsers(native).

v0 support is long gone :) and the original polyfills had issues, so were never supported.

I'm inclined to lean towards TypeScript. I like both, but I do this for a few reasons, unrelated to the type checking details and differences:

- It will simplify the build
- It *mostly* aligns to the ECMAScript standard following stage 2 (?) TC39 specs
- We don't have to use things that deviate from the specs (but we should definitely flag them)
- Custom element type checking; that is awesome
- It feels like it will speed up those using both TypeScript (more) or standard JS (less)
- This is a big one: TS is far more mature.

The things that worry me are:

- Dependencies that don't use TS
- We are trusting TS not to deviate to far from ES, seems there's a few gotchas currently
- I've seen issues with Skate JSX support, is this solvable? Maybe it is and I missed it.

Whatever happens out of this, there will need to be a blog post :p

andreawyss:
Thank you @treshugart, @justinfagnani and @DanielRosenwasser. 
It looks like most team members are concurring with moving forward with TypeScript.
Therefore, I close this PR so we can start working on a version of SkateJs written in TypeScript.

Hotell:
@treshugart 
> I've seen issues with Skate JSX support, is this solvable? Maybe it is and I missed it.

what issues do you mean?

If skate could export just `React` alias it would be even more easier.
Currently if one wants to use jsx he has to manually do `const React = { h: createElement }`

treshugart:
@Hotell sounds like TypeScript needs to change here, not Skate. cc @DanielRosenwasser

DanielRosenwasser:
@treshugart I'm not specifically familiar with what you mean, but in 2.1 we'll support the `jsxFactory` flag (https://github.com/Microsoft/TypeScript/issues/9582). If this doesn't cover your use-case and there isn't already an issue on us, feel free to file one on us so we can discuss it a little bit more. Our team has tried to make our JSX support flexible enough and sufficiently decoupled from React so that it can be used with multiple libraries.

vvakame:
I made a usecase for SkateJS 4.1.0 https://github.com/vvakame/til/tree/master/skatejs-newdef

> I've seen issues with Skate JSX support, is this solvable? Maybe it is and I missed it.

Is this mean https://github.com/Microsoft/TypeScript/issues/7004 ?
There are no signs to solve it at the moment.
We need support from @DanielRosenwasser I think 😉 

now, We can detour that issue. not cool.
```
    renderCallback() {
        const anyProps: any = {};
        return (
            <div>
                <CountUpComponent count={100} {...anyProps}></CountUpComponent>
            </div>
        );
    }
```

vascofernandes:
Just want to pitch in and cast my vote for Typescript!

Issue #919, opened 2016-11-16
"static get props" should be called only once per class 


andreawyss:
static get props returns props definitions that are constructed using Prop.create()
static get props is accessed multiple times during component livecycle.
Each time static get props is accessed a new set of props definitions get reconstructed!

To solve this issue, create a static getPropDefs() that caches the props definitions for a given class and then refactor the code to use static getPropDefs() instead of directly accessing static get props.

implementation would be something like this:

static getPropDefs() {
    if (!this.hasOwnProperty('_propDefs') ) {
        this._propDefs = this.props;
    }
    return this._propDefs;
}

Note: getPropDefs is static therefore "this" refers to the Constructor and this.props if not found on the current constructor will be looked up in an ancestor Constructor that the component is extending.

treshugart:
That's a great idea. PRs welcome :)

andreawyss:
This will be solved once PR #932 will be approved and merged.

andreawyss:
This is resolved.

PR #932, opened 2016-11-23
#932 Cache result of 'static get props'

andreawyss:
cache prop configs on the constructor   issue #919


andreawyss:
@treshugart please review.
I'm not sure why these 3 test fail with my props caching changes.
Thank you 

Firefox 50.0.0 (Mac OS X 10.11.0) lifecycle/property api set() for elements initialised from the DOM is called for an attribute prop when initialised as an attribute FAILED
        AssertionError: expected 0 to equal 1 (:0)

Firefox 50.0.0 (Mac OS X 10.11.0) lifecycle/property api set() for elements initialised from the DOM is called for a non-attribute prop when set as a property before it upgrades (polyfill) FAILED
        AssertionError: expected 0 to equal 1 (:0)

Firefox 50.0.0 (Mac OS X 10.11.0) lifecycle/property api set() for elements initialised from the DOM is called for a non-attribute prop when set as a property before it upgrades (native) FAILED
        AssertionError: expected 0 to equal 1 (:0)

src/api/component.js
```diff
@@ -2,12 +2,15 @@ import { patchInner } from 'incremental-dom';
 import {
   connected as $connected,
   created as $created,
+  ctor_props as $ctor_props,
```
treshugart: nit: it's preferable to not have underscores in `NON_UPPERCASE_CONSTANT` names. If you can `camelCase` it, that'd be awesome.

src/api/component.js
```diff
@@ -125,11 +128,12 @@ export default class extends HTMLElement {
...
   }
 
+  // returns inherited super props overwritten with this Component props
   static get props () {
-    return {};
+    return assign({}, super.props, this[$ctor_props]);
   }
   static set props (value) {
-    Object.defineProperty(this, 'props', { configurable: true, value });
+    this[$ctor_props] = value;
```
treshugart: I was getting stack size errors in IE11 if I didn't use `Object.defineProperty()`.

src/api/component.js
```diff
@@ -142,19 +146,17 @@ export default class extends HTMLElement {
 
     // TODO refactor to not cater to Safari < 10. This means we can depend on
     // built-in property descriptors.
-
-    // todo: createInitProps Must be defined on the constructor and not on an inherited Component
-    // if (!constructor.hasOwnProperty($props)) {
-    if (!constructor[$props]) {
-      constructor[$props] = createInitProps(constructor);
+    // Note: createInitProps must exist on the constructor and not from an inherited Component
+    if (!constructor.hasOwnProperty($ctor_createInitProps)) {
```
treshugart: nit: Probably don't need to call `hasOwnProperty()` if using a symbol.

src/util/cached-prop-configs.js
```diff
@@ -10,29 +11,29 @@ const PROP_CONFIGS_COUNT = '____skate_propConfigsCount';
...
-    Ctor[CACHED_PROP_CONFIGS] = getAllKeys(propConfigs).reduce((result, propName) => {
+    Ctor[$ctor_propConfigs] = getAllKeys(propConfigs).reduce((result, propName) => {
       count++;
       result[propName] = propConfigs[propName];
       return result;
     }, {});
 
-    // console.log('create', CACHED_PROP_CONFIGS, 'on constructor', Ctor.name, Ctor[CACHED_PROP_CONFIGS]);
-    Ctor[PROP_CONFIGS_COUNT] = count;
+    // console.log('created PropConfigs on', Ctor.name, Ctor[$ctor_propConfigs]);
```
treshugart: Remove

src/util/cached-prop-configs.js
```diff
@@ -10,29 +11,29 @@ const PROP_CONFIGS_COUNT = '____skate_propConfigsCount';
...
+    Ctor[$ctor_propConfigsCount] = count;
   }
   // else {
-  //   console.log('Cached PropConfigs already found on constructor', Ctor.name, Ctor[CACHED_PROP_CONFIGS]);
+  //   console.log('use PropConfigs on', Ctor.name, Ctor[$ctor_propConfigs]);
```
treshugart: Should probably remove all commented out code.

src/util/cached-prop-configs.js
```diff
@@ -0,0 +1,42 @@
...
+    Ctor[$ctor_propConfigs] = getAllKeys(propConfigs).reduce((result, propName) => {
+      count++;
+      result[propName] = propConfigs[propName];
+      return result;
+    }, {});
+
+    // console.log('created PropConfigs on', Ctor.name, Ctor[$ctor_propConfigs]);
+    Ctor[$ctor_propConfigsCount] = count;
+  }
+  // else {
```
nadiam84: nit: you probably don't need this any more?

src/util/cached-prop-configs.js
```diff
@@ -0,0 +1,42 @@
...
+  //   console.log('use PropConfigs on', Ctor.name, Ctor[$ctor_propConfigs]);
+  // }
+
+  return Ctor[$ctor_propConfigs];
+}
+
+export function getPropConfigsCount (Ctor) {
```
nadiam84: This function worries me a bit. You're defining a 'get' function, which is, according to its name, supposed to return a number, nothing more. But inside you're calling another function, which is, again, supposed to just return an object (according to the name). But this simple get function suddenly and secretly changes the object that was passed into it.

Doesn't seem like a good practice.

andreawyss:
@nadiam84 I did cleanup the code. Please review.

andreawyss:
@treshugart I have added a test

src/util/cached-prop-configs.js
andreawyss: I cleanup this with my latest changes.

src/api/component.js
andreawyss: I fix this in my latest changes

src/util/cached-prop-configs.js
andreawyss: Done

src/api/component.js
andreawyss: $ctorCreateInitProps is not really a symbol but just a string.

src/util/cached-prop-configs.js
andreawyss: Done

src/util/symbols.js
```diff
@@ -9,6 +9,13 @@ export const created = '____skate_created';
 // be passed to vdom functions as tag names.
 export const name = '____skate_name';
 
+// Used on the Constructor
```
treshugart: I wonder if these would be better put into where they are used. I've been getting confused lately because I'm not sure where in the codebase a symbol is used, or usage originated. No need to change right now as this is following this convention, but worth thinking about.

test/unit/api/component.js
```diff
@@ -31,6 +31,25 @@ describe('api/Component', () => {
         test: { attribute: true }
       });
     });
+
+    it('static props is called once', () => {
```
treshugart: Maybe "static props should be memoized"? I don't feel too strongly about this.

andreawyss:
@treshugart thank you for inviting me to join the skatejs organization.
Do you usually 'merge' or 'squash and merge' ?

treshugart:
No problem. You're having some great input and contributions, and also figured this would be simpler moving forward :).

I generally just merge because that's the default which makes clicking the button easier :p. I've read several things about each one's pros and cons. Merging makes your history harder to follow, but seems like it gives you the most flexibility.

PR #935, opened 2016-11-23
Try TypeScript

vvakame:
related #931

I tried to support TypeScript.
This is minimal implementation.
TypeScript has various super cool options, likes `--noImplicitAny`, `--strictNullChecks` and others.
But I made this change as small as possible.

Don't publish this code.
TypeScript users (it's me!) can't solved incremental-dom and window-or-global package definitions.
We should publish these definitions to [@types](https://www.npmjs.com/~types) and do `yarn add @types/incremental-dom @types/window-or-global`. (sure, I can do that!)

PR #938, opened 2016-11-23
#938 - Use TypeScript for Development

andreawyss:
This PR is to compare 1-to-1 converting one file at the time to TypeScript of this project vs. (PR #931) adding Flow annotations one file at the time.

Like in PR #931 we only added Flow types annotation //@flow to one file: /src/api/prop.js

Here also in this PR we only converted to TypeScript the same file: src/api/prop.js => prop.ts


andreawyss:
This branch compiles,  but tests fail because karma.config needs to be updated to handle ts files.
Can someone fix this. Thank You!

treshugart:
Sure. If you want to have a crack, you can override the karma config that is pulled in from skatejs-build in Skate's karma.config.js file. Is all that needs to happen is to just compile TS?

andreawyss:
@vvakame Masahiro can you please, help get the karma unit tests to run on my branch:
https://github.com/andreawyss/skatejs/tree/use-ts-for-dev
Thank You

vvakame:
sorry, I don't know the details of typings, webpack and rollup.
https://github.com/andreawyss/skatejs/blob/use-ts-for-dev/tsconfig.json#L9 is this right?
I tried to exec `./node_modules/.bin/tsc -p ./`, it succeeded.

andreawyss:
Ok, I fixed the tests by overriding the karma config.
I only needed to include the 'karma-typescript-preprocessor'

In PR #935 - Try Typescript, all src files were renamed from .js to .ts
This PR shows one other possible way of moving incrementally to TS a few files at the time.
In this case only **prop.js** was converted to **prop.ts**

This may minimize merge conflicts with other branches.
One has the option to keep, if needed, some files in ES a bit longer. 
Useful if encountering an issue using TS in particular area of the code.

karma.conf.js
```diff
@@ -1,13 +1,120 @@
...
+    // list of files / patterns to load in the browser
+    files,
+
+    // list of files to exclude
+    exclude: [],
+
+    // list of preprocessors
+    preprocessors,
+
+    typescriptPreprocessor: {
```
treshugart: Can't we reuse the base config from skatejs-build still and just set `webpack` to `null` and set teh `typescriptPreprocessor to this?` I'm probably missing a couple things, but it would seem we can reuse quite a bit still.

treshugart:
Can you change the destination branch to the "typescript" branch, please? Unfortunately GitHub doesn't allow you to do this, so you'll have to resubmit the PR. Sorry :(

andreawyss:
@treshugart @vvakame @Hotell
We are currently working on two separate TypeScript efforts.
1) Quickly fix a few issues with the current index.d.ts so that TS user can start to enjoy our latest version 4.1.
2) Incrementally refactor skatejs code into TypeScript. Once this second one will be completed then d.ts will be automatically generated by the TS compiler. And at that point we will no longer need 1). 

This two efforts are related but cannot coexist inside the same branch.
1) can be done in a few days 
2) may take a while to be complete since for some area of the code we may need to wait for TS 2.2 to be stable.


there are realy #940

vvakame:
We can do 2 ways both.
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

1st. we do `1.` course. add hand made `index.d.ts` and add `types` property to package.json.
2nd. we do `2.` course. have fun!
3rd. 2nd is done & next TypeScript (2.1.3? 2.2.0?) released. We should `rm index.d.ts` and change `types` property. use tsc generated definitions.

andreawyss:
Ok we can do 1) first. 
But before that we need to merge master into typescript branch.

andreawyss:
@DanielRosenwasser this is an other issue with TS that is still open:
https://github.com/Microsoft/TypeScript/issues/1863

@treshugart I still feel strongly, that it would have been less effort and less disrupting to the codebase, to first only annotate the code with Flow types, and hold on for TS for a little longer.

Then at a later time just rename all files from .js to .ts, and adjust a few annotations to take in account for the annotation syntax differences between Flow and TS.

It would be an easier migration if we break this up into 2 steps.
I'm not against TS.
I just think that it will be much easier to get there via a Flow intermediate step.

DanielRosenwasser:
@andreawyss supporting symbol indexing isn't a very trivial task, but it's something we've postponed because we wanted to see how symbols would be utilized by the JS community. Can you tell me a little bit more about how symbols are being used right now?

andreawyss:
@treshugart @DanielRosenwasser
We use symbols to implement a sort of "internal/protected" like visibility for certain component's properties. The symbols are only known by a set of classes.

We configure the properties of a WebComponent using an object map where the key in the map are the property names.
Since the property names can be a string or a symbol, we have an object map that contains a combination of string or symbol keys. 
When passing around the map to a function we like to declare it this way:  
    **propsMap:{[key:string|symbol]:any}**

This will prevent someone to write code that assumes that the property name is always a string, and perform an operation that would not be valid for a symbol.

andreawyss:
I'm closing this for now since TypeScript is not yet supporting some of the features that SkateJs uses.
I will revisit this at a later time.

DanielRosenwasser:
@andreawyss @treshugart sorry for the delay, and sorry to hear that it's a bit difficult for you guys, though, right now we do actually have a way of mitigating this in the type system.

Basically, if you have a `string` index signature, you can use `symbol`s to index as well. This actually seems to fall in line with the model that you're working with (i.e. `symbol`s are only used for privates, `string`s for anything else).

Other than that, I think we'll be trying to look into ways of expressing this more clearly in the future.

PR #941, opened 2016-11-25
#940 - Typings for component has been added

### LINKED ISSUE #940: Missing typings for Component class
**Component** just extends **HTMLElement** but has no methods (callbacks) declared.

hopkins-tk:
Callbacks added to **Component** type definition.

Fixes #940

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -7,12 +7,29 @@ export function props(elem: any, data?: PropData): void;
 export const emit: (elem: any, eventName: string, eventOptions = {}) => boolean;
 
 export class Component extends HTMLElement {
+  static readonly props: { [name: string]: PropOptions };
+  static readonly observedAttributes: string[];
   updatedCallback(prevProps: { [name: string]: any }): boolean;
   renderCallback(): any | undefined;
   renderedCallback(): void;
   connectedCallback(): void;
   disconnectedCallback(): void;
   attributeChangedCallback(name: string, oldValue: any, newValue: any): void;
+
+  // DEPRECATED
+  static created?(elem: Component): void;
+  // DEPRECATED
+  static attached?(elem: Component): void;
+  // DEPRECATED
+  static detached?(elem: Component): void;
+  // DEPRECATED
+  static attributeChanged?(elem: Component, data: { name: string, oldValue: any, newValue: any }): void;
+  // DEPRECATED
+  static updated(elem: Component, prevProps: { [name: string]: any }): boolean;
+  // DEPRECATED
+  static render?(elem: Component): any | undefined;
+  // DEPRECATED
+  static rendered(elem: Component): void;
 }
 
 export function define<C>(name: string, component: C): C;
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -7,12 +7,12 @@ export function props(elem: any, data?: PropData): void;
 export const emit: (elem: any, eventName: string, eventOptions = {}) => boolean;
 
 export class Component extends HTMLElement {
-  updatedCallback?(prevProps: { [name: string]: any }): boolean;
-  renderCallback?(): void;
-  renderedCallback?(): void;
-  connectedCallback?(): void;
-  disconnectedCallback?(): void;
-  attributeChangedCallback?(name: string, oldValue: any, newValue: any): void;
+  updatedCallback(prevProps: { [name: string]: any }): boolean;
+  renderCallback(): void;
+  renderedCallback(): void;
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -6,7 +6,14 @@ export function props(elem: any, data?: PropData): void;
 
 export const emit: (elem: any, eventName: string, eventOptions = {}) => boolean;
 
-export class Component extends HTMLElement {}
+export class Component extends HTMLElement {
+  updatedCallback?(prevProps: { [name: string]: any }): boolean;
```
Hotell: why do we need those methods as optional?

andreawyss: Because the implementer of the Component is not required to provide one.

Hotell: that is not true @andreawyss, if it was interface and user would use `implements`, then it would errored, not with this one.

By having it on class definition, user is not forced to use that, it just provides better intellisense  within editor.

src/index.d.ts
```diff
@@ -6,7 +6,14 @@ export function props(elem: any, data?: PropData): void;
...
+  renderCallback?(): void;
```
Hotell: this is wrong definition. `renderCallback` has to return `null` or `JSX.Element` or `vdom`

src/index.d.ts
andreawyss: @Hotell Yes, you are correct.

src/index.d.ts
hopkins-tk: I had glanced to the tests and there are even arrays being returned for this callback, so I returned the generic `any | undefined` that is used elsewhere as well.

So I am not sure what are all the possibilities.

src/index.d.ts
```diff
@@ -6,7 +6,14 @@ export function props(elem: any, data?: PropData): void;
 
 export const emit: (elem: any, eventName: string, eventOptions = {}) => boolean;
 
-export class Component extends HTMLElement {}
+export class Component extends HTMLElement {
```
Hotell: I would add whole API although it doesn't affect consumption, just to be consistent with implementation, until skate is migrated to TS 
```ts
class Component extends HTMLElement {
  static readonly props: { [name: string]: PropOptions };
// ....
}
```

src/index.d.ts
```diff
@@ -6,7 +6,14 @@ export function props(elem: any, data?: PropData): void;
...
+  updatedCallback(prevProps: { [name: string]: any }): boolean;
+  renderCallback(): any | undefined;
+  renderedCallback(): void;
+  connectedCallback(): void;
+  disconnectedCallback(): void;
```
andreawyss: updatedCallback(prevProps: { [name: string | symbol]: any }): boolean;

renderCallback(): any | any[] | undefined;

Hotell: there absolutely no reason to do `any | any[]`, because it merges just to `any`.

Those definitions needs to be rewritten with appropriate `JSX` namespace but that's not part of this PR

Hotell: also `renderCallback` shouldn't ever return `undefined` , the same as in preact or react. If it doest it's a programatic error

andreawyss:
Everywhere a prop name is passed should be defined as:  string | symbol

treshugart:
Can you change the destination branch to the "typescript" branch, please? Unfortunately GitHub doesn't allow you to do this, so you'll have to resubmit the PR. Sorry :(

Hotell:
> Everywhere a prop name is passed should be defined as: string | symbol

this isn't possible yet @andreawyss 
https://github.com/Microsoft/TypeScript/issues/1863

Hotell:
I would merge this as is, and continue with further type definitions tweaking, possibly with @vvakame contributions. wdyt? @treshugart @andreawyss

andreawyss:
I would just do this more:

attributeChangedCallback(name: string, oldValue: **null|string**, newValue: **null|string**): void;

static attributeChanged?(elem: Component, data: { name: string, oldValue: **null|string**, newValue: **null|string** }): void;

and then this looks good to me.

hopkins-tk:
I have changed the values types to **string** - I don't think **null** is needed, as all the fields are _nullable_.

Or do you feel that explicit **null** would be any good there?

Hotell:
if user uses `strictNullTypeChecks` we need to explicitly annotate types which should be nullable with `null` type, instead he would get error, that's why we need to be explicit

andreawyss:
I will merge this into typescript branch and then fix there before going top master

Issue #838, opened 2016-10-07
Proposal to change the way attributes / properties are reflected


treshugart:
After a recent discussion (https://twitter.com/rob_dodson/status/784152571220930561) I'm thinking the best algorithm for attribute / property reflection would be to use props as the source of truth but to have attribute sets affect properties if linked. So basically:

- Attribute set sets props both for init and over time. Attribute value is deserialised as normal.
- Property sets are never reflected back to attributes as this mutates the DOM. Things could become out of sync here but it's kinda like the state vs props thing in React.

### Pros

- Perf: props are faster than attrs (3x, I think), and syncing props to attrs incur the attr set + whatever the serialisation cost is.
- Source of truth is props.
- No DOM state is mutated.

### Cons

- Selectors (DOM or CSS) for attributes that have a linked prop, and the prop is set (but never linked back). This could always be mutated back in the property setter.
- Possible gotchas where most would think if it's linked one way, that it'd be linked both ways. As with the first con, this could be addressed.

### Possible solutions

- Two syncing styles: one-way and two-way.

jpnelson:
What is the advantage in not serialising back to the DOM? If you need to be able to deserialise it, there must be a way to serialise that data again anyway. I guess the advantage would be DOM mutation, which wasn't really discussed in that thread, but is the DOM mutation of the attributes really an issue?

treshugart:
Maybe a possible solution would be to have aggressive vs passive syncing?

robdodson:
Maybe worth noting that some things don't have corresponding properties. Specifically `aria-*` :\

treshugart:
@robdodson this is true. I think in this case one would expect to set an attribute. In the case of Skate, you could do:

```js
props: {
  ariaActiveDescendant: { attribute: true }
}
```

And that would link the attribute to the specified property. In the case of this proposal, setting the property wouldn't set the attribute. However, if you set the attribute, it would set as normal, but also sync to the property and trigger a `render()`. This can be done for `data-*` attributes as well.

I think no matter what choice anyone makes it will not be aligned with native elements, but I don't think there's anything we can do about that. I think the more important question here is what is a good pattern for Web Components, or at the very least, Skate. I think that it's very helpful though to get your insight, and others from the Polymer team about this. This feels like something that is borderline spec-able and even if it doesn't go into the spec, we might be able to align on standardised behaviour between our libraries.

marcoms:
[...]

PR #942, opened 2016-11-26
Refactor w/ PropDefinition to hold prop name and normalized prop options

andreawyss:
Using the new internal class 'PropDefinition' I have simplified a few functions.


treshugart:
It looks like you're playing around with changes related to #838 in here? What is the focus of this PR. It looks like you're moving stuff into a property definition class?

Maybe it's worth building this into a mixin where it can be used piecemeal (but also composed in to create `Component`).

andreawyss:
This PR is still WIP but I wanted you to see where I'm going.
Would you prefer that in the future I keep my WIP on my fork and just point you there?

The main focus of this work is the property definition class.
By having all the **prop options normalized into one single object that also contains the property name** it simplifies many function signatures and avoids duplicated logic to test if the attribute is a boolean or a string and dashCase the property name in more then one place.

This will then simplify the changes that will be required for #838.
In the property definition class we could also, if we decide to do it, perform options validation.

treshugart:
> Would you prefer that in the future I keep my WIP on my fork and just point you there?

Now that you have access, feel free to just create branches within the skate repo. Easier to collaborate that way. No need to PR unless it's a solid proposal.

> By having all the prop options normalized into one single object that also contains the property name it simplifies many function signatures and avoids duplicated logic to test if the attribute is a boolean or a string and dashCase the property name in more then one place.

> This will then simplify the changes that will be required for #838.
In the property definition class we could also, if we decide to do it, perform options validation.

👍 sounds good to me. It'd be nice if 1-way / 2-way attribute reflection could be split apart from this refactor.

andreawyss:
@treshugart I did remove changes related to #838
Now this PR is a pure refactor with no new feature.
Please, take a close at each **todo:** that I left in the code.
These are questions that we need to discuss and further refine the code, comments or the doc.

src/api/component.js
```diff
@@ -127,10 +105,11 @@ export default class extends HTMLElement {
   static get observedAttributes () {
     const attrsOnCtor = this.hasOwnProperty($ctorObservedAttributes) ? this[$ctorObservedAttributes] : [];
 
-    const props = getPropsMap(this);
-    const attrsFromLinkedProps = Object.keys(props).map(key => {
-      const { attribute } = props[key];
-      return attribute === true ? dashCase(key) : attribute;
+    const propDefs = getPropsMap(this);
+    // todo: with Object.keys we skip symbol properties!
```
treshugart: Good point, it *should*. The tests currently pass because we're transpiling, I believe, but it doesn't appear that it'd work in ES2015. If it takes too long for you to test and fix in this PR, maybe raise a separate issue and fix in that.

src/lifecycle/props-init.js
```diff
@@ -77,23 +81,15 @@ function createNativePropertyDefinition (name, opts) {
...
     if (this[$connected]) {
-      syncPropToAttr(this, opts, name, false);
+      syncPropToAttr(this, propDef, false);
     }
   };
 
   return prop;
 }
 
+// todo: This is only used from unit tests
```
treshugart: Can we drop this code and rewrite the tests? IIRC the tests have a single function that call out to this, so that'd be a single refactor point, probably.

src/lifecycle/props-init.js
```diff
@@ -52,21 +54,23 @@ function createNativePropertyDefinition (name, opts) {
     let { oldValue } = propData;
 
     if (empty(oldValue)) {
+      // todo: the doc is incorrect:  When the property is initialised, oldValue will always be undefined
+      // we probabbly need to update the doc
```
treshugart: (thumbsup) I don't remember why we do null here instead of `undefined`, but maybe nulling everything is better since you can redefine `undefined`.

src/util/prop-definition.js
```diff
@@ -0,0 +1,104 @@
...
+export default class PropDefinition {
+
+  // constructor(name:string|symbol, cfg:PropOptions) {
+  constructor (nameOrSymbol, cfg) {
+    this._name = nameOrSymbol;
+
+    cfg = cfg || {};
+
+    if (typeof cfg === 'function') {
+      // todo: Where is documented that a config can just be the coerce function?
```
treshugart: I don't think it is. This is an artefact of the 0.x API, I think. Happy for you to remove this in a minor as it was never documented in the major version APIs.

src/util/prop-definition.js
```diff
@@ -0,0 +1,104 @@
...
+      cfg = {coerce: cfg};
+    }
+
+    // Set Default values
+
+    this.attrName = null;
+
+    this.coerce = null;
+
+    // todo: we probabbly need to update the doc
```
treshugart: I'm not sure we need to. Retrieving metadata from props isn't really part of the public API.

src/util/prop-definition.js
```diff
@@ -0,0 +1,104 @@
...
+    // todo: should be JSON.stringify ?
+    // value was defined inside props-init.js
+    this.deserialize = value => value;
+
+    // "initial" option is truly optional and it cannot be initialized.
+    // Its presence is tested using hasOwnProperty()
+
+    this.get = null;
+
+    // todo: should be JSON.parse ?
```
treshugart: By default? No, I don't think so. That'd be unnecessarily expensive.

src/util/prop-definition.js
```diff
@@ -0,0 +1,104 @@
...
+    // from doc one would think default value is undefined
+    // value was defined inside props-init.js
+    this.default = null;
+
+    // todo: should be JSON.stringify ?
+    // value was defined inside props-init.js
+    this.deserialize = value => value;
+
+    // "initial" option is truly optional and it cannot be initialized.
+    // Its presence is tested using hasOwnProperty()
```
treshugart: Maybe using `'initial' in propDef` would be better?

src/util/prop-definition.js
```diff
@@ -0,0 +1,104 @@
...
+    this.attrName = null;
+
+    this.coerce = null;
+
+    // todo: we probabbly need to update the doc
+    // from doc one would think default value is undefined
+    // value was defined inside props-init.js
+    this.default = null;
+
+    // todo: should be JSON.stringify ?
```
treshugart: Not by default. As mentioned below, this would be unnecessarily expensive and opinionated.

src/util/prop-definition.js
```diff
@@ -0,0 +1,104 @@
...
+
+}
+
+function resolveAttrName (attrOption, nameOrSymbol) {
+  if (attrOption === true) {
+    if (typeof nameOrSymbol === 'string') {
+      return dashCase(nameOrSymbol);
+    }
+    if (typeof nameOrSymbol === 'symbol') {
+      // todo: should we even allow a symbol prop to have a linked attribute?
```
treshugart: Attributes are public API, symbols are private, so I don't think so. This would enforce a best-practice, I think.

src/util/prop-definition.js
andreawyss: ok

src/util/prop-definition.js
andreawyss: ok

src/api/component.js
andreawyss: Well since we say that symbol property cannot have linked attribute then, it is correct to only loop on properties that are not symbols.

andreawyss:
@treshugart review again all changes.
All existing tests pass. 
We really did normalized empty to null more than we thought.
Attribute cannot be undefined. DOM getAttribute() returns null.
'default' option value is null.
previous value if empty is normalized to null.
For consistency I also changed the coerce and deserialize of the prop.string to be null too.
This may be a small breaking change. But none of the existing tests failed.

andreawyss:
@treshugart this PR is now ready for you to review.
Besides a pure refactor there is a small change in how NaN works for a prop number.
Setting the prop to NaN will maintain the NaN value and serialize attribute to 'NaN' which deserializes back to NaN.
This should be a small breaking change that will not affect most common usecases of prop number.
I had to change one test and use Object.is for testing equality of NaN equal NaN.

andreawyss:
I will merge this into separate 'prop-definition' branch and then resubmit PR from that branch.

PR #952, opened 2016-12-01
rewrite TypeScript definitions

vvakame:
related: #940 #941
When this PR merged, users can use TSX under TypeScript@2.1.3 environment. (it is not released!)

* [x] polish up definitions
* [x] publish `@types/window-or-global` to npm
  * https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13040
* [x] update `@types/incremental-dom` to npm
  * [current version](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/incremental-dom/index.d.ts) doesn't have `patchInner`, `applyProp`, `attributes` ...
  * https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13041
* [ ] add definitions to `skatejs-web-components`
  * `skatejs` dependent on `skatejs-web-components`
  * https://github.com/skatejs/web-components/pull/30
  * publish & use it in this project.
* [ ] ?? publish `@types/whatwg-customelements` to npm
  * https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13045
  * https://github.com/Microsoft/TSJS-lib-generator/pull/176
* [x] ?? publish `@types/whatwg-shadowdom` to npm
  * TypeScript@2.1.3 already has a `HTMLSlotElement` and others. see https://github.com/Microsoft/TypeScript/blob/release-2.1/lib/lib.dom.d.ts#L13003-L13006
* [x] add tests
  * NOTE: https://github.com/vvakame/til/blob/49294d3185df832f93c01f5b34f811671d99a6ea/skatejs-newdef/test/index-spec.tsx


index.d.ts
```diff
@@ -2,33 +2,29 @@ export { } from "./jsx";
...
-  renderCallback(): any;
-}
-
-export interface OnRenderedCallback {
-  renderedCallback(): void;
-}
-
-export class Component extends HTMLElement {
-  // static readonly props: { [name: string]: PropAttr<Component, any> };
+export class Component extends HTMLElement implements ICustomElementV1, ISkateLivecycle {
```
vvakame: We should not to use `I` prefix.
My definitions does not have `ICustomElementV1` and `ISkateLivecycle` now.

index.d.ts
```diff
@@ -2,33 +2,29 @@ export { } from "./jsx";
...
+  static readonly props: { [nameOrSymbol: string]: PropOptions<El, any> };
```
vvakame: I want to add this line too.
But this definition can't compiling [this code](https://github.com/vvakame/til/blob/49294d3185df832f93c01f5b34f811671d99a6ea/skatejs-newdef/src/prop-examples.tsx#L7-L22).

index.d.ts
```diff
@@ -2,33 +2,29 @@ export { } from "./jsx";
...
   connectedCallback(): void;
   disconnectedCallback(): void;
-  attributeChangedCallback(name: string, oldValue: any, newValue: any): void;
+  attributeChangedCallback(name: string, oldValue: null | string, newValue: null | string): void;
+  adoptedCallback?(): void;
 
-  // SkateJS
-  updated(prev: any): boolean;
+  // SkateJS Livecycle
+  updatedCallback(previousProps: { [nameOrSymbol: string]: any }): boolean;
```
vvakame: We shoud use `previousProps: any` instead of `{ [nameOrSymbol: string]: any }`.
If code safety does not improment, it is better to be easy to learn and easy to use.

andreawyss: I think it is important to communicate that the keys in the map of values can be strings or symbols.

index.d.ts
```diff
@@ -2,33 +2,29 @@ export { } from "./jsx";
 
 export as namespace skate;
 
-export interface HasProps<El extends Component> {
-  props: { [key: string]: skate.PropOptions<El, any>; };
-}
-
-export interface OnUpdatedCallback {
```
andreawyss: I understand, but our users are extending the skate Component and not implementing an interface.
These are methods that exist on the Component class.
Oner can choose to, or not to, override them.
But if overridden one must call super.

andreawyss:
It works again.

index.d.ts
vvakame: I understand. You are right.
Base component class has each callbacks always. Not an "execute if it exists" model.
https://github.com/skatejs/skatejs/blob/316925b0b4a6d1d867abeafe8452123151a1b3d3/src/api/component.js#L291-L313

Hotell:
We need to make `Component` generic so we can define props and use proper TS intellisense when using other Skate Components within particular Skate implementation.

```ts
export class Component<Props> extends HTMLElement implements ComponentLifecycle<Props> {
  _props: Props & ComponentProps;
  static readonly props: { [name: string]: PropOptions };

  // SkateJS
  abstract renderCallback(): JSX.Element | JSX.Element[] | null
}
```

then withing `jsx.d.ts`

```ts
  namespace JSX {
    interface Element extends skate.HNode {}
    interface ElementClass extends skate.Component<any> {}

    interface ElementAttributesProperty<Props extends {}> {
      readonly _props: Props,
    }
```

this "secret" `_props` won't be used directly but it gives us intellisense exactly like with React which is AMAZING :) :D

showcase:

proper-skate-tsx-intellisense


I have this working here:
https://github.com/Hotell/skate-and-raw-WC/tree/proper-skate-type-defs

treshugart:
@Hotell that looks awesome!

vvakame:
sorry, I am very busy now and I think I can add updates on this weekend...

treshugart:
Hey @vvakame lots of awesome work here, but I think we're going to revisit the flow / typescript discussion as typesript doesn't seem to be working out so well thus far. See https://github.com/skatejs/skatejs/issues/972 (I've also pinged you there). Once the decision is made, we can either continue this or close it.

treshugart:
Going to close this as we'll try Flow and see how it works out. As mentioned in your other PR, thanks again for all the hard work. We'll reassess later and going to close this for now.

Hotell:
Oh wow @treshugart. We need this to be shipped, no matter what typed system will be used for skate. Current ts definitions are not sufficient and we have to override those at our projects at work, which very inconvenient :(

treshugart:
Sorry @Hotell, my bad. There were several conversations and seemingly conflicting prs. I'll reopen. Are you able to manage the TS compat please?

Hotell:
> Are you able to manage the TS compat please?
What do you mean by that?

treshugart:
I'm just asking if you can be in charge of it. I'm somewhat out of touch
with that (obviously) and don't have the capacity to focus on getting it in.

Hotell:
Sure, will do!

vvakame:
sorry for tooo late. I'm backed.
I added hotell suggested feature ( https://github.com/skatejs/skatejs/pull/952#issuecomment-264500153 )
It is pretty nice!

I want to try add customelements definitions to TypeScript repo, but it is not fine now.
I added `definitions/w3c-customelements/index.d.ts` to this PR.

vvakame:
oops... test failed...

Hotell:
hey @vvakame will review this week, sorry for delay... looks great so far!

definitions/w3c-customelements/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export interface OnAdoptedCallback {
+    adoptedCallback(oldDocument: any, newDocument: any): void;
+}
+
+export interface OnAttributeChangedCallback {
+    attributeChangedCallback(name: string, oldValue: any, newValue: any): void;
+}
+
+declare global {
```
Hotell: I think that empty `export {}` is missing here

index.d.ts
```diff
@@ -0,0 +1,101 @@
...
+[P in keyof T]: PropOptions<El, T[P]>;
+};
+
+interface ComponentDefaultProps {
+  children?: JSX.Element[];
+  key?: string;
+}
+
+export class Component<Props> extends HTMLElement {
+  _props: Props & ComponentDefaultProps;
```
Hotell: pls add line comment to this "special hack" for further reference

jsx.d.ts
```diff
@@ -0,0 +1,206 @@
...
+}
+
+declare global {
+  // https://www.typescriptlang.org/docs/handbook/jsx.html
+  namespace JSX {
+    interface ElementClass {
+    }
+
+    interface ElementAttributesProperty<Props> {
+      readonly _props: Props,
```
Hotell: also pls add comment here, why we added this "hack"

package.json
```diff
@@ -29,16 +29,21 @@
     "window-or-global": "1.0.1"
   },
   "devDependencies": {
+    "@types/incremental-dom": "^0.5.0",
```
Hotell: do we need those? since skate will not use TS internally? I mean we don't need this for our typings defs right?

tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+        "types": [
+            "w3c-customelements"
+        ],
+        "noImplicitAny": true,
+        "sourceMap": false,
+        "moduleResolution": "node",
+        "strictNullChecks": true,
+        "noEmit": true,
+        "jsx": "react",
+        "jsxFactory": "skate.h"
```
Hotell: pls add `{pretty: true}` for better error reporting

definitions/w3c-customelements/index.d.ts
vvakame: It works with the current code. Is this not okay?
NOTE: https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#augmenting-globalmodule-scope-from-modules

vvakame:
@Hotell thank you for your comment! I added a new commit.
My first language is japanese, not english. let me know if you have a better explanation. 😉

definitions/w3c-customelements/index.d.ts
Hotell: Right, my bad sorry!

treshugart:
This is looking really good!

mattdsteele:
Watching with great enthusiasm; we're building out a new component library, and eyeing SkateJS + TSX as our tooling.

Once this merges could you point me to how this would work in practice? Maybe an example repo, for instance?

treshugart:
@mattdsteele @Hotell has been using TS and Skate for awhile, I believe, though with some issues that this PR will address. He can probably give you some information about how you can start now while waiting for this. @vvakame / @Hotell  how much more work do you think there is?

Docs on this is a great idea.

vvakame:
I'll maintain new usage in https://github.com/vvakame/skatejs-todo/
about document, I am not good at writing English, I want someone to help it. (I ❤️ Google translate )

Hotell:
I can help you with that @vvakame , looks good.

@mattdsteele you can checkout our component library in all it's glory with TS and Skate ;) 
https://github.com/wc-catalogue/blaze-elements

can we merge this @treshugart ? 
It should be breaking change from typings perspective ( because now `Component<T>` is a generic, so if user installs it and used TS before he will have compilation errors ), or we will just proceed with new "feature" version bump with notes what needs to change?

Hotell:
When we merge this I'm going to write blogpost or two about this super powered functional reactive web component library with all mighty Typescript. Oh yes! The Force is strong! :)

treshugart:
@Hotell this is sort of an API grey area. TS support was never official or stable, so I think it's okay to do a feature bump for this. Thanks for checking. Hit the big green button when you're ready :D

treshugart:
@Hotell I didn't see any commitizen commits. Need to have a "feat" in order to do a feature release. Maybe we can update the docs to mention TS support and how to use it and then "feat" in that?

Hotell:
I'll add docs tomorrow

vvakame:
🎉

PR #953, opened 2016-12-02
Updated index.d.ts

andreawyss:
This will close issue #940

Further TypeScript will follow from @vvakame


Hotell:
Can we squash and merge this pls? and maybe release new patch version?

PR #954, opened 2016-12-02
(fix:with polyfill) fixes Firefox and Safari

andreawyss:
This fixes Firefox and Safari.
May also help with IE too.

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -152,8 +152,8 @@ export default class extends HTMLElement {
     setCtorNativeProperty(this, $ctorProps, value);
   }
 
-  constructor () {
-    super();
+  constructor (...args) {
+    super(...args);
 
     const { constructor } = this;
 


andreawyss:
This is such small fix that I will merge it right away.

Hotell:
what bug did it fix? It would be nice to have issue to every PR for further reference what did it solve

Hotell:
also pls clean after yourself, when merged, delete the branch

Hotell:
Last thing :D I dont know what conventions r u using within team @treshugart but IMHO it would be nice to get LGTM from one other member before merging anything, WDYT folks?

andreawyss:
Yes, I agree but since this was such a small obvious one liner fix, I just when ahead and merged it.

constructor (...args) {
    super(...args);

@Hotell @treshugart If one forgets to forward the arguments received in the constructors to the super this will not work when Custom Element is polyfilled. We will need to add this in the documentation for the constructor to avoid users of the library to repeat this same mistake.
For this I have created #956.

andreawyss:
@Hotell @treshugart It turns out that this was required by an older version of a polyfill.

treshugart:
Yeah this was only a requirement for `document-register-element`. Might be best to remove it as we're not only officially documenting support for the `webcomponents` org polyfills and the consumer can just pull in `skatejs-web-components` and be done with it.

PR #955, opened 2016-12-03
refactor(component): Small code cleanup

andreawyss:
Small code cleanup

### DIFF
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -21,7 +21,7 @@ import getOwnPropertyDescriptors from '../util/get-own-property-descriptors';
 import getPropsMap from '../util/get-props-map';
 import getSetProps from './props';
 import initProps from '../lifecycle/props-init';
-import {isFunction} from '../util/isType';
+import { isFunction } from '../util/isType';
 import setCtorNativeProperty from '../util/set-ctor-native-property';
 import syncPropToAttr from '../util/sync-prop-to-attr';
 import root from 'window-or-global';
@@ -303,12 +303,10 @@ export default class extends HTMLElement {
       return updated(this, prevProps);
     }
 
-    // updatedCallback() default implementation
-    //
+    // short-circuits if this is the first time
     if (!prevProps) {
       return true;
     }
-
     // Use getAllKeys to include all props names and Symbols
     const allKeys = getAllKeys(prevProps);
     // Use classic loop because 'for ... of' skips symbols
@@ -318,7 +316,6 @@ export default class extends HTMLElement {
         return true;
       }
     }
-
     return false;
   }
 
@@ -338,8 +335,7 @@ export default class extends HTMLElement {
   //
   // Maps to the static renderer() callback. That logic should be moved here
   // when that is finally removed.
-  // todo: rendererCallback is not documented
-  // Extension point for implementing a different rendering strategies.
+  // todo: finalize how to support different rendering strategies.
   rendererCallback () {
     // todo: cannot move code here because tests expects renderer function to still exist on constructor!
     return this.constructor.renderer(this);
--- a/src/api/component.js
+++ b/src/api/component.js
@@ -190,8 +190,9 @@ export default class extends HTMLElement {
     // static created()
     //
     // Props should be set up before calling this.
-    if (isFunction(constructor.created)) {
-      constructor.created(this);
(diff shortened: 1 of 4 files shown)


andreawyss:
@treshugart please review.

src/api/component.js
```diff
@@ -283,37 +281,70 @@ export default class extends HTMLElement {
...
+    // updatedCallback() default implementation
+    //
+    if (!prev) {
+      return true;
+    }
+
+    // use get all keys so that we check Symbols as well as regular props
+    // using a for loop so we can break early
+    const allKeys = getAllKeys(prev);
+    for (let i = 0; i < allKeys.length; i += 1) {
```
Hotell: `for` is error prone and hard to read, rather use `for .. of`, much smaller footprint and ceremony

```js
  for(const key of allKeys){
    if(prev[key] !== this[key]){
      return true;
    }
  }
```

src/api/component.js
```diff
@@ -179,6 +179,7 @@ export default class extends HTMLElement {
     // DEPRECATED
     //
     // static render()
+    // Note that renderCallback is an optional method!
```
andreawyss: It wasn't obvious to me. 
When we all worked on updating index.d.ts we didn't mark this method optional.
We now need to correct index.d.ts:
renderCallback?()

src/api/component.js
andreawyss: Actually here we cannot use for ... of because it will skip symbols.
I updated the code as follow:

```
// Use classic loop because for ... of will skips symbols
    for (let i = 0; i < allKeys.length; i++) {
      const nameOrSymbol = allKeys[i];
      if (prevProps[nameOrSymbol] !== this[nameOrSymbol]) {
        return true;
      }
    }
```

andreawyss:
@Hotell thank you for the feedback. I made all the requested updates and more. Please review PR.

src/api/component.js
andreawyss: @treshugart is renderCallback optional for optimization?

treshugart: Sometimes you don't need to render in leaf components. This just makes it usable for any type of component.

src/util/isType.js
```diff
@@ -0,0 +1,3 @@
+export function isFunction (val) {
```
treshugart: nit: can just do: `export const isFunction = val => typeof val === 'function';`

src/api/component.js
```diff
@@ -283,37 +283,70 @@ export default class extends HTMLElement {
...
+    if (isFunction(rendered)) {
+      return rendered(this);
+    }
   }
 
   // Skate
   //
   // Maps to the static renderer() callback. That logic should be moved here
   // when that is finally removed.
+  // todo: rendererCallback is not documented
```
treshugart: This is somewhat intentional at this point. I'd like to finalise its implementation before officially making it API.

src/api/component.js
```diff
@@ -283,37 +283,70 @@ export default class extends HTMLElement {
...
+  updatedCallback (prevProps) {
+    // DEPRECATED
+    //
+    // static updated()
+    const { updated } = this.constructor;
+    if (isFunction(updated)) {
+      return updated(this, prevProps);
+    }
+
+    // updatedCallback() default implementation
```
treshugart: Maybe better to say "default implementation short-circuits if this is the first render"?

src/api/component.js
```diff
@@ -283,37 +283,70 @@ export default class extends HTMLElement {
...
+    //
```
treshugart: nit: unnecessary?

src/api/component.js
```diff
@@ -21,6 +21,7 @@ import getOwnPropertyDescriptors from '../util/get-own-property-descriptors';
 import getPropsMap from '../util/get-props-map';
 import getSetProps from './props';
 import initProps from '../lifecycle/props-init';
+import {isFunction} from '../util/isType';
```
treshugart: nit: semistandard probably didn't catch this as it's consistent with itself, but generally there's spaces around object literals, I think.

andreawyss:
@treshugart this merge did not publish a new release 4.1.3.
Did I make a mistake with commitizen or do I need to manually release?

PR #957, opened 2016-12-05
#957 - class PropDefinition holds prop name and normalized prop options

andreawyss:
With the new internal class 'PropDefinition' I have simplified a few functions.

@treshugart @Hotell @bradleyayers this PR is now ready for you to review.
Besides a pure refactor there is a small change in how NaN works for a prop.number().
Setting the prop to NaN will maintain the NaN value and serialize attribute to 'NaN' which deserializes back to NaN.
This should be a small breaking change that will not affect most common usecases.
I had to only change one test and use Object.is for testing equality of NaN equal NaN.

There was also a bit of inconsistency in the code where empty was sometimes normalized to undefined and in other places to null. Now everywhere empty is normalized to null. This didn't affect any tests.
This should also be transparent to most users since each build-in prop had a default that was preventing them to be empty and the default value of the 'default' option was null also.
boolean --> false
number --> 0
string --> ''
array --> []

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,7 +497,7 @@ The parameters passed to the function are:
 
 #### `default`
 
-Specifies the default value of the property. If the property is ever set to `null` or `undefined`, instead of being empty, the `default` value will be used instead.
+Specifies the default value of the property. If the property is ever set to `null` or `undefined`, instead of being set to 'null', the `default` value will be used instead.
 
 ```js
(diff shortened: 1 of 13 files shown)


treshugart:
> This should be a small breaking change that will not affect most common usecases.

The `NaN` thing kinda sounds like a bug anyway. This fixes it for consistency since we can't make sure it's not a `NaN` 100% of the time anyways.

Issue #847, opened 2016-10-10
attribute linked properties reflect state after 2nd mutation in browsers with native shadow DOM ( Chrome,Safari)


Hotell:
When defining component property linked to DOM attribute via `{attribute:true}`, **it doesn't do anything on first mutation via `setAttribute`.**

Looks like this issue is only in browser which have native shadow dom v1 ( chrome, safari 10 )

**Chrome:**

![skate-linked-attribute-chrome-bug](https://cloud.githubusercontent.com/assets/1223799/19242383/25242f3e-8f13-11e6-9529-57fecb1e485e.gif)

**Safari:**

![skate-linked-attribute-chrome-bug_sa](https://cloud.githubusercontent.com/assets/1223799/19242390/2e205e78-8f13-11e6-8b9a-a2f7de69f240.gif)

**Firefox ( works as expected ):**

![skate-linked-attribute-chrome-bug_ff](https://cloud.githubusercontent.com/assets/1223799/19242394/340da4d0-8f13-11e6-86c0-007047509d6e.gif)

Example:

https://plnkr.co/edit/UNZYkOQ0EeUrGHeo85Iy?p=preview

maybe this is unrelated but why does `skatejs-web-components` add deprecated `registerElement` do `document` instead of polyfilling just actual v1 spec? "thanks to that" [both flags(customElement)](https://github.com/skatejs/skatejs/blob/master/src/util/support.js#L4-L5) are always true

treshugart:
> aybe this is unrelated but why does skatejs-web-components add deprecated registerElement do document instead of polyfilling just actual v1 spec? "thanks to that" both flags(customElement) are always true

That's a side-effect of [`document-register-element`](https://github.com/webreflection/document-register-element) that it brings in.

treshugart:
I think this is either an issue with [`document-register-element`](https://github.com/webreflection/document-register-element) not reflecting the change properly, or the internal flags that we keep being out of sync. We keep flags so that recursion doesn't happen when reflection attrs / props. I think we'll probably implement #838 as proposed which will remove the need for the flags and make reflection simpler.

Hotell:
I've switched from `document-register-element` to [custom-elements](https://github.com/webcomponents/custom-elements). Issue still remains.

https://plnkr.co/edit/yF83goIEcZ95dGelKb8z?p=preview

So there are 2 possible reasons why is this happening:
- both polyfills have bug
- skate has damaged tail or nose, so we can't ollie properly :)

[...]

PR #962, opened 2016-12-06
962 - fix(attribute): use attributes manager

andreawyss:
use attributes manager to simplyfy attributes syncing code

closes: 847


src/api/component.js
```diff
@@ -219,6 +209,9 @@ export default class extends HTMLElement {
     // Ensures the component can't be rendered while disconnected.
     this[$connected] = false;
 
+    // Suspend updating Attributes until re-connected
+    getAttrMgr(this).syncAttributes(false);
```
treshugart: I'd prefer a different method name here. Sort of a nit, but I feel this is a potential [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap).

src/api/component.js
```diff
@@ -241,24 +234,19 @@ export default class extends HTMLElement {
...
-      const propData = data(this, 'props')[propNameOrSymbol];
-
-      // This ensures a property set doesn't cause the attribute changed
-      // handler to run again once we set this flag. This only ever has a
-      // chance to run when you set an attribute, it then sets a property and
-      // then that causes the attribute to be set again.
-      if (propData.syncingAttribute) {
-        propData.syncingAttribute = false;
-      } else {
+      let hasChanged = getAttrMgr(this).onAttributeChanged(name, newValue);
```
treshugart: `const`?

src/util/attributes-manager.js
```diff
@@ -0,0 +1,114 @@
...
+const nullIfEmptyOrString = val => (empty(val) ? null : String(val));
+
+/**
+ * @internal
+ * Attributes Manager
+ *
+ * Postpones attributes updates when connected.
+ */
+class AttributesManager {
+
```
treshugart: nit: unnecessary space

src/util/attributes-manager.js
```diff
@@ -0,0 +1,114 @@
...
+  let mgr = elem[$attributesMgr];
+  if (!mgr) {
+    mgr = new AttributesManager(elem);
+    elem[$attributesMgr] = mgr;
+  }
+  return mgr;
+}
+
+/**
+ * todo: We could expose this API?
```
treshugart: Why expose this?

src/util/prop-definition.js
```diff
@@ -59,7 +60,7 @@ export default class PropDefinition {
         case 'get':
         case 'serialize':
         case 'set':
-          if (typeof optVal === 'function') {
+          if (isFunction(optVal)) {
             this[option] = optVal;
           } else {
             console.error(option + ' must be a function.');
```
treshugart: I must have missed this in a previous review? I would prefer to leave out all console error calls directly. If we're going to start doing this (and I've thought it might be a good idea, anyways, I'd like to do something like:

```js
import { isUndefined } from './util/is-type';
import root from 'window-or-global';

function error (message) {
  const { DEBUG } = root;
  if (!isUndefined(DEBUG) && DEBUG) {
    throw new Error(message);
  }
}
```

Then just do:

```js
error(`${option} must be a function`);
```

You should also use template literals here instead of string concatenation.

src/api/component.js
```diff
@@ -23,7 +24,6 @@ import { createNativePropertyDescriptor } from '../lifecycle/props-init';
 import { isFunction } from '../util/isType';
```
treshugart: nit: I didn't catch this before, this should follow the current convention of `is-type`, even if we decide to change this in the future to match the casing of the default export.

src/util/attributes-manager.js
andreawyss: I guess users should not set attributes in the constructor.
Most user's controlled attributes (e.g. aria) are set in the 'set' option function.
I was thinking of a way to expose the deferred 'setAttributeWhenConnected' API to allow them to set those anywhere but to enforce V1 spec.

treshugart: They can just do it in the `connectedCallback()`, right?

andreawyss: Yes but aria attributes are base on the value of a property and it is a lot easier to write that code in the 'set' option function which can execute when the component is not connected.

treshugart: I'm hoping the prop `set` callbacks can be removed in a major release. They promote poor practices IMO, such as DOM side-effects. They may also be based on multiple values from props which would make more sense to do in `updatedCallback()` or `renderedCallback()` instead.

A way that you can get around this in a prop way would be to add aria props

```js
props: {
  ariaLabel: { attribute: true }
}
```

And then:

```js
constructor () {
  this.ariaLabel = 'my label';
}
```

Thus it wouldn't get applied until it's connected anyway. If you're basing it off of existing properties then I think the best way to do this would be in the rendered callback:

```js
renderedCallback () {
  this.setAttribute('aria-label', this.label);
}
```

I feel this function is too low-level to be good as public API, especially since there's other ways to do this in a more "idiomatic skate" way.

andreawyss:
@treshugart this is now pretty clean and ready for merge.
Please review in detail.

src/util/attributes-manager.js
andreawyss: I agree. I will remove that todo: 

I also will add test for #847 to validate that this fixes it.

src/api/component.js
andreawyss: done

src/api/component.js
andreawyss: done

src/util/attributes-manager.js
andreawyss: done

src/util/prop-definition.js
andreawyss: done

karma.conf.js
```diff
@@ -3,6 +3,9 @@ const webpackConfig = require('./webpack.config');
 module.exports = (config) => {
   base(config);
 
+  // Easy way to test locally with a different browser:
```
treshugart: You should be able to just do:

```sh
sk-tests-watch --browsers Chrome,Firefox
```

src/util/prop-definition.js
```diff
@@ -59,18 +61,18 @@ export default class PropDefinition {
...
+            error(`${option} must be a function`);
           }
           break;
         case 'default':
         case 'initial':
           this[option] = optVal;
           break;
         default:
-          console.error(option + ' is not a valid option.');
+          error(`${option} is not a valid option`);
```
treshugart: Could also state what types it supports.

src/util/error.js
```diff
@@ -4,6 +4,8 @@ import root from 'window-or-global';
 export default function error (message) {
   const { DEBUG } = root;
   if (!isUndefined(DEBUG) && DEBUG) {
+    console.error(message);
```
treshugart: What's the rationale here? What are other libs doing in dev mode? Incremental DOM throws. Not sure what React does. I believe they throw, as well.

andreawyss: You suggested some code that was throwing only when DEBUG was true. 
I think we can remove this file and always throw.

src/util/prop-definition.js
andreawyss: Here was you suggestion.

PR #966, opened 2016-12-09
feat(component): Pass element as argument to renderCallback() for eas…

### LINKED ISSUE #958: Pass props as argument to renderCallback
This allows for simple destructuring:

```js
renderCallback = ({ value }) => (
  <input type="text" value={value} />
)
```

Rather than:

```js
renderCallback () {
  return (
    <input type="text" value={this.value} />
  );
}
```

treshugart:
…ier destructuring.

fixes #958


andreawyss:
LGTM

PR #967, opened 2016-12-10
967 - one-way attribute

andreawyss:
@treshugart this needs update to doc.
I need some help writing unit tests for this new feature.
Please take a look at the code. Thank you


treshugart:
@andreawyss lookin good. What aspect of the tests do you want help with? I was thinking that the tests could be co-located with the existing tests in https://github.com/skatejs/skatejs/blob/master/test/unit/lifecycle/properties.js#L189. I know that file is getting a bit large, but we could refactor it after if we feel the need. Thoughts?

src/util/prop-definition.js
```diff
@@ -76,6 +85,8 @@ export default class PropDefinition {
           break;
       }
     });
+
+    this.attrTargetIsSource = this.attrTarget && (this.attrTarget === this.attrSource);
```
treshugart: Do we need to even have the first part of this check here (`this.attrTarget && `)?

src/util/prop-definition.js
```diff
@@ -54,7 +57,13 @@ export default class PropDefinition {
       // Only accept documented options and perform minimal input validation.
       switch (option) {
         case 'attribute':
-          this.attrName = resolveAttrName(optVal, nameOrSymbol);
+          if (isObject(optVal)) {
+            this.attrSource = resolveAttrName(optVal.source, nameOrSymbol);
```
treshugart: Would be nice to descructure the `optVal` props we're accessing here:

```js
const { source, target } = optVal;
```

src/util/prop-definition.js
```diff
@@ -54,7 +57,13 @@ export default class PropDefinition {
...
+            this.attrTarget = resolveAttrName(optVal.target, nameOrSymbol);
+          } else {
+            this.attrSource = resolveAttrName(optVal, nameOrSymbol);
+            this.attrTarget = this.attrSource;
```
treshugart: Wherever we can it would be nice to save references to the props we're accessing. Sort of a nit, though.

src/util/error.js
```diff
@@ -1,11 +1,3 @@
...
-import root from 'window-or-global';
-
 export default function error (message) {
-  const { DEBUG } = root;
-  if (!isUndefined(DEBUG) && DEBUG) {
-    console.error(message);
-  } else {
-    throw new Error(message);
-  }
+  throw new Error(message);
```
treshugart: Why the change here? Don't we still want to error in dev mode?

andreawyss: This will always throw. 
I removed the condition and the DEBUG flag. 
It will never console.error(), it always throws now.

andreawyss:
@treshugart this PR is now ready for merge. Please, review it in details.

src/util/error.js
treshugart: I probably should have worded that differently. We'd always want it to throw in dev, but don't we want this to be a noop in prod (i.e. keep DEBUG)?

treshugart:
I've restarted the builds. This happens a lot with the browser testing services where they just fail for cryptic reasons.

src/util/error.js
andreawyss: I think it should also throw in prop.

treshugart: Ok, let's leave that for now. Can revisit when we use it for more things.

treshugart:
Now getting errors in IE https://travis-ci.org/skatejs/skatejs/builds/183068651#L665

andreawyss:
@treshugart ok this PR is now ready. It works again in IE too.
Please review the changes to the doc too.

src/util/prop-definition.js
```diff
@@ -72,26 +94,28 @@ export default class PropDefinition {
           this[option] = optVal;
           break;
         default:
-          error(`${option} is not a valid option. Options are: attribute, initial, default, coerce, deserialize, serialize.`);
+          deprecated(`${option} is not a valid option. Options are: attribute, initial, default, coerce, deserialize, serialize.`);
```
andreawyss: @treshugart here is the deprecated warning when using undosumented options.

src/util/deprecated.js
```diff
@@ -0,0 +1,5 @@
+export default function deprecated (message) {
+  // TODO start throwing in a future version
```
treshugart: I would say instead that we would think about just removing the deprecation message along with the functionality. New, breaking code shouldn't contain errors for legacy code, I don't think.

src/api/component.js
```diff
@@ -90,6 +91,10 @@ function createInitProps (Ctor) {
   };
 }
 
+function tagName (elem) {
+  return elem && elem.tagName ? elem.tagName.toLowerCase() : '';
```
treshugart: Should be able to rely on `elem` to always be there. You can also use `localName` instead of `tagName.toLowerCase()`.

src/util/prop-definition.js
treshugart: I wouldn't deprecate at all without a path forward. I see this as pretty low priority in the long run. Maybe we can just leave the old behaviour?

src/util/deprecated.js
```diff
@@ -0,0 +1,5 @@
...
+  // throw new Error(message);
+  console.warn(message);
```
andreawyss: Ok. What is the best way to know if you are in Dev or prod?

src/api/component.js
andreawyss: done

src/util/deprecated.js
andreawyss: done

andreawyss:
@treshugart can we merge this?

test/unit/lifecycle/properties.js
```diff
@@ -810,4 +810,132 @@ describe('lifecycle/property', () => {
       });
     });
   });
+
+  describe('attribute one-way', () => {
+    it('source is not updated', (done) => {
+      const safe = uniqueId();
+      define(safe, {
```
treshugart: You don't need to pass a name anymore, you can do an anonymous define() :)

src/util/deprecated.js
```diff
@@ -0,0 +1,5 @@
+export default function deprecated (elem, oldUsage, newUsage) {
+  // TODO: add logic to only warn in dev mode
```
treshugart: A common pattern seems to be `process.env.NODE_ENV`. Most libs tend to favour this and it gets built out of the dist files.

treshugart: Happy to either leave this as is and raise an issue to address it, or to do it now. If nether then I'd rather remove it as it'd be unused. I'll leave it up to you.

Issue #976, opened 2016-12-14
ReferenceError: process is not defined


QiuKejian:
Uncaught ReferenceError: process is not defined
    at deprecated.js:2
    at root (index.js:4)
    at index.js:5

andreawyss:
@QiuKejian Please try version 4.4.1. Thank you

Qiu-Kejian:
@andreawyss 
Hello. I tired 4.4.1. It seems still not working.

Uncaught ReferenceError: process is not defined
    at is-production.js:1
    at root (index.js:4)
    at index.js:5


![image](https://cloud.githubusercontent.com/assets/16420190/21209910/98660c30-c2ba-11e6-9613-2be308908e46.png)

Qiu-Kejian:
Is this because of "use strict"?

treshugart:
Needs to be `typeof process !== 'undefined'`.

treshugart:
Fixed in #979.

treshugart:
I did changed a bit in that PR. Will annotate.

andreawyss:
@QiuKejian Please try version 4.4.2. Thank you

PR #977, opened 2016-12-14
fix(environment): process is not defined

andreawyss:
closes: #976


karma.conf.js
```diff
@@ -15,10 +15,10 @@ module.exports = (config) => {
 
   // Override process.env.NODE_ENV
   webpackConfig.plugins[1] = new webpack.DefinePlugin({
-      'process.env': {
-        NODE_ENV: JSON.stringify('test')
-      }
-    });
+    'process.env': {
+      NODE_ENV: JSON.stringify('test')
```
treshugart: The general convention here seems to be to use `'production'` instead of other strings.

treshugart: Oh, never mind me. Sorry.

src/util/is-production.js
```diff
@@ -0,0 +1,7 @@
+let environment = process && process.env ? process.env.NODE_ENV : null;
```
treshugart: Will uglification still remove the dead code if we're using this utility? I'm not sure t will. I think it explicitly needs to see `process.env.NODE_ENV === 'development'` to remove it.

PR #979, opened 2016-12-15
fix: Fix proces.env issue with builds.

### LINKED ISSUE #976: ReferenceError: process is not defined

treshugart:
Removed process.env.NODE_ENV checks as they're kinda long to dev with and just used a DEBUG global
instead. Added replacements to the proper builds in order to remove this in prod mode and ensured
the karma config is up to date. Removed the is-production util and made the deprecated() util a noop
if DEBUG is off.

fixes #976


src/api/component.js
```diff
@@ -154,7 +154,7 @@ export default class extends HTMLElement {
     // static render()
     // Note that renderCallback is an optional method!
     if (!this.renderCallback && constructor.render) {
-      deprecated(this, 'static render', 'renderCallback');
+      DEBUG && deprecated(this, 'static render', 'renderCallback');
```
treshugart: `DEBUG` is much easier to type and we need to have it at the call site of the thing we want to do so that it gets removed during dead code elimination.

andreawyss:
Thank you for taking care of this. 
Build and configuration management are not at all my expertise.

treshugart:
No probs. Team effort!

package.json
```diff
@@ -25,6 +25,7 @@
   ],
   "dependencies": {
     "incremental-dom": "0.4.1",
+    "rollup-plugin-replace": "^1.1.1",
```
andreawyss: @treshugart should this be a devDependency?

andreawyss:
@treshugart can you explain why we use rollup to build index.js and webpack to build index-with-deps.js?

treshugart:
The rollup bundle is much smaller. I had issues when bundling deps with the main build, thus webpack. Whe webpack 2 is released, we can try consolidating.

treshugart:
In fact, we could try the 2.2 rc.

andreawyss:
Doesn't Webpack have any tree shaking capability?
I would prefer to just use webpack only.

treshugart:
Webpack 2 does, but until we upgrade we'll have to keep the rollup build for size. It's significantly bigger with only Webpack.

PR #982, opened 2016-12-18
docs(Readme): add references to gitbook

Hotell:
@treshugart pls also add the link in repo About, I don't have edit rights


treshugart:
This: https://skatejs.gitbooks.io/skatejs/content

...is the correct one, right?

treshugart:
Done!

PR #996, opened 2017-01-05
Add 'object' to built-in props

### LINKED ISSUE #637: prop.object() prop type
```js
import { Component, prop } from 'skatejs';

export default class extends Component {
  static props = {
    myObj: prop.object()
  }
}
```

Defaults options:

- `default` should be a function that returns an object
- `deserialize` should use `JSON.parse()`
- `serialize` should use `JSON.stringify()`
- `coerce` shouldn't do anything since it's hard to infer the meaning of the input

zzarcon:
Closes #637

### DIFF
--- a/src/api/prop.js
+++ b/src/api/prop.js
@@ -9,10 +9,12 @@ export function create (def) {
   };
 }
 
+const parseIfNotEmpty = val => (empty(val) ? null : JSON.parse(val));
+
 export const array = create({
   coerce: val => (Array.isArray(val) ? val : (empty(val) ? null : [val])),
   default: () => [],
-  deserialize: val => (empty(val) ? null : JSON.parse(val)),
+  deserialize: parseIfNotEmpty,
   serialize: JSON.stringify
 });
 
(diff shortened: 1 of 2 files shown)


src/api/prop.js
```diff
@@ -42,3 +42,9 @@ export const string = create({
   deserialize: toNullOrString,
   serialize: toNullOrString
 });
+
+export const object = create({
+  default: {},
```
treshugart: Should be a function that returns an object (`() => ({})`). This currently would return the same object for all props using this.

src/api/prop.js
```diff
@@ -42,3 +42,9 @@ export const string = create({
...
+  deserialize: JSON.parse,
```
treshugart: See `prop.array()` for what should probably be done here. They can share this function, too.

test/unit/api/properties.js
```diff
@@ -184,6 +184,36 @@ describe('api/prop', () => {
     });
   });
 
+  describe('object', () => {
+    let elem;
+
+    beforeEach((done) => {
+      elem = create(prop.object());
+      afterMutations(done, 1);
```
treshugart: I know you got this call from somewhere else, but it doesn't take the `1` argument. Just need `done`.

zzarcon: My fault, removed the other occurrences as well.

zzarcon:
@treshugart 👀

treshugart:
Can you use `git cz` please and do a `feat` commit? Or squash and merge with `feat` please?

zzarcon:
@treshugart done :crossed_fingers:

PR #997, opened 2017-01-05
fix(ts-typings): move *.d.ts to /src

vvakame:
related #952
sorry, I was leaking a little fix...
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

and other issue I found.
I can't found *.d.ts in 4.5.0.
Should we also change the build process?

```
$ cat node_modules/skatejs/package.json | grep version
  "version": "4.5.0"
$ ls node_modules/skatejs
CHANGELOG.md LICENSE      README.md    dist         package.json src
```


treshugart:
It was in the old one because the `.ts` files were in `src/` and `src/` is in the `files` array in `package.json`. We can either:

- Put the `.ts` files back in `src/`
- Add the `.ts` files to the `files` array in the `package.json`

I vote for the former to keep the config simpler. Are there any issues with that? cc @Hotell

vvakame:
ah, files! I'm using .gitignore or .npmignore always, I forgot it.
I added a commit that do `Put the .ts files back in src/`.

Issue #1004, opened 2017-01-07
Add static "is" property to constructor instead of having to specify name to define()


treshugart:
### Rationale

The purpose of the `define()` function is to add convenience on top of the native `customElements.define()` method. A common pattern, when defining a component, is to use a static `is` property to define the component:

```js
const MyElement extends HTMLElement {
  static is = 'my-element'
}

customElements.define(MyElement.is, MyElement);
```

To simplify this, we can use this property in `define()` automatically:

```js
const MyElement = define(class extends HTMLElement {
  static is = 'my-element';
});
```

*Note: you can pass a vanilla custom element to `define()`.*

#### Additional use case

Another use case for specifying an `is` attribute, that is somewhat orthogonal to using it in `define()`, is being able to consume a component that uses the property in other libraries. For example, React:

```js
const MyElement = skate.define(class extends HTMLElement {
  static is = 'my-element'
});

const MyReact = () => (
  <MyElement.is />
);
```

There is a [react integration library](https://github.com/webcomponents/react-integration) for this, but that might be overkill if you're not exporting the result of reactification. For example, a web component app that might use [react-storybook](https://github.com/storybooks/react-storybook) for developing its components.

### Behaviour

- It overrides the deprecated `name` argument.
- If `is` has already been registered, a unique name is generated, just as with the deprecated `name` argument.
- The `is` property should be set on the passed constructor so that can be made consistent with the name it was registered with

### Other changes

- The `name` argument to `define()` should be deprecated.
- Refactoring of deprecated calls to `define()` in tests.
- The vdom module should use `is` if it's available, but cache using the previously used `$name` symbol.

Hotell:
I would rather say that common pattern is to use static is property `static get is()`. It makes it immutable, of course can be overwritten by Object.defineProperty but regular user won't easily mutate the is name by some mistake if it's just a getter

treshugart:
@Hotell either one works, really. I think it'd be desirable for it to reflect the unique value, if generated. WDYT? For example, if someone specifies:

```js
class MyElement extends HTMLElement {
  static is = 'my-element'
}
```

But that is already registered, what should it do? Skate currently allows this to avoid naming conflicts between multiple versions. Global names is something a lot of people have been vocal against (a +1 for React et al).

What it does in this pr is try and set `is` to the unique value. It feels dirty, but I think it's more desirable to have it in sync than to not.

Hotell:
[...]

PR #1005, opened 2017-01-07
feat(define): Add support for a static "is" property.

treshugart:
implements #1004

Most of the test file changes were removing deprecated `define()` calls.


src/api/define.js
```diff
@@ -11,38 +11,40 @@ export default function (...args) {
...
-    // blocks of logic.
-    if (typeof name === 'string') {
-      throw new Error('When passing only one argument to define(), it must be a custom element constructor.');
-    } else {
-      Ctor = name;
-      name = uniqueId();
-    }
+  // DEPRECATED remove when removing the "name" argument.
+  if (DEBUG && args.legnth === 2) {
+    console.warn('The "name" argument to define() is deprecated. Please define a `static is` property on the constructor instead.');
```
treshugart: @andreawyss this didn't really fit into how `deprecated()` was designed. Do you think it's worth - for the little amount of calls we have to it - to just do `console.warn()`?

src/api/define.js
```diff
@@ -11,38 +11,40 @@ export default function (...args) {
...
+  // "is" takes precendence over "name" and falls back to a unique id
+  let is = Ctor.is || name || uniqueId();
+
+  // Ensure there's no conflicts.
+  if (customElements.get(is)) {
+    is = uniqueId(is);
+  }
+
+  // Ensure the "is" property is consistent.
+  Ctor.is = is;
```
treshugart: All custom elements that go through `define()` get this. This means it mutates the original constructor, but I couldn't find a better way to make this consistent. I'd prefer consistency here, if possible. Curious as to what others think.

src/api/define.js
```diff
@@ -11,38 +11,40 @@ export default function (...args) {
...
-    // an object literal, we can change this to check "function" and invert the
-    // blocks of logic.
-    if (typeof name === 'string') {
-      throw new Error('When passing only one argument to define(), it must be a custom element constructor.');
-    } else {
-      Ctor = name;
-      name = uniqueId();
-    }
+  // DEPRECATED remove when removing the "name" argument.
+  if (DEBUG && args.legnth === 2) {
```
treshugart: The `name` argument is now deprecated.

src/api/vdom.js
```diff
@@ -145,17 +144,22 @@ function resolveTagName (name) {
...
+  if (name.is) {
+    return name.is;
   }
 
-  // If it's a custom element, we get the tag name by constructing it and
-  // caching it.
+  // Get the name for the custom element by constructing it and using the
+  // localName property. Cache it and lookup the cached value for future calls.
   if (name.prototype instanceof HTMLElement) {
+    if (name[$name]) {
```
treshugart: I tried removing having to need to use the `$name` symbol, but decided it was best to keep because I think it's okay to set `is` on CEs passed to `define()`, but don't want to set it on *any* ol custom element as a way to cache the result just in case.

src/index.d.ts
```diff
@@ -3,7 +3,7 @@ export { } from "./jsx";
 export as namespace skate;
 
 export type ComponentProps <El, T> = {
-[P in keyof T]: PropOptions<El, T[P]>;
+  [P in keyof T]: PropOptions<El, T[P]>;
```
treshugart: boyscout

Hotell: 🗡 :D

src/api/define.js
Hotell: why do we need to do this mutation?  It doens't make much sense. If the name si generated by uniqueId and user want's to debug the element, he can just open the devTools and lookup the element name and easy get it from browser global customElements registry

treshugart: I think that's a lot of indirection. The point of this isn't for debugging, though. It enables you to do things with the correct element instance that you wouldn't normally be able to do. For example in React without having to use the React integration lib:

```js
<WebComponent.is />
```

I'm not 100% attached to this idea, but I can't think of a better way for it to work. Maybe a better way is to expose a getter that returns the name that it was registered with?

```js
const MyElement extends HTMLElement {
  static is = 'my-element'
}

// null
MyElement.definedWith;

skate.define(MyElement);

// "my-element"
MyElement.definedWith;
```

Another idea is to only try and define a unique name *if* the `is` property is *not* provided, thus safe to set. Otherwise it tries to use it and errors if it's already registered. In fact, I like that idea the best, I think. Thoughts?

src/api/define.js
```diff
@@ -1,48 +1,49 @@
-import { name as $name } from '../util/symbols';
 import Component from './component';
 import uniqueId from '../util/unique-id';
 import root from 'window-or-global';
 
 export default function (...args) {
-  const { customElements } = root;
+  const { customElements HTMLElement } = root;
```
treshugart: Typo. Interesting how it passed. I'll fix up.

src/api/define.js
```diff
@@ -1,48 +1,49 @@
...
-    // blocks of logic.
-    if (typeof name === 'string') {
-      throw new Error('When passing only one argument to define(), it must be a custom element constructor.');
-    } else {
-      Ctor = name;
-      name = uniqueId();
-    }
+  // DEPRECATED remove when removing the "name" argument.
+  if (DEBUG && args.legnth === 2) {
+    console.warn('The "name" argument to define() is deprecated. Please define a `static is` property on the constructor instead.');
```
treshugart: I don't think it matters here does it? This is just a warning for deprecated code.

src/api/define.js
```diff
@@ -1,48 +1,49 @@
...
   }
 
-  // Ensure there's no conflicts.
-  if (customElements.get(name)) {
-    name = uniqueId(name);
+  // DEPRECATED remove when removing the "name" argument.
+  if (args.length === 1) {
```
treshugart: I thought about that, but I didn't want to unnecessarily call `uniqueId()`. If we move to only generating a `uniqueId()` *if* `is` is *not* present, then this logic changes anyways.

src/api/define.js
```diff
@@ -1,48 +1,49 @@
...
-  // find its name from the constructor in the vdom module, thus improving
-  // performance but still falling back to a robust method.
-  Ctor[$name] = name;
+  // Ensure a custom element is passed.
+  if (!(Ctor.prototype instanceof HTMLElement)) {
+    throw new Error('You must provide a constructor that extends HTMLElement to define().');
+  }
+
+  // "is" takes precendence over "name" and falls back to a unique id
+  let is = Ctor.is || name || uniqueId();
```
treshugart: That is left over from uncommitted changes. TY!

treshugart: This needs `let` here in the current incarnation. Will wait to see what we decide with generating a `uniqueId` in the absence of `is`.

src/api/define.js
Hotell: > The point of this isn't for debugging ergonomics, though, it's a nice side-effect. It enables you to do things with the correct element instance that you wouldn't normally be able to do. For example in React without having to use the React integration lib:

yeah I know what is the purpose of `is` , I'm using it everywhere by now :)

> Another idea is to only try and define a unique name if the is property is not provided, thus safe to set. Otherwise it tries to use it and errors if it's already registered. In fact, I like that idea the best, I think. Thoughts?

yup makes sense! that way, we can enforce proper Elements definition by defining `static is`

src/api/define.js
Hotell: well `uniqueId` will be called only if `is` is undefined or `name` is undefined

here is compiled version:

```js
var _a = args.length === 2 ? args : [args[0].is, args[0]];
var _b = _a[0];
var name = _b === void 0 ? uniqueId() : _b;
var Ctor = _a[1];

```

treshugart: @Hotell I wonder if that's how native behaves?

treshugart: I'll update that when I do the changes for only calling it in the absence of `is`.

treshugart:
@Hotell updated!

src/api/define.js
treshugart: I tried doing this in the last updates, but I found that it was less readable, and I still needed to cater to a `Ctor` that may be an object literal. I found that it's much more clear what's deprecated and what's going on if things were in separate blocks. At the very least, once we remove the deprecations, we can remove the `let` blocks and things should be come just as simple, it'll just take until the next major release.

Hotell: yeah, you're right 👍

treshugart:
Thanks! Will fix up the build and merge it when green.

PR #1006, opened 2017-01-09
Update ts-typings to match latest API additions

Hotell:
- adds `prop.object`
- makes `prop.object` and `prop.array` generic for better type inference
- adds new object type of `attribute` within prop definition


treshugart:
Approved, but I almost feel this is a bugfix rather than a feature.

Hotell:
Well it depends :) but let's make it a "bugfix" this time

PR #1007, opened 2017-01-09
chore(gitbook): tweak demos and add plugins

Hotell:
plugins additions: 

screen shot 2017-01-10 at 10 39 23 am

in action:

skate-gitbook-plugins

### DIFF
--- a/book.json
+++ b/book.json
@@ -1,10 +1,27 @@
 {
   "title": "skatejs",
-  "plugins": ["jsfiddle"],
+  "plugins": [
+    "github",
+    "edit-link",
+    "anchors",
+    "jsfiddle"
+  ],
   "pluginsConfig": {
+    "edit-link": {
+      "base": "https://github.com/skatejs/skatejs/tree/master",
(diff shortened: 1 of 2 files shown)


book.json
```diff
@@ -1,12 +1,28 @@
 {
   "title": "skatejs",
-  "plugins": ["jsfiddle"],
+  "plugins": [
+    "github",
+    "edit-link",
+    "anchors",
+    "jsfiddle"
```
treshugart: I was trying to see how this was presented to the user, but couldn't find a good explanation on the plugin itself. Is there a good example out there?

treshugart:
LGTM for now, but would like to see some jsfilddle examples. I see there's plugins for the various *bin apps. Can always iterate on this if need be.

book.json
Hotell: screen shot 2017-01-10 at 10 39 23 am

skate-gitbook-plugins

Hotell: I will add those pics, gif to PR description for further reference

PR #1009, opened 2017-01-10
[WIP] - Webpack 2

treshugart:
and some misc cleanup:

- indentation
- remove rollup
- move to not depending on skatejs-build because all it's used for now is karma which will be abstracted separately
- move `src/util/deprecated.js` to only be included in the `src/api/component.js` file as it's only used there and its usage is pretty specific to that file
- use `process.env.NODE_ENV` instead of `DEBUG` as Webpack 2 automatically removes it when doing `webpack -p`

The bundle size with rollup as increased from 4k up to 6k since last measuring. Webpack generates a 7k bundle, so for the sake of maintainability, I feel it's a fine trade-off. This means that:

1. GitHub is hardcore caching the image for the bundle size
2. There seem to have been some significant additions, we should look at deprecating anything else we don't need in the next major release.
3. I've removed the callouts on size. Yeah, Skate is small but it sucks having to manually update this stuff. And it doesn't matter that much (within reason) if you tree shake. Don't worry, Skate will always be small and focused.


src/util/root.js
```diff
@@ -0,0 +1,3 @@
+export default (typeof self === 'object' && self.self === self && self) ||
```
Hotell: with this util do we still need `window-or-global` ?

treshugart: No, but this was an experiment because Rollup was erroring on it. It also errored on iDOM so I just nuked it and used only Webpack 2. We can restore the dependency now, if we want.

treshugart:
@Hotell can you rereview pls?

webpack-blocks.js
```diff
@@ -0,0 +1,88 @@
+const { entryPoint, group, setOutput, sourceMaps } = require('@webpack-blocks/webpack2');
```
treshugart: More specifically I'm wondering if you have any better ideas for this file. It's shared code for the webpack configs.

Hotell: hmm, not really, maybe just putting all webpack related stuff to some subfolder, to make root "clean".

Also I'm not a huge fan of using magic things like webpack-block, but that's your call I guess. 

otherwise LGTM

Hotell:
sry for late response on this one, Im on vacation in Alps, so not very good internet connection etc :P 🍻

treshugart:
No worries! I'll leave as is and merge in. I've created https://github.com/treshugart/webpack-blocks-presets but we can pull that in at a later point.

webpack-blocks.js
treshugart: Webpack blocks massively simplified the config for me and allowed use of preset style stuff. Also, there wasn't a need to change configs from Wp1 to Wp2, so win. I know what you mean, though.

PR #1012, opened 2017-01-11
fix(ts-typings): update ts typings

### LINKED ISSUE #1010: ts-typings errors
Following errors are happening because missing TS definitions for `IntrinsicElements` and need to be fixed ASAP

- `class` doesn't work
- `styles` are using core ts dom.lib.d.ts `CSSStyleDeclaration` where are all props mandatory 

skate-ts-errors

Hotell:
Closes #1010 

- also adds proper type checking and intelisense for Stateless Components 🍺 

TODO: 
- [x] extract common types which should be private ( for instance consumer shouldn't be able to import `VDOMElement` from 'skatejs' )

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,142 +1,21 @@
-import { Attributes, HTMLProps, Key, Ref } from './jsx';
-export { } from './jsx';
-
+// UMD library
 export as namespace skate;
 
-export type ComponentProps <El, T> = {
-  [P in keyof T]: PropOptions<El, T[P]>;
-};
+// empty reexport so user get's JSX to global
+export { } from './ts-typings/jsx';
 
-interface ComponentDefaultProps {
-  children?: VDOMElement<any>[];
-  key?: Key;
-}
-
-export interface StatelessComponent<Props> {
-  (props: Props, children?: VDOMNode): VDOMElement<any>,
-}
-export type SFC<P> = StatelessComponent<P>;
-
-interface ComponentClass<PropsType> {
-  new (props?: PropsType): Component<PropsType>;
-}
-export class Component<Props> extends HTMLElement {
-  // Special hack for own components type checking.
-  // It works in combination with ElementAttributesProperty. It placed in jsx.d.ts.
-  // more detail, see: https://www.typescriptlang.org/docs/handbook/jsx.html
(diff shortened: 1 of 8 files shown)


Hotell:
hmm there is some strange issue during build 
`Cannot find module 'regenerate'`  😢

Hotell:
when #1009  is merged can you review this one pls? we are kind of dependent on having these changes backed in ( within https://github.com/wc-catalogue/blaze-elements ), so we can upgrade to latest skate.

cheers!

treshugart:
@Hotell yeah, I upgraded Rollup hoping that it'd be okay, but it broke stuff. That is what ultimately spurred the Webpack 2 rewrite. I've merged that in and now there's a conflict in the tsconfig.

It seems the tsconfig changed in #1009, but I don't remember changing it. Not sure if it was an auto-update from VSCode or what. It looks the same, but I'll let you resolve. Everything else LGTM.

treshugart:
Just a general question, do you think we should have a `src-ts` instead of putting it in `src`? Just thought I'd 🚲 shed that one :)

d-schiffner:
I just pulled the changes but I'm not able to use it with HTMLAnchorElements:

```
Property 'toString' is missing in type 'HTMLProps<HTMLAnchorElement>'.
```

I tried to extend the element with a `toString?: string`, but had no luck. I think this might be related to https://github.com/Microsoft/TypeScript/issues/13345 but without the actual exception.

I changed the definition a bit to be a single interface and it seem to work

Hotell:
@d-schiffner I'm of that issue. But it was there before this PR.  Once I'll resolve this, merge will follow

d-schiffner:
@Hotell ,
thx, good work btw!

If I can help, just give me a note

Hotell:
@treshugart 
> Just a general question, do you think we should have a src-ts instead of putting it in src? Just thought I'd 🚲 shed that one :)

Probably not, we wanna be "consistent" with root public api entry point which is `index.js` and mirror it with `index.d.ts` if we moved the `/ts-typings` folder outside `/src` we would end up with not very nice import file references and so on... 

Maybe in next iteration we can do that, and mirror actual js source files structure with `.d.ts` definitions. Question is if it's worth it ( probably not ? ) :D

PR #1018, opened 2017-01-17
Fix global

treshugart:
Looks like the Webpack 2 refactor broke the global name for `incremental-dom` (`IncrementalDOM`). The `window-or-global` package also doesn't export a global it seems, so I just refactored that out and created a smaller util for it. @Hotell should be happy about that :p

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "build:global": "skate",
+  "global": "skate",
   "name": "skatejs",
   "description": "Skate is a library built on top of the W3C web component specs that enables you to write functional and performant web components with a very small footprint.",
   "license": "MIT",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,5 +1,14 @@
 {
-  "build:global": "skate",
+  "global": "skate",
```
treshugart: For build purposes. This was more of a quick fix. I want to abstract out the blocks we've used and make them more configurable. I'd like to use them in other repos if possible. I'm trying other things than trying to keep skatejs-build up to date. Overriding configs has always been messy and I'm hoping that consuming building blocks that let you compose configs helps. Karma is a mess, though. I hate that config file. I'm hoping we can get wct running some day as it's a bit simpler to set up and use.

Issue #1020, opened 2017-01-18
Stateless components are ignoring slot and ref attributes


Hotell:
I have these SFC ( Stateless functional components === Functional helpers )

```js
const Header = () => (<h1>Hello World</h1>)
const Body = () => (<div>Hey yo body!</div>)
const Footer = () => (<span>Footer baby</span>)
```

and this regular skate WC
```js
class Test extends Component {
  static get is(){ return 'my-test' }
  renderCallback(){
  	return <div>Im TEST</div>
  }
}

class Page extends skate.Component {
  static get is() { return 'my-page' }
  renderCallback() {
    return (
      <main>
        <header><slot name="header" /></header>
        <section><slot name="body" /></section>
        <footer><slot name="footer" /></footer>
      </main>
    )
  }
}

define(Test)
define(Page)
```

root WC:
```js
class App extends skate.Component {
  static get is(){ return 'my-app' }
  _ref = new WeakMap();
  renderCallback() {
    return [
      <Page>
        <Header slot={Page.slots.header} />
        <Body slot="body" ref={ _r => this._ref.set(_r,'Body SFC') } />
        {/* regular skate component projection works. Also ref will trigger. */}
        <Test slot="body" ref={ _r => this._ref.set(_r, 'TEST C') } />
        {/* regular DOM elements projection works as expected */}
        <article slot="body">I do work</article>
        {/* Stateles functional component projection doesnt work. Also ref will not trigger. */}
        <Footer slot="footer" ref={ _r => this._ref.set(_r,'Footer SFC') } />
      </Page>,
      <button onClick={()=>console.log(this._ref)}>show refs</button>,
      <List>
        {['Trey','Martin','ReactiveWC'].map((who,idx) => <ListItem key={idx} who={who}/>)}      
      </List>
    ]
  }
}

define(App)
```

`Page` is used as the projection origin from within `App` component.
If I wanna project those SFCs `Header`,`Body`,`Footer` it wont do anything. The same applies if i wanna `ref` them. 

For regular DOM Elements and regular web components created via `skate.Component` everything works (`Test` component )

bug demo:

https://jsfiddle.net/hotell/xg2sq289

treshugart:
[...]

PR #1022, opened 2017-01-19
fix(ts-typings): add proper emit,props defs and make SFC not allow ref and slot props by default 

Hotell:
What this PR does:
- as discussed within #1020 SFC don't support ref and slot by default. So this is addressed within this fix
- adds proper type annotations to `emit`
skate-ts-emit-annotated

- adds proper type annotations to `props`
skate-ts-props-annotated

- adds missing `role` attribute

### DIFF
--- a/src/ts-typings/api.d.ts
+++ b/src/ts-typings/api.d.ts
@@ -99,7 +99,14 @@ export var prop: {
   object<T extends Object>(attr?: PropOptions<any, T>): PropOptions<any, T>;
 };
 
-export function props(elem: Component<any>, props?: any): void;
+/**
+ * The props function is a getter or setter depending on if you specify the second argument.
+ * If you do not provide props, then the current state of the component is returned.
+ * If you pass props, then the current state of the component is set.
+ * When you set state, the component will re-render synchronously only if it needs to be re-rendered.
+ */
+export function props<P>(elem: Component<P>): P;
+export function props<P>(elem: Component<P>, props: Pick<Component<P>, '_props'>['_props']): void;
 
 export function ready(elem: Component<any>, done: (c: Component<any>) => void): void;
 
--- a/src/ts-typings/api.d.ts
+++ b/src/ts-typings/api.d.ts
@@ -78,7 +78,12 @@ export interface EmitOptions {
   composed?: boolean;
   detail?: any;
 }
-export function emit(elem: EventTarget, name: string, opts?: EmitOptions): void;
+
+/**
+ * Emits an Event on elem that is composed, bubbles and is cancelable by default.
+ * The return value of emit() is the same as dispatchEvent().
+ */
+export function emit(elem: EventTarget, eventName: string, eventOptions?: EmitOptions): boolean;
 
 export var h: typeof vdom.element;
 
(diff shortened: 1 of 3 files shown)


treshugart:
> and make SFC not allow ref and slot props by default

does this mean that someone can't do `<Sfc ref={myref} slot="myslot" />`?

Hotell:
you can ofc, but from Soundness and how SFC works perspective, that's wrong. Consumer has to annotate that SFC with those props like this:

```ts
type Props = { slot?: string, ref?: Function }
const Sfc = (props: Props) => (<div {...props}>Hello Skate!</div>)
```

or via `SFC` skate type
```
type Props = { slot?: string, ref?: Function }
const Sfc: SFC<Props> = (props) => (<div {...props}>Hello Skate!</div>)
```

Then he can within other component do `<Sfc ref={myref} slot="myslot" />`

treshugart:
I'm still kinda lost I think, but it looks like you've defined a TS interface for SFC components to be able to use in order to accept `slot` and `ref` without having to define it themselves? This would be the case with any prop, though. Is it just that `ref` and `slot` aren't given to us by definitely typed or something? In that case, `slot` is a standard attr, so it should probably be pushed back to where the standard definitions live. Ref, I get we'd have to define.

Again, I could be way off, though. I haven't had much time to dig into the typescript stuff.

Hotell:
> In that case, slot is a standard attr, so it should probably be pushed back to where the standard definitions live 

it's already there.

right, my comment could be confusing for non TS native person, sorry for that.

Explanation:

so let's say we have a Functional Component

```js
// javascript
const SayMyName = (props) => <div>{props.name}</div>
```

used by App web component:
```js
// javascript
define( class App extends Component {
  name = 'Tony Hawk;l
  static get is(){ return 'my-app' }
  renderCallback(){
   return (
     <div>Greetings <SayMyName name={this.name}/></div>
   )
  }
});
```

Of course you can paste anything you want to to `SayMyName`

```js
 renderCallback(){
   return (
     <div>Greetings <SayMyName role="foo" id="skate" slot="legend" name={this.name}/></div>
   )
  }
```

but that's programatic error, because SayMyName is a function not a WC or regular DOM and it's api is just `name`. Passing any other prop is pointless because it does nothing.

If you try to do the same in statically typed JS via TS for instance, it will throw compile errors because `SayMyName` props public api is just `name`, which needs to be explicitly defined:

```ts
// typescript
type SayMyNameProps = { name: string }
const SayMyName = (props: SayMyNameProps) => <div>{props.name}</div>
```

now if you do the same within `App`

```ts
 renderCallback(){
   return (
     <div>Greetings <SayMyName role="foo" id="skate" slot="legend" name={this.name}/></div>
   )
  }
```

it will produce compile errors
skate-issue-description-sfc-1

and that's a good thing, because you are trying to use props ( function arguments ) that don't exist in function definition. Also passing slot, ref has to be explicitly within that function, so it really works.

so final correct API would look like this:
```ts
// typescript
type SayMyNameProps = { name: string, slot?: string, id?: string, role?: string }
const SayMyName = ({name,...rest}: SayMyNameProps) => <div {...rest}>{props.name}</div>
```

treshugart:
Yep, that makes total sense now. I'm assuming you can do `...rest` props as an object of `any`, right, just in case you wanted to pass on everything but only type check certain things. Happy for this to be merged, but I'll leave that to you.

PR #1040, opened 2017-01-24
refactor(emit): Reverse behaviour of the composed flag.

### LINKED ISSUE #768: Reverse behaviour of emit() and the composed flag
Currently, `emit()` ensures that events that happen in a shadow root bubble outside of the shadow root because this was the default behaviour before Shadow DOM v1 was finalised and released which breaks anything built on top of existing expectations.

We should consider reversing this behaviour so that we get the ideal scenario where events are scoped by default. Currently, you can get around this by setting `{ composed: false }` in the emit options:

``` js
emit(elem, 'myevent', { composed: false });
```

References #767 where it is fixed to ensure events bubble.

treshugart:
It was originally inverse to the default behaviour to preserve backward compatibility when the spec
churned. Now that we can make breaking changes towards 5.0, we can make this behave as it should by default, for encapsulation purposes.

BREAKING CHANGE: fixes #768

### DIFF
--- a/test/unit/api/emit.js
+++ b/test/unit/api/emit.js
@@ -3,57 +3,43 @@
 import emit from '../../../src/api/emit';
 import fixture from '../../lib/fixture';
 
-describe('api/emit', () => {
-  let child;
-  let parent;
-  let triggered;
-
-  beforeEach(() => {
-    child = document.createElement('child');
-    parent = document.createElement('parent');
-    parent.appendChild(child);
-
-    triggered = 0;
-
-    // All listeners should increment triggered.
-    // Child should always prevent default.
-    child.addEventListener('test', () => ++triggered);
-    child.addEventListener('test', e => e.preventDefault());
-    parent.addEventListener('test', () => ++triggered);
-
-    // It's not spec'd whether or not detached elements should bubble. In some
-    // they do, in some they don't. We ensure they do by adding the nodes to
-    // the document.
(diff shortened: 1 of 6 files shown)


treshugart:
Is it really worth writing a test to assertion built-in behaviour? Maybe to see that we're adhering to the default behaviour.

Hotell:
yup exactly

treshugart:
So I've refactored the tests to see that we're passing the correct options, custom options and custom detail. I've removed the old tests as they were testing built-in behaviour. All we're doing is passing along information. Pls review.

test/unit/api/emit.js
```diff
@@ -3,57 +3,43 @@
...
-    const canceled = !emit(child, 'test');
-    expect(triggered).to.equal(2);
-    expect(canceled).to.equal(true);
-  });
-
-  it('{ bubbles: false } eventOptions', () => {
-    const canceled = !emit(child, 'test', { bubbles: false });
-    expect(triggered).to.equal(1);
-    expect(canceled).to.equal(true);
+describe.only('api/emit', () => {
```
Hotell: remove `only`

treshugart: <3, thought there was a lint rule for that!

treshugart:
Closing to retarget to the 5.0 branch.

treshugart:
Oh, you can retarget now. Win!

PR #1041, opened 2017-01-24
#804 - add way for base Component class to extend base classes

treshugart:
This enables both custom base classes as well customised built-ins.

```js
import { Element } from 'skatejs';

class Base extends HTMLElement {};
export class Default extends Element();
export class Custom extends Element(Base);
export class BuiltIn extends Element(HTMLDivElement);
```

It might be worth raising an issue at some point to detect the tag name of the base class the person is trying to extend so that we can automate customised built-ins within `define()`. For example:

```js
import { Element, define } from 'skatejs';

export default define(class extends Element(HTMLDivElement) {});
```

And internally in `define()`, we could do:

```js
const tagName = getTagNameFromBuiltInClass(HTMLDivElement);
customElements.define(CustomElement.is, CustomElement, { extends: tagName });
```


Hotell:
Hmm AFAIK element extension `{ extends: tagName }` wont be implemented by Apple, also Polymer 2 is saying bye bye to it

Hotell:
so basically Element is a mixin now right?

treshugart:
> Hmm AFAIK element extension { extends: tagName } wont be implemented by Apple, also Polymer 2 is saying bye bye to it

They've started implementing it in the custom element polyfill, so I'm assuming Polymer has changed their stance on it. I've reached out to @justinfagnani to see if he can clarify.

treshugart:
> so basically Element is a mixin now right?

Yes, and still backward compat in this PR. When each part is broken down into a mixin, this can be composed of those. That said, I'm wondering what the performance implications might be of having a long prototype chain between HTMLElement and the resulting custom element.

justinfagnani:
Clarification: Polymer core and elements won't _use_ customized built-in elements, but nothing prevents users from doing that.

treshugart:
It'd be nice to be consistent with `Polymer.Element` if we can.

Hotell:
Hmm I dunno, I don't like to be consistent for all costs with some project X. Skate is much better than Polymer IMHO ( thanks to them the platform is pushed forward, kudos to that without doubt ). So I guess it's your call ;)

treshugart:
It's also consistent with [Bram](https://github.com/matthewp/bram). The only difference is that Skate makes you invoke it like a function to pass on the base impl.

Hotell:
ok

treshugart:
Will merge now and we can discuss prior to releasing 5.0.

PR #1049, opened 2017-01-25
fix(ts-typings): allow renderCallback to pass props via args

### LINKED ISSUE #1045: ts-definitions: add optional props arguments to `renderCallback`
we can use renderCallback args for props access. current TS definitions are missing this nice feature.

Hotell:
Closes #1045

### DIFF
--- a/src/ts-typings/api.d.ts
+++ b/src/ts-typings/api.d.ts
@@ -37,7 +37,8 @@ export class Component<Props> extends HTMLElement {
 
   // SkateJS life cycle
   updatedCallback(previousProps: { [nameOrSymbol: string]: any }): boolean | void;
-  renderCallback(): VDOMElement<any> | VDOMElement<any>[] | null;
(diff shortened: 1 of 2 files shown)


src/ts-typings/api.d.ts
```diff
@@ -37,7 +37,8 @@ export class Component<Props> extends HTMLElement {
 
   // SkateJS life cycle
   updatedCallback(previousProps: { [nameOrSymbol: string]: any }): boolean | void;
-  renderCallback(): VDOMElement<any> | VDOMElement<any>[] | null;
+  // NOTE: infering generics work only on instances, not on implementation type. So this will not give you type safety, you still have to manually annotate those props in your code
```
treshugart: Feels like a given?

Hotell: what do you mean by that?

treshugart: It just feels like that comment explains what one would already know?

Hotell: well not everyone. I was struggling with it for a while. Then I re read TS generic docs and made me tick

PR #1051, opened 2017-01-25
fix(ts-typings): make define generic so it returns proper constructor

### LINKED ISSUE #1050: ts-definitions: make define generic
Currently `define` accepts just `Function` type and returns `any` which is kinda cumbersome, when used with anonymous classes, because we loose proper type from it.

now:
```ts
 const Ctor2 = skate.define(class extends HTMLElement {
    static is = 'x-test-2';
    foo: number;
  });

const inst = new Ctor2();
console.log(inst.helloIdontExistSowhat); // valid no error because  Ctor2 is `any`
```

demanded:
```ts
 const Ctor2 = skate.define(class extends HTMLElement {
    static is = 'x-test-2';
    foo: number;
  });

const inst = new Ctor2();
const IamNumber: number = inst.foo; // valid, known property on Ctor2
console.log(inst.helloIdontExistSowhat); // compile error because  Ctor2 typeof is now properly inferred 
```

Hotell:
Closes #1050

### DIFF
--- a/src/ts-typings/api.d.ts
+++ b/src/ts-typings/api.d.ts
@@ -67,10 +67,17 @@ export interface PropOptions<El, T> {
   set?: (elem: El, data: { name: string; newValue: T | null | undefined; oldValue: T | null | undefined; }) => void;
 }
 
-export var define: {
-  (name: string, ctor: Function): any;
-  (ctor: Function): any;
-};
+interface Define {
+  <T extends Partial<HTMLElement>>(ctor: T): T;
+  /**
+   *  @Deprecated - will be removed in 5.0
+   */
+  <T extends Partial<HTMLElement>>(name: string, ctor: T): T;
+}
+/**
+ * The define() function is syntactic sugar on top of customElements.define() that allows you to specify a static is property on your constructor that is the name of the component, or omit it altogether.
+ */
+export var define: Define;
 
 export interface EmitOptions {
   bubbles?: boolean;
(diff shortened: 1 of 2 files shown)


src/ts-typings/api.d.ts
```diff
@@ -67,10 +67,17 @@ export interface PropOptions<El, T> {
...
-};
+interface Define {
+  <T extends Partial<HTMLElement>>(ctor: T): T;
+  /**
+   *  @Deprecated - will be removed in 5.0
+   */
+  <T extends Partial<HTMLElement>>(name: string, ctor: T): T;
+}
+/**
+ * The define() function is syntactic sugar on top of customElements.define() that allows you to specify a static is property on your constructor that is the name of the component, or omit it altogether.
```
treshugart: Do we need this comment? Seems superfluous as it's in the docs.

src/ts-typings/api.d.ts
```diff
@@ -67,10 +67,17 @@ export interface PropOptions<El, T> {
...
+ */
+export var define: Define;
```
treshugart: `const`?

src/ts-typings/api.d.ts
Hotell: the comment shows proper docs hint within IDE/Editor(vscode)

src/ts-typings/api.d.ts
Hotell: hmm, right. but I'm not 100% sure if in only `var` is supported in ambient definitions . Will take a look

treshugart: That's kinda why I was asking. I'm not 100% either, but I would think it would be.

Hotell: works with `const`

Issue #949, opened 2016-11-28
Split up updatedCallback()


treshugart:
References #945.

To better serve a single responsibility, we should consider splitting up the `updatedCallback()` into two parts:

- `shouldUpdateCallback()` that returns whether or not the component should update (same as React)
- `updatedCallback()` that is called if `shouldUpdateCallback()` returns `true`, and is called before `render()`

*This is a BREAKING CHANGE as you'd be moving the return value for updating from `updatedCallback()` to `shouldUpdateCallback()`. This is a minor breakage because unless someone has overridden the default `updatedCallback()`, it will be transparent to them.*

Hotell:
I would maybe slightly change naming to:
`shouldUpdateCallback`
`updatedCallback` -> `willUpdateCallback`

updatedCallback sounds confusing to me ( like what is this doing, it's calling this LCHook when it's updated ? `will` verb is more explicitly saying in what situation it will be called )

as for https://github.com/skatejs/skatejs/issues/916

it will have React counterparts like this:

**shouldUpdateCallback** ->
- componentWillReceiveProps()
- shouldComponentUpdate()

**willUpdateCallback** ->
- componentWillUpdate()

WDYT?

andreawyss:
In skate we talk about 'render' cycle and not an 'update' cycle
```
   shouldRenderCallback
 
   willRenderCallback
   renderCallBack
   didRenderCallback (renderedCallback)
```

Still not really sure why we need to split this.
What kind of code we expect users to write in willRenderCallback?

I have a usecase where I don't want to use a shadowRoot and I do my rendering with innerHTML inside the current updatedCallback like this:
```
updatedCallback(prevProps) {
        // Using idom when rendering a svg fails silently.
        // The output looks good but does not render the paths!?
        if (!prevProps || prevProps.name !== this.name || prevProps.size !== this.size) {
            this.innerHTML = `<svg viewBox="0 0 24 24" width="${this.size}" height="${this.size}">${icons[this.name]}</svg>`;
        }
        return false;
}
```

andreawyss:
@treshugart
Should updatedCallback execute when the component is disconnected?

React cycle is completely independent of the DOM.  
Current Web Component specs specify different behavior between connected vs disconnected.

To solve this issue on could proxy getAttribute and setAttribute to store attribute values internally and reflect these externally only on connectedCallback.

[...]

PR #1053, opened 2017-01-27
#928 - remove get() and set() from the props definition options

### LINKED ISSUE #928: Remove prop get() / set() and advocate using updatedCallback() instead

treshugart:
Fixes #928.

Also adds a section for documenting migration between major versions.


treshugart:
WTF. Tests are not running in Sauce anymore, but FF 31?!!

Hotell:
hmm this change is kind of unfortunate. Till now user was able to easily create runtime component API validation within `set` now we are "forcing" user to do this within `updatedCallback` which feels dirty ( prop type and validation are in separate places ). This is like if React would enforce to use `shouldComponentUpdate` for `propTypes`. 

It's also unfortunate because I was planning to implement decorators for TS power users so one could write code like this:

```ts
interface UserProps {
 name: string,
 age: number,
}
@customElement('my-user')
class User extends Component<UserProps> {
  @prop() name: string;
  @prop() age: number;
  renderCallback(){}
}
```

where ` @prop() name: string;` would under the hood add this prop definition
```ts
prop.string({
  set(name,{name,newValue}){
    if(typeof newValue !=='string'){
       throw new Error(`${name} - must be a type of 'string'. You used ${typeof newValue}`);
    }
}
  }
})
```

treshugart:
> This is like if React would enforce to use shouldComponentUpdate for propTypes.

I don't really understand the comparison as it doesn't take away props and there's other hooks to do this in if you must, such as `coerce`, which is very similar to what you want to do anyway. This issue has been open for awhile, but hasn't garnered much attention. I figured people just didn't really use them. I've only used them in very edge-casey scenarios where I could just use `updatedCallback()` instead.

To the point of `updatedCallback()`, I'm hoping we can split it up in #949 so that there is better semantic meaning for a `shouldUpdateCallback()` (synonymous with `shouldComponentUpdate()`) and a `propsUpdatedCallback()` (synonymous with `componentWillReceiveProps()`), or something like that. There's no stopping decorators from hooking into those.

I find `updatedCallback()` for handling props a bit hacky, but if split out, I'd much prefer it. It becomes a single entry point that receives batched updates and has consistent microtask timing.

Happy to discuss further if you can find any use cases that can't be worked around.

Hotell:
yes the comparison is not "equally adequate". 

But doing runtime type checks now means that I have to override and reimplement `coerce` for all existing `prop` functions, which is kinda meh.

> There's no stopping decorators from hooking into those.

of course , but  again, from decorator implementation perspective, that would mean monkey patching existing `updatedCallback` which I'm kind of worried.

>  It becomes a single entry point that receives batched updates and has consistent microtask timing.

yup that's great

treshugart:
We'll split out `updatedCallback()` in #949 as stated. This will give the power for decorators to mixin to `propsUpdatedCallback()`. I'm thinking that maybe we should add `set` back into the prop options but not `get`. Thoughts? I'd prefer a different name if so, though, so that it doesn't infer that it will have a getter counterpart.

treshugart:
I'm just not sure if there's a valid use case for it. It'd be quite simple and straight forward to do something in `propsUpdatedCallback()`. For example:

```js
function DecorateSetteres (setterHandlers, Base) {
  return class extends Base {
    propsUpdatedCallback (oldProps, newProps) {
      super.propsUpdatedCallback(props);
      Object.keys(props).forEach(name => setterHandlers[name](
        this, {
          name,
          oldValue: oldProps[name], 
          newValue: newProps[name]
        }
      ));
    }
  }
}
```

Hotell:
> I'd prefer a different name if so, though, so that it doesn't infer that it will have a getter counterpart.

yup definitely there is no need for getter and yup different name as well. 

maybe just `prop.string({propUpdatedCallback(name,oldVal,newVal){}})` to be consistent with  `propsUpdatedCallback` LC hook

> I'm just not sure if there's a valid use case for it.

only one that I can see is that when implementing decorators for skate ( which I'm going to do ) they wont be poisoning/monkeyPatching on LC hook ( which is dirty IMHO )

treshugart:
- `updated`
- `write`

?

treshugart:
> only one that I can see is that when implementing decorators for skate ( which I'm going to do ) they wont be poisoning/monkeyPatching on LC hook ( which is dirty IMHO )

I don't think it's dirty, nor is it "monkey patching". It's doing exactly what classes are designed to support.

Hotell:
> I don't think it's dirty, nor is it "monkey patching". It's doing exactly what classes are designed to support.

we are probably talking about different things.

when I will implement decorator for prop 

```ts
type FooProps = { name: string }
class Foo extends Component<FooProps>{
  @prop() name: string
}
```

under the hood, the decorator would have to monkey patch `propsUpdatedCallback`

```ts
function prop(config){
  return function( proto,propName,descriptor ){
    const ctor = proto.constructor;
    const type = Reflect.getMetadata('desing:type');
    Object.defineProperty(ctor,{
       get props(){   return { prop[type](config) }
    });
    const origUC = proto.propsUpdatedCallback;
    proto.propsUpdatedCallback = function(oldProps, newProps){
        const oldVal = oldProps[propName];
        const newVal = newProps[propName];
        if(typeof newVal !== type){ throw new Error(`whoops ${newVal} is not STRING!` ) }
       origUC.call(this,oldProps, newProps);
    }
   
  }
}
```

instead if we add back the "setter" to prop definition it would be much simpler 

```diff
function prop(config){
  return function( proto,propName,descriptor ){
    const ctor = proto.constructor;
    const type = Reflect.getMetadata('desing:type');
+    const typeCheckedConfig = {...config,{updated(oldVal,newVal){
+      if(typeof newVal !== type){ throw new Error(`whoops ${newVal} is not STRING!` ) }
+    } }}
    Object.defineProperty(ctor,{
       get props(){   return { prop[type](typeCheckedConfig) }
    });
-   const origUC = proto.propsUpdatedCallback;
-   proto.propsUpdatedCallback = function(oldProps, newProps){
-     const oldVal = oldProps[propName];
-     const newVal = newProps[propName];
-    if(typeof newVal !== type){ throw new Error(`whoops ${newVal} is not STRING!` ) }
-      origUC.call(this,oldProps, newProps);
-   }
   
  }
}
```

Hotell:
- [x] updated
- write

treshugart:
It seems you could use `coerce` for that, instead of readding this as `updated`? It actually kinda feels like that's a better hook for you anyways as it has a closer meaning to what you're trying to do. Something that might be worth considering is instead of adding `updated`, is to make coerce accept an `Error` as its return value, similar to how custom React `PropTypes` work. If you can't coerce it (or don't want to), return an `Error`. You retain the functionality you need, just through a more semantic callback. As with React, we would throw in dev mode and ignore in prod.

treshugart:
Just created https://github.com/skatejs/skatejs/issues/1057. Let me know if this won't work for your use case.

PR #1055, opened 2017-01-27
#1054 - add support for react vnodes

### LINKED ISSUE #1054: Accept React as children to a vnode
This would help in integrating better with React so that consumers of both Skate and React could pass off React nodes to Skate and Skate will convert them to iDOM vNodes when rendering.

`custom-element.js`

```js
/** @jsx h **/
import { Component, define, h } from 'skatejs';

export default define(class extends Component {
  static props = {
    tbody: {}
  }
  renderCallback ({ tbody }) {
    return (
      <table>{tbody}</table>
    );
  }
});
```

`react-element.js`

```js
/** @jsx React.createElement */

import React from 'the-gist-above-that-patches-react';
import { render } from 'react-dom';
import CustomElement from './custom-element';

render(<CustomElement tbody={<tbody />} />);
```

treshugart:
fixes #1054


src/api/vdom.js
```diff
@@ -295,6 +295,21 @@ function newElementVoid (tag, ...args) {
...
 
+export function child (ch) {
+  const ctype = typeof ch;
+  if (ctype === 'function') {
+    ch();
+  } else if (ctype === 'string' || ctype === 'number') {
+    newText(ch);
+  } else if (Array.isArray(ch)) {
+    ch.forEach(child);
+  } else if (ctype === 'object' && ch.$$typeof) {
```
Hotell: hmm so you don't like commenting code. Me neither. Next time maybe it could be better to name things appropriately via function helper or variable.

What I would like to see when reading the code would be something like this:
```js
export function child(ch){
 const ctype = typeof ch;

 if (isFunctionHelper(ctype)) {
  ch();
  } else if (isTextNode(ctype)) {
     newText(ch);
  } else if (Array.isArray(ch)) {
     ch.forEach(child);
  } else if (isReactNode(ctype)) {
     // ...
  }

}
 
 function isFunctionHelper(child){
   return ctype === 'function'
 }
 function isTextNode(child){
   return ctype === 'string' || ctype === 'number';
 }
  function isReactNode(child){
    return ctype === 'object' && ch.$$typeof
  }
}
```

WDYT?

src/api/vdom.js
```diff
@@ -295,6 +295,21 @@ function newElementVoid (tag, ...args) {
...
+    const { key, props, ref, type } = ch;
```
Hotell: I completely understand that you need to support React because you are using it at work, but IMHO this shouldn't be in core `skate`. this interop should be provided by some plugin system via separate package in the future. Relying on some 3rd party implementation is compatibility and maintenance hell in my experience.

treshugart: I had my concerns, but it really hit me when I was doing something simple in React (not at work) and wanted to pass off a node to a custom element table (that can't have slots because `<tr>` can't be a child of a custom element, etc). I wanted to use Storybook, and opted to just use react for the main story parts to render out the custom elements within the storybook. No can do.

I also think this is a major step in supporting framework interop. I'm not sure exactly how we'd do this as a separate package because we'd need to expose hooks for it. I think in some way, shape or form, we need to support the large frameworks in some way, and make it simple for consumers to use (like not have to search for and find new libs). I was actually caught off guard that this *didn't* work until I figured out that I was passing react nodes off to iDOM. Passing props and it "just working" is something that's easy to take for granted. I think others would be surprised, even though it makes sense when you think about it.

This probably also works with Preact (cc @developit). I'd also like to support real DOM nodes. All of these things are things that https://github.com/skatejs/dom-diff already supports, along with web workers.

If we made this more generic and removed the `$$typeof` check, then this actually would be mostly agnostic to react nodes and it would be just accepting an object with a particular schema, or schemas if we supported more reactalike libs.

Open to suggestions about how to do this in a more decoupled manner, but I think this is a pretty rad feature that would make writing web components within other virtual dom frameworks a breeze. It would definitely one-up other offerings.

treshugart: Maybe one way to do this would be to offer a special prop for it. Like `prop.node()`?

treshugart: Created #1056.

developit: `preact-compat` does support `$$typeof` the same as React, yup.  Preact itself currently relies on duck type checking of VNodes, and the format for VNodes is different - it uses the DOM names (`{ nodeName:'', attributes:{}, children:[] }`).  That may change in the future, since we'll likely drop the `VNode` constructor entirely and go to POJOs, in which case it might be useful to add a `$$typeof` flag.  It might also be worth just switching to `{ type, props, children }` as React uses, since `props` is used as a term basically everywhere else.

treshugart: Cool, thanks for the clarification @developit!

Issue #1039, opened 2017-01-24
Attribute linking should be one-way (attr -> prop) by default for all built-in props


treshugart:
All built-in props should be updated to specify: `{ attribute: { source: true } }`.

Hotell:
Closed by #1062

PR #1062, opened 2017-01-31
#1039 - one-way attribute reflection (attribute -> property)

treshugart:
Fixes #1939.


test/unit/api/properties.js
```diff
@@ -31,7 +31,7 @@ function testTypeValues (type, values, done) {
   }, 1);
 }
 
-describe('api/prop', () => {
+describe.only('api/prop', () => {
```
Hotell: remove `only`

treshugart: Sigh. As much as I hate eslint config, semistandard isn't cutting it :p

treshugart: done

treshugart:
Updated docs!

treshugart:
Will merge now (since I've fixed stuff and you're probably asleep) but happy to fix any lingering stuff in 5.x after the fact. Cheers!

Hotell:
Nice! Oh it's only 11am in Prague so all good. Haha

Issue #978, opened 2016-12-15
Remove ready() function


treshugart:
This has proven to be an edge case for imperative coding. We should remove this in a future major release.

treshugart:
Done in #1066.

PR #1066, opened 2017-02-02
#978 - remove ready() function and related code

treshugart:


test/perf/react.js
```diff
@@ -1,129 +0,0 @@
-/* eslint-env jasmine, mocha */
```
treshugart: This was waaay out of date and used `ready()` so just removed.

test/unit/deprecated/lifecycle.js
```diff
@@ -1,113 +0,0 @@
-/* eslint-env jasmine, mocha */
```
treshugart: This used `ready()` and was testing stuff that the custom element polyfill should be tested to perform.

Hotell:
But the main removal is already merged via https://github.com/skatejs/skatejs/pull/1063 right?

treshugart:
@Hotell that just removed one of the tests.

PR #1068, opened 2017-02-02
#1067 - docs(index): emphasise that you don't need to worry about transpile p…

### LINKED ISSUE #1067: Update es2015 classes docs
In https://skatejs.gitbooks.io/skatejs/content/#transpilation-and-native-custom-element-gotchas we should massage the description to add "If you’re using Babel or some other tool to transpile your ES2015 code to ES5, you don’t need to worry about this.".

treshugart:
…roblems if you use skatejs-web-components and are transpiling

Fixes #1067.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ If you want finer grained control about which polyfills you use, you'll have to 
 
 ### Transpilation and native custom element gotchas
 
-*If you’re using Babel or some other tool to transpile your ES2015 code to ES5, simply import `skatejs` and `skatejs-web-components` (or selectively include the polyfills) as normal and ignore the following.*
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -140,6 +140,8 @@ If you want finer grained control about which polyfills you use, you'll have to
 
 ### Transpilation and native custom element gotchas
 
+*If you’re using Babel or some other tool to transpile your ES2015 code to ES5, you don’t need to worry about this if including `skatejs-web-components`.*
+
```
aendrew: One of the biggest questions I had was whether you *should* include `skatejs-web-components` (my impression now is that the answer to that is "yes, almost definitely"). Maybe reword to something like:

> If you’re using Babel or some other tool to transpile your ES2015 code to ES5, simply install `skatejs-web-components` and ignore the following.

treshugart: Done! Added a few words for clarification. Let me know if it's too much.

PR #1074, opened 2017-02-03
#1073 - props overhaul

### LINKED ISSUE #1073: prop.* overhaul
- remove `create()`
- make each prop type a frozen object rather than a function
- update docs
- consider making default `array` and `object` a shared frozen array / object (we should somehow communicate this and encourage immutability and more functional constructs like `concat()` and spread

treshugart:
Fixes #1073.

The goal of this change is to:

- Remove code that's hardly used (`create()`) in favour of composing objects yourself. Can just use `{ ...propArray, ...{ attribute: true } }` if you want.
- Immutability, via `Object.freeze()`, as we're not using a `function` anymore, but an `object` literal instead.

*Exporting each prop individually isn't necessary. The current model will actually [tree shake](http://rollupjs.org/?version=0.41.4&shareable=JTdCJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm1vZHVsZU5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwJTdCJTIwcHJvcCUyMCU3RCUyMGZyb20lMjAnLiUyRmluZGV4JyUzQiU1Q24lNUNuZXhwb3J0JTIwY29uc3QlMjB0ZXN0JTIwJTNEJTIwcHJvcC5hcnJheSUzQiUyMiU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJpbmRleC5qcyUyMiUyQyUyMmNvZGUlMjIlM0ElMjJpbXBvcnQlMjAqJTIwYXMlMjBwcm9wJTIwZnJvbSUyMCcuJTJGcHJvcCclM0IlNUNuZXhwb3J0JTIwJTdCJTIwcHJvcCUyMCU3RCUzQiUyMiU3RCUyQyU3QiUyMm5hbWUlMjIlM0ElMjJwcm9wLmpzJTIyJTJDJTIyY29kZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwYXJyYXklMjAlM0QlMjAlN0IlN0QlM0IlNUNuZXhwb3J0JTIwY29uc3QlMjBib29sZWFuJTIwJTNEJTIwJTdCJTdEJTNCJTVDbmV4cG9ydCUyMGNvbnN0JTIwbnVtYmVyJTIwJTNEJTIwJTdCJTdEJTNCJTVDbmV4cG9ydCUyMGNvbnN0JTIwb2JqZWN0JTIwJTNEJTIwJTdCJTdEJTNCJTVDbmV4cG9ydCUyMGNvbnN0JTIwc3RyaW5nJTIwJTNEJTIwJTdCJTdEJTNCJTIyJTdEJTVEJTdE), at least in Rollup.*

- [x] remove `create()`
- [x] make each prop type a frozen object rather than a function
- [x] update docs
- [x] make default array and object a shared frozen array / object (we should somehow communicate this and encourage immutability and more functional constructs like concat() and spread
- [x] Update TS definitions

This will update props to now be used like:

```diff
import { define, prop } from 'skatejs';

define(class extends Element () {
  static props = {
-   test: prop.array()
+   test: prop.array
  }
});
```


Hotell:
Looks good so far. I guess it will even simplify Typescript definitions and inference.

marcoms:
@treshugart to clarify, is the `static x = y;` syntax in a class expression TS only? I couldn't find it in an ecmascript spec but that's quite cool otherwise.

treshugart:
@marcoms see https://github.com/tc39/proposal-class-public-fields

treshugart:
@Hotell can you pls rereview when you get a chance. Particularly, the TS typings. I did my best :)

src/ts-typings/api.d.ts
```diff
@@ -99,11 +97,11 @@ export const h: typeof vdom.element;
 export function link(elem: Component<any>, target?: string): (e: Event) => void;
 
 export const prop: {
-  number<E extends Component<any>, T extends number>(attr?: PropOptions<E, T>): PropOptions<E, T>;
-  boolean<E extends Component<any>, T extends boolean>(attr?: PropOptions<E, T>): PropOptions<E, T>;
-  string<E extends Component<any>, T extends string>(attr?: PropOptions<E, T>): PropOptions<E, T>;
-  array<E extends Component<any>, T>(attr?: PropOptions<E, T[]>): PropOptions<E, T[]>;
-  object<E extends Component<any>, T extends Object>(attr?: PropOptions<E, T>): PropOptions<E, T>;
+  number: PropOptions;
```
Hotell: add `readonly` to make TS aware that those are frozen and immutable

```ts
{
  readonly number: PropOptions;
  readonly boolean: PropOptions;
  // ...
}
```

src/ts-typings/api.d.ts
```diff
@@ -58,15 +58,13 @@ type AttributeReflectionConfig = AttributeReflectionBaseType | {
...
   attribute?: AttributeReflectionConfig;
-  coerce?: (value: any) => T | null | undefined;
-  default?: T | null | undefined | ((elem: El, data: { name: string; }) => T | null | undefined);
-  deserialize?: (value: string | null) => T | null | undefined;
-  get?: <R>(elem: El, data: { name: string; internalValue: T; }) => R;
-  initial?: T | null | undefined | ((elem: El, data: { name: string; }) => T | null | undefined);
-  serialize?: (value: T | null | undefined) => string | null;
-  set?: (elem: El, data: { name: string; newValue: T | null | undefined; oldValue: T | null | undefined; }) => void;
+  coerce?: <T>(value: any) => T | null | undefined;
+  default?: any | ((elem: HTMLElement, data: { name: string; }) => any);
```
Hotell: `any` === that's sad :( :D but yeah, TS doesn't support generics on properties/variables ( yet )

test/definitions/misc.tsx
```diff
@@ -440,33 +440,34 @@
...
+  skate.prop.array;
+  skate.prop.object;
+
+  const customProp = {
+    ...skate.prop.boolean, ...{
+      coerce() {
+        // coerce it differently than the default way
+        return false;
+      },
+      set() {
```
Hotell: we can remove this, as `set` is gone in 5.x

src/ts-typings/api.d.ts
treshugart: :( I was trying hard to find another way, but no can do unless `PropOptions` is a function, it seems.

src/ts-typings/api.d.ts
treshugart: Done!

test/definitions/misc.tsx
treshugart: and done

treshugart:
@Hotell done with changes.

src/ts-typings/api.d.ts
Hotell: Yeah I know. It's fine for now, we are using decorators for defining props, which feels more first class citizens to TS users ( because definition and type annotation of prop are together )

Hotell:
Thanks and LGTM!

Issue #1065, opened 2017-02-02
Remove defined attribute from being set during element upgrading


treshugart:
Since we're close to shadow DOM support, there's the Shady CSS polyfill and the consumer can do the same thing, I think we should remove support for setting the `defined` attribute.

Hotell:
hmm, ShadyCss does that?

Hotell:
also [ShadyCss doesn't work with skate](https://github.com/skatejs/skatejs/issues/1027) at this moment, and probably won't in the future, so the question is why use it at all.

About the define, we are adding it just to "polyfill" `:defined` in non native browsers right?

treshugart:
I believe they polyfill `:defined`, but I could be wrong. I actually got it working before, but ever since they started removing the style tag, I couldn't get the transformed css from it. There must be a way; it feels far to connected to a single particular way to write web components using <template> which isn't ideal for many.

Hotell:
So they are not polyfiling it http://codepen.io/Hotell/pen/ggQELy?editors=1010 

AFAIR it is not possible, to polyfill it

Hotell:
<img width="1015" alt="screen shot 2017-02-11 at 2 08 58 pm" src="https://cloud.githubusercontent.com/assets/1223799/22853996/cc2e0f46-f063-11e6-90bc-069ca2b56835.png">

Hotell:
https://developers.google.com/web/fundamentals/getting-started/primers/customelements#polyfill

TL;DR

>  Note: the :defined CSS pseudo-class cannot be polyfilled.

thepian:
I think it would be nice to know what the target for Skate is conceptionally in terms of browser support and the type of website. It currently seems aimed at supporting prototyping rather than mainstream production, but that would of course shift as older browsers fall off the grid. 

If I understand this ticket correctly, it implies that you will aggressively move forward, which sounds correct for prototypes, but not good for building something widely deployed.

treshugart:
Interesting. It'd be possible if the ShadyCSS polyfill hooked into the custom Custom Element polyfill. `<custom-style>` would be able to transform `:defined` to something that the CE fill adds when a component is upgraded. Why wouldn't that be polyfillable?

Maybe it's worth leaving this in then. I'm wary of the performance implications as attribute sets are three times slower than property sets and doing that for every upgrade across the board seems a bit much when the consumer can opt into it by simply doing it manually, using decorators or using mixins.

thepian:
[...]

Issue #1038, opened 2017-01-24
Remove complex iDOM wrapping code


treshugart:
We wrap *A LOT* around Incremental DOM. The notable parts are:

- We try and make skipping more intuitive, as in you can turn it off and on selectively. This probably doesn't need to be done anymore if they now don't empty trees that have never had virtual nodes declared, but have DOM in them.
- Stateless functional components can be done by passing a function as a tag name. This forces us to have to wrap every function so we can buffer calls, and "build up" a tree which we can pass a `children` argument to. We probably don't need to do this if we return functions for everything.

Skipping isn't really necessary, especially if Incremental DOM now ignores trees that never had any nodes in the first place.

Stateless functional components are more complex. We should consider simplifying this as much as possible, even if it makes the API change a little.

- [ ] prove Incremental DOM ignores trees with no virtual nodes declared (use `ref` to update the node and do several renders)
- [ ] ensure you can still do stateless functional components in a similar manner (probably calling a function rather than passing a function to a tag name

thepian:
Sounds great. It is really hard to extend the iDOM functionality.

treshugart:
Replaced Preact with iDOM. See referenced PR for details.

Issue #992, opened 2017-01-04
Offer a dist with native class support


treshugart:
References #988.

If you're using native ES classes, you need a way to consume the untranspiled base `Component` class. However, since the consumer shouldn't have to worry about what ES features we're using, we should still transpile all other features we use.

- [ ] Transpile to latest stable ES spec (currently ES2015)
- [ ] Transpile to `dist/es-next`
- [ ] The `module` field in the `package.json` is the only exposed API for the `es` entry point.
- [ ] Consuming `dist/es/index.js` directly is not supported, the consumer should use `module`.
- [ ] Existing `main` should be moved to `dist/es-base`.
- [ ] Remove note in the README referencing this issue and add docs on how to consume the new file

Open questions:

- [ ] Should we still offer a minified file?

Hotell:
consider this ( which is kind of standard nowadays )
- `bundles/skate.umd.js` - UMD bundle -> `"main":"bundles/skate.umd.js"`
- `lib/esm/index.js` - es6 ( non transpiled and not bundled ) -> `"module":"lib/esm/index.js", "jsnext:main":"lib/esm/index.js"`
- `lib/fesm/skate.js` - bundled es6 ( [Flat Ecmascript modules](https://github.com/angular/angular/blob/master/CHANGELOG.md#flat-es-modules-flat-esm--fesm) - pioneered by Angular team - top notch parse performance and smaller payload - achievable only via Rollup ) -> `"es2015":"lib/fesm/skate.js"`

treshugart:
So, in #1075, I implemented NWB. I was having a build issue and just said f-it and it only took a few mins. On the plus side, it's removed a lot of maintenance and it added an `es` dir to the published package, which has es5 + esm files. Along side it, it also has `lib` for commonjs es5 and `umd` for, you guessed it, umd.

I think this is a good compromise for now as we get to offload the build choices to another (popular) library. Will close for now and we can raise issues to iterate on that front from there.

Hotell:
k I left some comments in that PR

Issue #961, opened 2016-12-06
npm install failing on prepublish sk-bundle ( windows only )


kentendo:
I tried this on multiple windows boxes with different versions of npm and node. Works great on Mac :)

windows ( 7 - 10 )
node ( 4.2.6 - 6.9.1 ) 
npm ( 2.14.12 - 3.10.8 )

1. git clone https://github.com/skatejs/skatejs.git
2. npm install
3. fail

```
The system cannot find the path specified.

npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run" "prepublish"
npm ERR! node v6.9.1
npm ERR! npm  v3.10.8
npm ERR! code ELIFECYCLE
npm ERR! skatejs@ prepublish: `sk-bundle`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the skatejs@ prepublish script 'sk-bundle'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the skatejs package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     sk-bundle
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs skatejs
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls skatejs
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     C:\Git\skatejs\npm-debug.log

```

treshugart:
Could you try and find the relevant error in `npm-debug.log` and post that, please? Thanks.

kentendo:
Reading this, the error is still not clear to me, but here are the relevant logs.
```
52 silly lifecycle skatejs@~preinstall: no script for preinstall, continuing
53 silly executeActions Starting
54 silly install executeActions
55 silly doSerial global-install 0
56 silly doParallel fetch 0
57 verbose correctMkdir C:\Users\kcosca\AppData\Roaming\npm-cache\_locks correctMkdir not in flight; initializing
58 verbose lock using C:\Users\kcosca\AppData\Roaming\npm-cache\_locks\staging-ff118b6297fd35a4.lock for C:\Git\skatejs\node_modules\.staging
59 silly doParallel extract 0
[...]

PR #1075, opened 2017-02-04
Several refactors (issues noted in description)

treshugart:
There are several refactors in this issue since they were very tightly coupled to one another. Issues noted below.

- [x] Split out `Component` into mixins (#927)
- [x] Use Preact instead of IncrementalDOM (#1038, #1094)
- [x] Remove `defined` attribute from being set (#1065)
- [x] Switch to NWB for build (#961, #992)
- [x] Split up `updatedCallback()` (#949)
- [x] Separate tests for each mixin
- [x] Update docs


treshugart:
Added a note: I don't think we need more than two mixins:

- A base without a renderer
- A default renderer extending the base

package.json
```diff
@@ -16,76 +7,57 @@
...
     "semantic-release": "^6.3.2",
     "semistandard": "^9.0.0",
     "skatejs-web-components": "5.2.1",
     "typescript": "^2.1.4",
-    "typescript-formatter": "^5.0.0",
-    "webpack": "2.2.1"
+    "typescript-formatter": "^5.0.0"
+  },
+  "peerDependencies": {
+    "preact": "^7.2.0"
```
Hotell: so this is very intense change indeed.

Can you elaborate pls a little why are we moving away from IncrementalDom ? 

Don't get me wrong I love preact and we are using it with Skate at work already, but you know I have to have good reasoning for peer devs, why and how. 

Also thumbs up for using peerDeps ;)
cheers!

package.json
```diff
@@ -16,76 +7,57 @@
     "type": "git",
     "url": "https://github.com/skatejs/skatejs"
   },
-  "main": "dist/index.js",
-  "jsnext:main": "src/index.js",
+  "main": "lib/index.js",
```
Hotell: this is path to a bundled umd ?

package.json
```diff
@@ -16,76 +7,57 @@
     "type": "git",
     "url": "https://github.com/skatejs/skatejs"
   },
-  "main": "dist/index.js",
-  "jsnext:main": "src/index.js",
```
Hotell: this is needed for SystemJS users AFAIK

nwb.config.js
```diff
@@ -0,0 +1,23 @@
+const browsers = require('./test/browsers');
```
Hotell: r we going to change build tools every month or so ? :P :D  :trollface:  next one --> https://neutrino.js.org/

treshugart: I haven't seen that one. Looks promising. If it makes sense, yeah, worth a change. Investing in a tool that you have minimal configuration for means that switching is cheap, especially if you haven't exposed your directory structure as API (using "main" and "modules").

package.json
treshugart: Preact has most of the same semantics that we were wrapping into incremental DOM that we now get for free. We lose the several hundred of lines of code we had to write and maintain.

package.json
treshugart: This is the path to the commonjs code. UMD sits in `umd`.

package.json
treshugart: https://github.com/rollup/rollup/wiki/pkg.module

package.json
treshugart: https://github.com/skatejs/skatejs/issues/1094

package.json
Hotell: perfect, actually SystemJS doesn't care as you can configure it whatever you want ( also it uses Rollup under the hood so I guess it's default behaviour is like rollups ).  

Thanks for clarifying!

package.json
Hotell: hmm well and that's the thing, we don't need specific commonjs build, I think main should point to UMD build

nwb.config.js
Hotell: Right, don't get me wrong, for sure easier setup === easier life for maintainers, but if we just used webpack 2, every new contributor could be on same page as it's generely known bundler.  But yeah in the end I don't mind.

package.json
Hotell: reasoning:
- I use es6 modules aware bundler ( webpack, rollup, systemJS ) -> I will consume `es6 module build`
- I use oldish bundler like requireJS, browserigy -> I will consume `UMD` 
- I don't use bundelr -> I will consume `UMD`
- I'm NodeJS -> I will consume `UMD`

treshugart: This is the default configuration. I'd rather just keep it as is because it's less opinion for us to have to worry about. If it changes, we update the `package.json`.

nwb.config.js
treshugart: NWB is pretty well documented. All people need to worry about, really, is writing code. I'm not too worried about this. If people have questions, they can refer to those docs or ask questions, but it's less config and stuff in the way to worry about in the end for everyone.

package.json
Hotell: yup, fair enough

treshugart:
Since most of this done, it's going to the 5.x branch and it's green, I'm going to merge it. I feel like we should keep 5.x alive until we're happy with it.

package.json
developit: @treshugart - might be worth using `*` for the peer dep. I've run into some nasty issues with locked versions.

treshugart: Good advice. Cheers, @developit :)

PR #1076, opened 2017-02-05
vdom symbols exported

thepian:
It seems that some symbols are deprecated from the source comments. I needed to programatically set skip, but the skip symbol isn't in symbols.js

I assume that it should be along with the others.

### DIFF
--- a/src/api/vdom.js
+++ b/src/api/vdom.js
@@ -9,7 +9,13 @@ import {
   symbols,
   text
 } from 'incremental-dom';
-import { name as $name, ref as $ref } from '../util/symbols';
+import {
+    name as $name,
+    ref as $ref,
+    skip as $skip,
+    currentEventHandlers as $currentEventHandlers,
+    stackCurrentHelperProps as $stackCurrentHelperProps,
+} from '../util/symbols';
 import propContext from '../util/prop-context';
 import root from '../util/root';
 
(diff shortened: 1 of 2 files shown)


treshugart:
I'm curious of your use case and also how you're setting skip programatically.

If you're importing anything other than files in `dist/`, or using any API outside of `src/api/`, that's currently unsupported API, just FYI. If you're exposing `$skip` in hopes to import it and use it in your code, there might be a better way to workaround the issue you're having.

How are you using it?

thepian:
I have an alternate implementation of the `h()` which supports custom elements inside, triggering the rendererDebounced on it.

```JS
        const node = vdom.elementOpenEnd(tname);

        // Skate Custom Elements without children are skipped and made to render themselves
        if (node[$rendererDebounced] && chren.length === 0) {
            node[$skip] = true;
            if (node[$rendererDebounced]) node[$rendererDebounced](); // render when element is closed
        } else {
            chren.forEach((ch) => {
                const ctype = typeof ch;
                if (ctype === 'function') {
                  ch(vdom);
                } else if (ctype === 'string' || ctype === 'number') {
                  vdom.text(ch);
                } else if (Array.isArray(ch)) {
                  ch.forEach(sch => sch(vdom));
                }
            });
        }

        return vdom.elementClose(tname);
```

treshugart:
Sure, but what is your `h` implementation solving? I'd like to drill down your use case further to see if there's a better way to do what you want. Why do you need to manually trigger rendering?

thepian:
I render to the normal DOM, it doesn't work with nested custom elements without this. Otherwise I need to put a skip attribute on the JSX where I use the custom elements. That works as well, but it seems like an implementation exposed in interface for this use-case.

treshugart:
I really don't want to export these symbols as public API especially since that part of it is changing in 5.0. It really feels like a custom iDOM renderer should be used instead. Either that or your suggestion on using the skip attribute seems fine. It's not an exposed implementation detail, it's part of the 4.x API.

PR #1081, opened 2017-02-11
Docs refactoring + code reusability recipes

Hotell:


docs/recipes/README.md
```diff
@@ -1,342 +1,4 @@
...
-      skate.h('style', '.my-class { display: block; }'),
-      skate.h('div', { class: 'my-class' }),
-    ];
-  }
-});
-```
-
-If you want to ensure your styles are encapsulated even if using a polyfill, you can use something like [CSS Modules](https://github.com/css-modules/css-modules).
+These are some use cases and code snippets to get you started with Web Components and Skate in a real app.
+They assume you understand what WebComponents are and [basic skate API knowledge](/docs/api).
```
treshugart: WebComponents -> Web Components

docs/recipes/README.md
```diff
@@ -1,342 +1,4 @@
...
-    return [
-      skate.h('style', '.my-class { display: block; }'),
-      skate.h('div', { class: 'my-class' }),
-    ];
-  }
-});
-```
-
-If you want to ensure your styles are encapsulated even if using a polyfill, you can use something like [CSS Modules](https://github.com/css-modules/css-modules).
+These are some use cases and code snippets to get you started with Web Components and Skate in a real app.
```
treshugart: Maybe just: "with Web Components and Skate" -> "with Skate"

docs/recipes/design-patterns/README.md
```diff
@@ -0,0 +1,5 @@
+# Design Patterns
+
+Every Library or technology has it's desing patterns.
```
treshugart: it's -> its

docs/recipes/design-patterns/README.md
```diff
@@ -0,0 +1,5 @@
...
+
+This chapter describes common patterns used in web components development via Skate
```
treshugart: "used in web components development" -> "used in Web Component development"

docs/recipes/design-patterns/component-composition.md
```diff
@@ -0,0 +1,133 @@
+# Component composition/extending
+
+Since Skate is already close to the platform, you can easily extend other component classes in 2 ways:
```
treshugart: Maybe remove "in 2 ways", that way it's easier to add without updating `2`, it also feels kinda unnecessary to say.

docs/recipes/design-patterns/component-composition.md
```diff
@@ -0,0 +1,133 @@
...
+### Extending Props API via inheritance
+
+Let's say you want to inherit from some component which already defines some common `props`.
+
+**Solution is following:**
+
+```js
+// some common component
+class Colors extends Component {
+   static get props () {
```
treshugart: You use a mixture of  space indentations in this area, but used 2 above. I believe most of the code examples use 2.

docs/recipes/design-patterns/component-composition.md
```diff
@@ -0,0 +1,133 @@
+# Component composition/extending
+
+Since Skate is already close to the platform, you can easily extend other component classes in 2 ways:
+
+- via [Inheritance](#inheritance)
+- via [Mixins](#mixins)
```
treshugart: It might be nice to add decorators to this list at some point. Doing so would mean we need to update this section's title and the description under it, too.

docs/recipes/design-patterns/component-composition.md
```diff
@@ -0,0 +1,133 @@
...
+}
+
+customElements.define('super-component', SuperComponent);
+```
+
+### Extending Props API via inheritance
+
+Let's say you want to inherit from some component which already defines some common `props`.
+
+**Solution is following:**
```
treshugart: Did you mean to say "the solution is as follows"? Maybe saying this isn't necessary as it's already implied?

docs/recipes/design-patterns/component-composition.md
```diff
@@ -0,0 +1,133 @@
...
+}
+
+customElements.define('super-component', SuperComponent);
+```
+
+
+### Extending Props API via mixin
+
+
+**Solution is following:**
```
treshugart: See comment above for the other one.

docs/recipes/design-patterns/component-composition.md
Hotell: hmm right, we can address that in another PR

docs/recipes/design-patterns/component-composition.md
Hotell: yeah, that's how it looks like when code is written in "plain text" :D

Hotell:
pls re review, thx!

PR #1085, opened 2017-02-19
fix(ts-typings): add missing api and fix LC hooks

### LINKED ISSUE #1084: fix(ts-typings): expose PropOptions so they can be consumed by 3rd party decorators
User needs `PropOption` to be able to be consistent with skate Typings and api without rewriting it itself.

Usage:

- https://github.com/wc-catalogue/blaze-elements/blob/mh/mixins-with-new-ts/packages/_helpers/decorators.ts#L1
- https://github.com/wc-catalogue/blaze-elements/blob/mh/mixins-with-new-ts/packages/_helpers/decorators.ts#L33

Additional things to fix:
- `is` static member declaration missing from Component
- all Life cycle methods on `Component` cannot be optional ( they are always there )

Hotell:
Closes #1084

### DIFF
--- a/src/ts-typings/api.d.ts
+++ b/src/ts-typings/api.d.ts
@@ -26,6 +26,7 @@ export class Component<Props> extends HTMLElement {
   // this is not possible yet? ... without this we have to duplicate props definition with class props definition
   // [K in keyof Props]: Props[K],
 
+  static readonly is: string;
   static readonly props: ComponentProps<any, any>;
   static readonly observedAttributes: string[];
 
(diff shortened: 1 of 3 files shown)


src/ts-typings/api.d.ts
```diff
@@ -26,19 +26,20 @@ export class Component<Props> extends HTMLElement {
...
+  static readonly is: string;
   static readonly props: ComponentProps<any, any>;
   static readonly observedAttributes: string[];
 
   // Custom Elements v1
   connectedCallback(): void;
   disconnectedCallback(): void;
   attributeChangedCallback(name: string, oldValue: null | string, newValue: null | string): void;
-  adoptedCallback?(): void;
+  adoptedCallback(): void;
```
treshugart: Totally forgot this was a thing!

Issue #991, opened 2017-01-03
Remove "initial" option from props


treshugart:
When the attribute manager was added it seems that the `default` option broke and now *always* sets the corresponding attribute even if the incoming property value is equal to the default value. I think this is okay moving forward, but since this is the case, the `initial` property is now redundant. Moving forward if you want to set an initial prop, you can do it in the constructor.

The original behaviour was always difficult because if you have default be a function, and it returns an array, the default value will always be different, thus set as an attribute.

Hotell:
finally this will become a thing, high 5 for TS users! :)
```ts
@prop({
  type: Number
}) count = 20;
```

treshugart:
Done in #1098.

PR #1098, opened 2017-03-13
#991 - Remove prop initial option

treshugart:
BREAKING CHANGE: remove prop initial option #991.

Some other related and unrelated changes based off this PR:

- [x] Rename `Mixins` namespace to just functions prefixed with `with*`
- [x] Rename `prop` namespace to just functions prefixed with `prop*`
- [x] Rename `props()` function to `getProps()` and `setProps()`
- [x] Refactor all tests
- [x] Remove redundant utils and functionally separate them in to different util modules
- [x] Keep similar utils in same module file, just export them as functions / consts


test/unit/props.spec.js
```diff
@@ -17,20 +17,16 @@ describe('api/props', () => {
...
-          }
+          prop1: null,
+          prop2: null,
+          prop3: null
         };
       }
       constructor () {
         super();
         this._rendered = 0;
+        this.prop1 = 'test1';
```
Hotell: maybe can we one of the props move as property initialisers? I know that this is what it does under the hood ( babel / TS ) but just to have it in test as documentation reference would be nice.

```js
elem = new (define(class extends Mixins.Props() {

 prop2 = 'test2';

 constructor(){
   super();
   this._rendered = 0;
   this.prop1 = 'test1'
 }

})

```

src/with-props.js
```diff
@@ -0,0 +1,176 @@
...
+      return { ...super.props, ...this[_props] };
+    }
+
+    static set props (value) {
+      this[_props] = value;
+    }
+
+    constructor () {
+      super();
+      const { constructor } = this;
```
Hotell: so correct me if I understood this correctly ( I'm asking because of setting "default/initial" values of property via Class property initialisers )

if I have following component:

```js
class MyCmp extends Component {
  static get props(){
     return {
        name: prop.string,
        year: prop.number,
     }
  }

  name = 'Rodney Mullen';
  year = 1998;

  renderCallback({name,age}){
    return (
      <p>First DarkSlide was done by <b>{name}</b> in <b>{age}</b></p>
    )
  }
  
}
```

At first getter/setter will be created from props on `this`
-> defineProps(constructor);
after that `name = 'Rodney Mullen';` ->transpiledTo-> `this.name = 'Rodney Mullen';` will properly invoke defined prop setter (`name`) and within it it will store the value to internal `_value` 

-> 
```
 set (val) {
 +        this[_value] = coerce(val);
 +        syncPropertyToAttribute(this, target, serialize, val);
 +        this[_updateDebounced]();
 +      }
```

src/with-props.js
```diff
@@ -0,0 +1,176 @@
...
+});
+
+export function getProps (elem) {
+  return keys(elem.constructor.props).reduce((prev, curr) => {
+    prev[curr] = elem[curr];
+    return prev;
+  }, {});
+}
+
+export function setProps (elem, props) {
```
treshugart: :)

src/index.js
```diff
@@ -1,20 +1,7 @@
...
-  Component,
-  define,
-  emit,
-  h,
-  link,
-  Mixins,
-  prop,
-  props
-};
+export * from './define';
```
treshugart: :D

src/define.js
```diff
@@ -19,7 +18,7 @@ export default function (Ctor) {
     // If we used defineProperty() then the consumer must also use it and
     // cannot use property initialisers. Instead we just set it so they can
     // use whatever method of overridding that they want.
-    Ctor.is = uniqueId();
+    Ctor.is = `x-${uniqueId()}`;
```
treshugart: Sort of. `prefix` within `uniqueId()` tightly coupled it to this use case. Didn't want that as it's also used in the `sym` function now.

package.json
```diff
@@ -27,6 +27,7 @@
   ],
   "dependencies": {},
   "devDependencies": {
+    "@skatejs/web-components": "^1.0.0",
```
Hotell: oh we have a npm namespace ( finally yay! ) ? good to know ;)

package.json
```diff
@@ -35,15 +36,14 @@
...
+    "preact": "*",
     "semantic-release": "^6.3.2",
     "semistandard": "^9.0.0",
-    "skatejs-web-components": "5.2.1",
     "typescript": "^2.1.4",
     "typescript-formatter": "^5.0.0"
   },
   "peerDependencies": {
-    "preact": "^7.2.0"
+    "preact": "*"
```
Hotell: I know @developit mentioned that he is doing it like this because he run into some issues. 

In my experience this could be dangerous, because with this you say to the user, use whatever version you like, but skate may not work with every possible version of Preact right ? I think `^7.2.0` is right way ( any feat.patch of Preact 7 )

src/with-props.js
```diff
@@ -0,0 +1,182 @@
...
+      return { ...super.props, ...this[_props] };
+    }
+
+    static set props (value) {
+      this[_props] = value;
+    }
+
+    constructor () {
+      super();
+      if (this[_constructed]) return;
```
Hotell: what's the use-case for this check?

package.json
developit: Preact 8 will be released very shortly, which might cause people some pain - the issue is that peer deps don't update when you bump their versions in this package.json.  So the user will update Skate and then get an error saying it's incompatible.

src/with-props.js
treshugart: There's a comment where it's defined with the other flag at the top. Polyfill pains again.

package.json
treshugart: Cool, thanks for that. There's such a minimal API surface that we actually use (we really only just render and directly export h) that I'm not too worried about it at this point. If there's issues we can revisit without affecting our API, most likely.

Issue #1000, opened 2017-01-05
Make props async and add ability to pass function to props()


treshugart:
Currently props is synchronous but it'd be good if it had the same semantics as setting props normally and allowed a function to be passed to it so that you could calculate your next set of props from the previous one in a predictable manner, like in React.

The previous prop values are recorded using a symbol, so those props can be retrieved and passed into the user-defined function as the previous props.

See https://medium.com/@shopsifter/using-a-function-in-setstate-instead-of-an-object-1f5cfd6e55d1#.bfgyrbjud for more information for those who may not be aware of the pitfalls of async rendering and calculate next props.

Hotell:
Why is this needed? I mean setState in React is async, props() in skate is not AFAIK?

treshugart:
@Hotell updated the issue.

Setting props normally is async and `props()` is sync. So we have this problem (with no way to get the previous props) when setting properties normally, forcing the component author to have to render sync, which isn't ideal if you have multiple calls to it. Additionally, I think it'd be better for them to have the same semantics, thus needing a standard way to retrieve the previous state.

Hotell:
thanks for clarification. yup totally makes sense 🍺

alexlafroscia:
Is the plan to support either an object _or_ a function? Or only a function?

treshugart:
The PR changed quite a bit. In retrospect, there's no foreseeable reason to support a function as property values reflect immediately, renders are queued. So no need for a function. There's now a `props` getter / setter. See the PR for more details.

PR #1100, opened 2017-03-15
#1000 - add ability to pass function to setProps for prev props

treshugart:
Implements #1000.


src/with-props.js
```diff
@@ -178,5 +181,8 @@ export function getProps (elem) {
 }
 
 export function setProps (elem, props) {
+  if (typeof props === 'function') {
```
Hotell: we should have for these cases general facade-lang.js file

`typeof props === 'function'` => `isFunction(props)` etc

treshugart:
@Hotell maybe it'd be good to so the `set*` on the instance now?

Hotell:
hmm so I'm not 100% on board with this implementation/feature

- accessing just props via `this.props` is OK ( the same as react/preact does )
- setting props async via `this.props = {}` or `this.props = updateMe()` feels very strange and confusing. as I've said I will rather preffer API constraints/design which is battle-proofed so rather `setProps(props: object | Function )`
- this may overall add confusion to consumer:
  -- defining props for component is done via static props ( kind of like React does for propTypes )
  -- setting prop value directly via `this.foo="hello"` is sync
  -- setting prop value via `this.prop`.foo="hello" is async
  -- setting via `this.props` is possible only via internal component calls not from outside 
  -- from typed JS view - user has to provide definitions for all props via generics and then again annotate class instance properties

```ts
type Props = {
 test: string
}
class MyComponent extends Component<Props> {
  // Defines the props.
  static props = {
    test: propString
  }

  // inferred as from generic `<Props>` definition ( default props kinda style, but not really )
  props = {
    test: 'some value'
  }
  
  // again I have to define and annotate class prop on instance
  test: string;

}

customElements.define('x-test', MyComponent);

////

const elem = new MyComponent();

// { test: 'some value' }
elem.props = { test: 'another value' };

// 'another value'
elem.test;

// 'another value'
elem.test = 'another value again' ;

// 'another value again'
elem.test;
elem.props.test;
```

Hotell:
Hmm so I was thinking a little today and actually having props on `this.props` is fantastic ( as a author of a component (with typed JS) I dont have to specify those props again on instance, because from my perspective I will always operate with `this.props` and `this.setProps()`. If someone want's to consume the component by manually instantiating it and setting props directly ( he wont get support for autocomplete from type definition ), I'm perfectly fine with that use case.

So yeah I just vote `props` setter to `setProps()` instance method change, otherwise LGTM

treshugart:
I feel having a `setProps()` and a `props` getter is super confusing. In fact, I'm not even sure that having a `setProps` is necessary when you can just `Object.assign()`. We also don't have the same problems as React where prev props might be different at the time of setting state. Even though setting props from any angle is async in Skate, the prop values reflect immediately. I'd be fine having just a `props` getter on the `prototype`, I think.

treshugart:
> setting prop value directly via this.foo="hello" is sync

That's actually also async. The value is reflected immediately, but a render is still queued.

treshugart:
Some changes in this PR:

- Subclass `static props` do not automatically return parent class definitions if the subclass has overridden them. This is because there's no way to undo the merging, for example if you only want a subset, not a superset. It's also more explicit.
- Instance `props` setter (old `props(object)`) only affects declared props. Undeclared props are not set anymore. To do this use `Object.assign(host)`.
- Instance `props` setter is always async, just like normal prop setting.
- Property values *always* propagate immediately rather than in batches, renders only happen in batches. This prevents pitfalls where you might want to use a previous value after setting it to affect a future value. This could be happening in different parts of the application. In React, you'd use `setState(callback)` in order to do this.

PR #1101, opened 2017-03-16
4. docs: Update docs for 5.x changes.

treshugart:


matthewp:
These have been useful for me! Are they ready to merge into the 5.x branch?

freeman:
@treshugart The result of merging this branch is not working because the branch 5.x-docs was still using `propString` but the `5.x` was already on `props.string`. 

So now in the 5.x branch you have type definitions for the wrong way. 

If you're OK, I'll make a PR this week-end to fix the definitions temporarily while waiting for the full fix from @Hotell

treshugart:
@freeman yeah, that'd be great if it doesn't take too much time so that we can have a working 5.x branch. I have plans to autogenerate TS defs from Flow sources, which means quite possibly we'll move to using Flow and not have to maintain the TS defs. I want to wait a couple weeks to see where the generator is at because it would be nice to be able to make this change as part of 5.x.

freeman:
Ok, thanks, expect my PR soon.

I have seen your branch about Flow. 
Have you considered using typescript instead of Flow ? (Sorry if it has already been debated to death...)

I think that it enables easier contribution (flow is still kind of a pain with windows) and with the 2.3 release, we could migrate very progressively with the checkJs flag ( see https://blogs.msdn.microsoft.com/typescript/2017/04/27/announcing-typescript-2-3/ )

treshugart:
Yeah, I've seen that. As it stands, I'm going to be a little opinionated
about it because there is going to be a tool made available soon to output
both TS and Flow definitions from Flow source. Consumers > contributors,
IMO.

Once Babylon gets TS support via a plugin, using either will be much more
viable. But until then, I don't want to be locked into the TS featureset,
compiler and only able to export TS types.

I don't want this to spark any further discussion, so I'm just going to say
that, for now, Flow will probably be the direction and I'm happy to be
blamed for that as to curb any possible 🔥 wars that come from it :p

(Thanks for asking, though!)

PR #1104, opened 2017-03-22
fix node refresh

juicelink:
when the web component rerender after propertiies changes, the new node is appended to the shadow root but the old ones are not removed. (same result cross browser)

I don't know if it is the right fix but this one seems to work


treshugart:
Yeah, I think this is necessary for it to avoid appending. Thanks!

Issue #1102, opened 2017-03-16
refactor(ts-typings): update definitions to match skate 5.x


Hotell:
5.x introduces BC to the API

@treshugart already updated ts-definitions, but we should review if everything is up to date

Also Preact definitions needs some love before we can proceed
- issue https://github.com/developit/preact/issues/639
- depends on https://github.com/developit/preact/pull/621

treshugart:
I don't think I added them for the `when*` mixins or separation of the `props()` function into `setProps()` and `getProps()`.

freeman:
Also missing: prop.string, prop.number, etc. should become propString, propNumber, etc.

Hotell:
yeah that's gonna actually change in #1100

treshugart:
Yeah, `using props.*` (plural) actually makes more sense especially since the convenience of importing a single namespace outweighs the benefits of tree-shaking (object properties don't shake out).

freeman:
I've seen the fix in #1100. Thanks.

Also now that skate is using preact, shouldn't it rely on preact's type definitions for the JSX part instead of redeclaring it ?

Hotell:
That will need to be refactored. Preact typings are not in very good shape at the monent. I'll work on it soon

treshugart:
> Also now that skate is using preact, shouldn't it rely on preact's type definitions for the JSX part instead of redeclaring it ?

Yeah, I think that's probably a good idea. Would this be possible if forwarding the export from `with-props.js` like we're doing now?

Hotell:
> Also now that skate is using preact, shouldn't it rely on preact's type definitions for the JSX part instead of redeclaring it ?

NOPE. as I've said:

> That will need to be refactored. Preact typings are not in very good shape at the monent. I'll work on it soon

treshugart:
Ahh right. At least that will benefit Preact!

freeman:
@treshugart are you confident on the automated generation of TS typings from the flow source ? 

Or should we still invest in making the manual TS definitions complete ? (some thing are missing like the `prop` function and `props.any` for which I can make a PR)

treshugart:
@freeman I would still invest in making the TS definitions complete. There's still some work that needs to be done in the converter.

Hotell:
[...]

PR #1105, opened 2017-03-22
Typescript defs

juicelink:
typescript definitions enter in conlict with preact JSX definitions => removed JSX typescrript definitions in skatejs and keeped only the api definitions.
- Added prop types (propString, etc...) definitions.
- Removed not implemented anymore definitions. 
- Added getProps and setProps, I'm not sure if I had to.
- Re-exported h from preact in index.d.ts


treshugart:
These will change slightly, but perfect for now until #1000 is merged. Thanks!

Hotell:
this is very unfortunate change. Please don't "hurry up" things next time. Preact JSX definitions are in bad shape, which I addressed in https://github.com/skatejs/skatejs/issues/1102

So this "rapid" deletion was really unnecessary

Hotell:
also commenting code and leaving it there is very nasty thing 
-> `// return <div statics={['attr1', 'prop2']}></div>`

treshugart:
Yep, we need to do a big sweep on this anyways. I'm less worried about getting this in than getting everything to 5.x for a TS review then a large review.

freeman:
Do you have any roadmap on when you'd like to release skatejs 5 ? 
We're using it now in a PoC but obvisouly a stable release would be better as we scale the project.

treshugart:
@freeman I'm hoping within the next couple of weeks.

PR #1108, opened 2017-03-24
Move unique tag name generation to a mixin, instead of being done in define()

treshugart:
## Todo

Still doesn't work for SSR; names would be different across server / client. I'm not sure there's a way to do this without having hardcoded names because you need to account for out-of-order loading. Names must be 100% deterministic.

## About the change

Instead of a unique tag name being generated for a class in `define()`, it's now done in a mixin called `withUnique()` which defines a static getter (and setter so it can be overridden) that will return a unique tag name, unless it's overridden.

As with other mixins, `withUnique()` extends `HTMLElement` by default, but you can pass it an optional base class if you want to.

Something to note about the new way is that there's no side effects on your class. Before it would define an `is` class property if one didn't exist. Now it's explicit.

Previously you'd do:

```js
import { define } from 'skatejs';

export default define(class extends HTMLElement {});
```

Now you'd do:

```js
import { define, withUnique } from 'skatejs';

export default define(class extends withUnique() {});
```

## About the PR in general

Most of the PR is just updates to tests.


src/with-unique.js
```diff
@@ -0,0 +1,14 @@
+import { HTMLElement, sym, uniqueId } from './util';
+
+const _is = sym();
+
+export function withUnique (Base = HTMLElement) {
+  return class extends Base {
+    static get is () {
+      return this[_is] || (this[_is] = `x-${uniqueId().substring(0, 8)}`);
```
Hotell: just a nitpick, but there is "too much going on" in that getter.

Maybe extracting ``x-${uniqueId().substring(0, 8)}`` to `generateUniqueName()` function would be more readable 

```js
// util.js

export function generateUniqueName(prefix='x'){
  return `${prefix}-${uniqueId().substring(0, 8)}`
}


// with-unique.js

static get is () {
  return this[_is] || (this[_is] = getUniqueName())
}

```

WDYT?

src/with-component.js
```diff
@@ -2,9 +2,10 @@ import { render } from 'preact';
 import { HTMLElement } from './util';
 import { withProps } from './with-props';
 import { withRender } from './with-render';
+import { withUnique } from './with-unique';
 
 export function withComponent (Base = HTMLElement) {
-  return class extends withRender(withProps(Base)) {
+  return class extends withUnique(withRender(withProps(Base))) {
```
Hotell: so now user created components will get 4 levels of prototype chain by default. hmm :D let's see how this goes from performance perspective

```
UserClassComponent
  --> uniqueClass
    --> renderClass
      --> propsClass
        --> HTMLElement
```

Let's see how this goes with perf as we discussed already. 
Just thinking what could we use if we ran into perf problems? I guess just directly mutating user `UserClassComponent` class?

src/with-unique.js
```diff
@@ -0,0 +1,56 @@
...
+    define(constructor);
+    return Base.call(this);
+  }
+  Object.defineProperty(Ctor, 'is', {
+    get () {
+      return this[_is] || (this[_is] = generateName(this));
+    },
+    set (is) {
+      this[_is] = generateName(this, is);
+      define(this);
```
treshugart: We define on the setting of `static is` so that the user can define a name, or at least hint at a name prefix (if there are more than one it's deduped). The user would need to do this if they are going to SSR, or are exposing their component as an HTML API. No way to auto-define by just defining a class. Needs an action such as setting to implicitly invoke.

src/with-unique.js
```diff
@@ -0,0 +1,56 @@
...
+    suffix++;
+  }
+  return formatName(prefix, suffix++);
+}
+
+export function withUnique (Base = HTMLElement) {
+  function Ctor () {
+    const { constructor } = this;
+    define(constructor);
+    return Base.call(this);
```
treshugart: I think if using native ES2015, we'll need to somehow detect that and use `Reflect.construct()` at that point. However, this works now, and we can revisit that later, I think.

src/with-unique.js
```diff
@@ -0,0 +1,56 @@
+import { customElements, dashCase, HTMLElement, keys, sym } from './util';
```
treshugart: All this code because we can't access the constructor in an ES5 way before calling `super()`. `new.target` works, but that only works in ES2015. All we need to do is `this.constructor`, but nooooo.

src/with-unique.js
```diff
@@ -1,14 +1,56 @@
...
+function generateName (Ctor, hint) {
+  const prefix = hint || dashCase(Ctor.name) || 'element';
+  while (customElements.get(formatName(prefix, suffix))) {
+    suffix++;
+  }
+  return formatName(prefix, suffix++);
+}
+
+export function withUnique (Base = HTMLElement) {
+  function Ctor () {
```
Hotell: any particular reason why are you suddenly switching to "low level / es5" api ?

src/with-unique.js
```diff
@@ -1,14 +1,56 @@
...
 const _is = sym();
+const baseFunctionStatics = keys(function () {});
 
-export function withUnique (Base = HTMLElement) {
-  return class extends Base {
-    static get is () {
-      return this[_is] || (this[_is] = `x-${uniqueId().substring(0, 8)}`);
+let suffix = 0;
+
+function extendStatics (Ctor, Base) {
```
Hotell: this and `formatName`, `generateName` should be in util.js

src/with-unique.js
Hotell: also if there is a valid reason I guess we should be consistent and apply this on all mixins

src/with-unique.js
```diff
@@ -1,14 +1,56 @@
...
+  while (customElements.get(formatName(prefix, suffix))) {
+    suffix++;
+  }
+  return formatName(prefix, suffix++);
+}
+
+export function withUnique (Base = HTMLElement) {
+  function Ctor () {
+    const { constructor } = this;
+    define(constructor);
```
Hotell: this could be an issue for us at `blaze-elements` as we register classes to WC container only in root exports, otherwise we are using plain classes to be able to redefine them and not pollute customElements namespace registry

*example:*

**Button**

https://github.com/wc-catalogue/blaze-elements/blob/master/packages/button/index.ts#L5

-> user can:
`import {Button} from '@blaze-elements/button'`

which will give him registered WC element so he can use `<bl-button>click me </bl-button>`

but if he want's to register it via custom name, he can do:
`import Button from '@blaze-elements/button/Button'` , which will give him raw class, so he needs to do `customElements.define('my-button',Button)` -> `<my-button>I own you</my-button>`

this technique is further applied across other packages that are using existing WC classes like here 
https://github.com/wc-catalogue/blaze-elements/blob/master/packages/calendar/components/Button.tsx
-> this makes every package really atomic and self contained

freeman: I agree that that Components should not be auto `define`d. Letting the component author decide when to do it seems better

src/with-unique.js
treshugart: In order to get the `constructor` reference, we have to get it from `this` and you can't access `this` before calling `super()`, however, we need to `define()` it be fore `HTMLElement`'s constructor can be called.

I don't think consistency for the sake of consistency is good here. This is a valid reason to break away, but there would be a lot of friction in moving away from ES2015 classes.

src/with-unique.js
treshugart: Yeah, I'll do that once the approach is finalised. Still spiking the ideas :)

src/with-component.js
treshugart: Yeah, I'm concerned about that, too. If we just used functions, then they could be used as decorators and not add to the chain. Normal `instanceof` checks don't work with this style of mixins without a [bit of intervention](http://justinfagnani.com/2016/01/07/enhancing-mixins-with-decorator-functions/), anyways.

src/with-unique.js
treshugart: FWIW, consumers can still import and manually define. It will then just bypass auto-defining.

treshugart: I think a good compromise here might be to bake this into `define()`. It's still explicit, but gives you simpler ergonomics from being able to use the class name if `static is` is not defined. Mixins are opt-in anyways.

In fact, if we made everything decorators / functions, then this could be baked in without the hacks:

```js
import { component, define } from 'skatejs';

@component
@define
class HelloWorld extends HTMLElement {}
```

src/with-component.js
treshugart: Did a simple perf test https://jsperf.com/prototype-chains. This basically means that each extension slows down by about 100k ops / sec, in Chrome, on a mid-2015 MacBook Pro.

treshugart: Actually, Canary is acceptable: 
stable
canary

treshugart: I reckon keeping the ergonomics of classes is worth it. Thoughts, @Hotell?

Hotell: hmm well I guess mixins are our only option if we don't wanna directly mutate user defined component class.

From statically typed js point of view -> using decorators wont add props and other methods definition to class type definition

Hotell: My results on 5 year old Alienware

image



image

treshugart: That's not horrible, considering the deltas. It's also Chrome 56 :)

PR #1110, opened 2017-03-26
Fixed multiple definition of properties on the same constructor.

### LINKED ISSUE #1111: lost props when creating multiple instance of a component simultaneously
When creating multiple instance of a given component in 1 shot through this kind of code: 
```js
var host = document.querySelector('#container');
host.innerHTML = '<x-hello-world name="first"></x-hello-world>' +
  '<x-hello-world name="second"></x-hello-world>';
```

Only the second one will render with the right name props. The first one will have an undefined name props during render.

freeman:
fixes #1111 

Properties were defined at each instantiation of a given element. These caused the backing symbol to be changed for each instance but as the properties are defined on the prototype, this meant that all instances apart from the last actually lost track of the props value.

This was visible in usage when instantiating the same component multiple times in a page at the same time (with innerHTML for example). Only the last one actually got access to the props when render time comes (after debounce).

### DIFF
--- a/src/util/with-props.js
+++ b/src/util/with-props.js
@@ -8,9 +8,10 @@ const _syncingPropertyToAttribute = sym('_syncingPropertyToAttribute');
 export const _updateDebounced = sym('_updateDebounced');
 
 export function defineProps (Ctor) {
-  if (!Ctor[_definedProps]) {
-    Ctor[_definedProps] = true;
+  if (Ctor[_definedProps]) {
+    return;
   }
+  Ctor[_definedProps] = true;
 
   const { prototype } = Ctor;
   const props = normPropDefs(Ctor);
--- a/test/unit/props.spec.js
+++ b/test/unit/props.spec.js
@@ -9,9 +9,10 @@ import { define, withProps, withUnique } from 'src';
 
 describe('api/props', () => {
   let elem;
+  let ElemClass;
 
   beforeEach(done => {
-    elem = new (define(class extends withUnique(withProps()) {
+    ElemClass = define(class extends withUnique(withProps()) {
       static get props () {
         return {
           prop1: null,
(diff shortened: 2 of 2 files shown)


Hotell:
Can you pls create an issue before submitting PR and referencing it within PR ?

thanks!

freeman:
sorry about that, done.

freeman:
I see that this PR no longer merges cleanly. Do you want me to rebase it ?

Is there any issue with it that prevents it from being integrated ?

treshugart:
@freeman shouldn't be. Feel free to rebase then we'll merge.

freeman:
rebased onto `5.x`

treshugart:
Cool! @Hotell can you give the thumbs up on this, please?

The build would be failing because it's not coming from the skatejs org. Eventually we'll need to come up with a solution for contribs. Happy to merge if you think it's okay.

src/util/with-props.js
```diff
@@ -8,7 +8,9 @@ const _syncingPropertyToAttribute = sym('_syncingPropertyToAttribute');
 export const _updateDebounced = sym('_updateDebounced');
 
 export function defineProps (Ctor) {
-  if (!Ctor[_definedProps]) {
+  if (Ctor[_definedProps]) {
+    return;
+  } else {
```
Hotell: just a nitpick.

Else is unnecessary and makes code less readable

consider:
```js
export function defineProps (Ctor) {
  if (Ctor[_definedProps]) {
    return
  }

  Ctor[_definedProps] = true;

  const { prototype } = Ctor;
  const props = normPropDefs(Ctor);

// ...

}

freeman:
@Hotell absolutely clearer this way. Just pushed an amended version, thanks.

PR #1113, opened 2017-03-28
1. Streamline render

treshugart:
The goal of this is to make the renderer API as simple as possible, but also as extensible as possible for consumers when extending `withRender()`. Edge cases can just extend `withProps()` and use `propsUpdatedCallback()`.

### DIFF
--- a/src/with-component.js
+++ b/src/with-component.js
@@ -6,9 +6,8 @@ import { withRender } from './with-render';
 import { withUnique } from './with-unique';
 
 export const withComponent = (Base = HTMLElement) => class extends withUnique(withRender(withProps(Base))) {
-  propsChangedCallback (next, prev) {
-    super.propsChangedCallback(next, prev);
-    this.rendererCallback(render);
+  rendererCallback (shadowRoot, renderCallback) {
+    render(renderCallback(), shadowRoot);
   }
 };
 
--- a/src/with-render.js
+++ b/src/with-render.js
@@ -1,18 +1,18 @@
 import { HTMLElement } from './util';
 
 export const withRender = (Base = HTMLElement) => class extends Base {
+  propsUpdatedCallback (next, prev) {
+    super.propsUpdatedCallback(next, prev);
+    if (!this.shadowRoot) {
+      this.attachShadow({ mode: 'open' });
+    }
+    this.rendererCallback(this.shadowRoot, () => this.renderCallback(next));
+    this.renderedCallback();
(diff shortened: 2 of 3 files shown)


src/with-component.js
```diff
@@ -6,9 +6,8 @@ import { withRender } from './with-render';
 import { withUnique } from './with-unique';
 
 export const withComponent = (Base = HTMLElement) => class extends withUnique(withRender(withProps(Base))) {
-  propsChangedCallback (next, prev) {
-    super.propsChangedCallback(next, prev);
-    this.rendererCallback(render);
+  rendererCallback (shadowRoot, renderCallback) {
+    render(renderCallback(), shadowRoot);
```
treshugart: This makes it so they can decide when to "render", but also not have to worry about the function signature, making it consistent across renderers.

src/with-render.js
```diff
@@ -1,18 +1,18 @@
...
-
-  // Called to render the component.
-  rendererCallback (render) {
+  propsUpdatedCallback (next, prev) {
+    super.propsUpdatedCallback(next, prev);
     if (!this.shadowRoot) {
       this.attachShadow({ mode: 'open' });
     }
-    render(this.renderCallback(this), this.shadowRoot);
+    this.rendererCallback(this.shadowRoot, () => this.renderCallback(next));
```
treshugart: A possible issue here is creating an extra function every time we render. I'm not convinced there's actually an issue here. Incremental DOM uses this method in its examples.

src/with-render.js
```diff
@@ -1,18 +1,18 @@
...
-
-  // Called to render the component.
-  rendererCallback (render) {
+  propsUpdatedCallback (next, prev) {
+    super.propsUpdatedCallback(next, prev);
     if (!this.shadowRoot) {
       this.attachShadow({ mode: 'open' });
     }
-    render(this.renderCallback(this), this.shadowRoot);
+    this.rendererCallback(this.shadowRoot, () => this.renderCallback(this));
```
treshugart: A possible issue here is creating an extra function every time we render. I'm not convinced there's actually an issue here. Incremental DOM uses this method in its examples.

freeman:
Beware this still has the issue mentioned in #1104 (which was only merged in `5.x-docs`).

treshugart:
Good point, thanks @freeman. I think we'll merge that soon to get these aligned then begin on docs again.

PR #1115, opened 2017-03-28
2. Support no shadow dom

treshugart:


Hotell:
yesssssss!

treshugart:
@Hotell do you think this is the best API for this? I just figured I'd reuse the existing `attachShadow` interface. I didn't initially think there'd be any implications for typed interfaces here, but thought I'd just check anyways. I did feel that overloading the built-in might be odd? Maybe a `static useShadowRoot` option might be better upon a second take. Thoughts?

Hotell:
you're absolutely right! overloading native api is definitely not good idea. maybe we can encapsulate everything to a `static get config(){ return {useShadowDOM: false } }`

treshugart:
I'm not sure I'm a fan of having a generic `config` object. I know Polymer does this but it would make inheritance more of a pain and then not be inline with the current native convention of `observedAttributes`. Would we put `props` in here? Should we shoe-horn `observedAttributes` into it? I think just using a `static` prop keeps it simple.

Hotell:
yeah, deal ;)

treshugart:
@Hotell Cool, I'll make that change. Proposed name is `useShadowRoot`.

Hotell:
hmm, that name is fine. I would propose `useShadowDOM` because if it's false yeah there won't be any shadow root --> so no shadowDOM encapsulation.

Also when turned off, projection wont work right?


```
class Greeting extends Component {
  static is = 'my-greeting';
  static useShadowRoot = false;
  renderCallback(){
    return <div>Hello <slot/></div>
  }
}
```

```html
<my-greeting>World!</my-greeting>
```

will not produce  -->
```
<my-greeting><div>Hello World</div></my-greeting>
```

but
```
<my-greeting><div>Hello </div></my-greeting>
```

treshugart:
That's correct, projection will be off. The consumer will have to use alternate methods of passing data. I think this is okay, though.

Hotell:
right, so what's the real world use case for this?
- server side rendering?

I know it also adds "benefit" of making less polyfill overhead in non native WC browsers and removes style encapsulation ofc ( which may be desirable in some usecases - using css frameworks in efficient way for instance )

treshugart:
I'd say that's a real world use case.

- [todo] I want to see if it's possible to override `children` and have Preact pass that around. That would be cool!
- Another is when you just want to have leaf nodes. You can disable shadow DOM completely and have consistent styling behaviour (this was why I originally wanted to use `attachShadow`) rather than relying on tons of extra checks.

thepian:
The way I implemented this included the ability for the element to specify the root element, so here it would be `this`, for shadow DOM it would be the shadow root. However I think you might have elements with a wrapper structure where the dynamic render should be within a child rather than the element itself.

treshugart:
@thepian that's a good point. If implementing a custom shadow root was as simple as overriding `shadowRoot`, that might be just fine. Maybe the best thing to do is to, internally in Skate, use the return value of `attachShadow()` so that it can be overridden to provide a custom shadow root. This means all Skate has to check for is `attachShadow` and whether or not it's been called already.

treshugart:
On second inspection, though, that's already what you can do here: override `shadowRoot`, you don't even have to worry about `attachShadow`. I'll write at est for that.

treshugart:
Ok, so I've updated to now use the `shadowRoot` property anymore. It will still exist because we're calling `attachShadow({ mode: 'open' })`, but it's no longer coupled API. We now use the return value of `attachShadow()`. I've written tests to ensure that this is strictly followed. The nice part about this is that you can override attachShadow() to provide your own shadow root, or use closed shadow roots if we enable that at some point, while still defaulting to `open` without changing the API.

PR #1135, opened 2017-04-14
Call renderer when props change

### LINKED ISSUE #1134: Doesn't rerender when props change
I think [this line](https://github.com/skatejs/skatejs/blob/dac2c4dff7824221aebf3bf55c402c04b0ad8674/src/with-props.js#L92) is wrong.  It should be `!keys(prev).every(k => prev[k] === next[k])`.

Going to try my hand at writing a test for this :D

matthewp:
This fixes #1134.

When properties change, the renderer function is called. This is done by implementing propsChangedCallback within with-render.

Also fixes an issue where any renders after the first would not result in the shadowRoot being updated, because we weren't passing the preact component back into preact's `render` method.

### DIFF
--- a/src/with-component.js
+++ b/src/with-component.js
@@ -1,4 +1,4 @@
-import { h, render as preactRender } from 'preact';
+import { h, render } from 'preact';
 
 import { HTMLElement, sym } from './util';
 import { withProps } from './with-props';
@@ -7,16 +7,12 @@ import { withUnique } from './with-unique';
 
 const _preactDom = sym('_preactDom');
 
-function render (vnode, root) {
-  const dom = preactRender(vnode, root, this[_preactDom]);
-  if (!this[_preactDom]) {
-    this[_preactDom] = dom;
-  }
-}
-
 export const withComponent = (Base = HTMLElement) => class extends withUnique(withRender(withProps(Base))) {
   rendererCallback (shadowRoot, renderCallback) {
-    render.call(this, renderCallback(), shadowRoot);
+    const dom = render(renderCallback(), shadowRoot, this[_preactDom]);
+    if (!this[_preactDom]) {
+      this[_preactDom] = dom;
+    }
   }
 };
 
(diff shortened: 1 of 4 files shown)


src/with-component.js
```diff
@@ -1,13 +1,18 @@
...
 import { withRender } from './with-render';
 import { withUnique } from './with-unique';
 
+const _preactDom = sym('_preactDom');
+
 export const withComponent = (Base = HTMLElement) => class extends withUnique(withRender(withProps(Base))) {
   rendererCallback (shadowRoot, renderCallback) {
-    render(renderCallback(), shadowRoot);
+    const dom = render(renderCallback(), shadowRoot, this[_preactDom]);
+    if (!this[_preactDom]) {
```
treshugart: I wonder if you even have to have this check? Could you just do something like:

```js
this[_preactDom] = render(renderCallback(), shadowRoot, this[_preactDom]);
```

matthewp: I always have avoided that out of perf fears. Is setting a property over-and-over again like that basically ignored when you're setting it to the same value over time?  Or is the branching caused by a conditional worse?

treshugart: Given you can get tens of millions of operations per second setting a property, I don't feel that it would matter too much. Happy to keep as is if you feel it's better. I think the Preact MO is just starting to rub off on me and I was trying to save a few bytes :)

treshugart:
Do whatever you think is best. Merge when you feel comfortable.

src/with-component.js
matthewp: That's fair too, I hadn't thought about size. I have no problem changing it. I am curious which is better thought, probably you're right that the branching is worse. I don't know why my instinct is to add the diff.

matthewp:
I did what they tell you to do when facing a hard comp sci problem; ask the experts on twitter: https://twitter.com/matthewcp/status/852944353911853058

Storing the value sounds fine, so went with that for the size savings. Will merge when tests pass.

PR #1138, opened 2017-04-15
Add version field to package.json

matthewp:
Yarn complains if a package doesn't have a version field, even when pulled from GitHub. This adds one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "skatejs",
+  "version": "5.0.0-alpha.0",
   "description": "Skate is a library built on top of the W3C web component specs that enables you to write functional and performant web components with a very small footprint.",
   "license": "MIT",
   "author": "Trey Shugart <treshugart@gmail.com> (http://treshugart.github.io)",


matthewp:
Think this is non-controversial.

Hotell:
Hmm what about that beta release to npm @treshugart this feels dirty to me :-|

matthewp:
This was just to get yarn to not complain when using the GitHub branch. I didn't publish anything :)

matthewp:
Why do you think this is dirty?

Hotell:
yup I know why you did that @matthewp. Anyway semantic release automatically bumps version and during maintainers call we discussed this and came to a conclusion that beta release is the best and standard way to go, not installing from github.

matthewp:
If you publish with `--tag beta` then normal installs will get the 4.x releases. This is what I assume you want, no?

Anyways, you're right that installing from git is not the norm, but I have no other choice at this time. I'm happy to release a beta from 5.x if that is what you all want. Is it?

PR #1141, opened 2017-04-19
Allow elements to create their own shadow

### LINKED ISSUE #1140: Shouldn't call attachShadow if the element already has a shadow DOM
I have a custom element that creates shadow roots on behalf of other elements. When one of those other custom elements is a Skate component it will throw because of [this line](https://github.com/skatejs/skatejs/blob/b5b7097b2f2fd5c154fab56841876aae4c87c23a/src/with-render.js#L13).

I think that it should check if `this.shadowRoot` exists already and prefer that over calling attachShadow. Or create a attachShadowIfNeeded function.

Thoughts @treshugart @Hotell ?

matthewp:
This ensures that we do not throw on the first render if the element already has a shadow DOM (such as if the element creates its own shadow in the constructor).

Closes #1140

### DIFF
--- a/src/with-render.js
+++ b/src/with-render.js
@@ -9,9 +9,13 @@ function attachShadow (elem) {
 }
 
 export const withRender = (Base = HTMLElement) => class extends Base {
-  propsChangedCallback () {
+  get renderRoot () {
     this[_shadowRoot] = this[_shadowRoot] || (this[_shadowRoot] = (this.shadowRoot || attachShadow(this)));
-    this.rendererCallback(this[_shadowRoot], () => this.renderCallback(this));
+    return this[_shadowRoot];
+  }
+
+  propsChangedCallback () {
+    this.rendererCallback(this.renderRoot, () => this.renderCallback(this));
     this.renderedCallback();
   }
 
--- a/src/with-render.js
+++ b/src/with-render.js
@@ -10,7 +10,7 @@ function attachShadow (elem) {
 
 export const withRender = (Base = HTMLElement) => class extends Base {
   propsChangedCallback () {
-    this[_shadowRoot] = this[_shadowRoot] || (this[_shadowRoot] = attachShadow(this));
+    this[_shadowRoot] = this[_shadowRoot] || (this[_shadowRoot] = (this.shadowRoot || attachShadow(this)));
     this.rendererCallback(this[_shadowRoot], () => this.renderCallback(this));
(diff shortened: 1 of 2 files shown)


src/with-render.js
```diff
@@ -10,7 +10,7 @@ function attachShadow (elem) {
 
 export const withRender = (Base = HTMLElement) => class extends Base {
   propsChangedCallback () {
-    this[_shadowRoot] = this[_shadowRoot] || (this[_shadowRoot] = attachShadow(this));
+    this[_shadowRoot] = this[_shadowRoot] || (this[_shadowRoot] = (this.shadowRoot || attachShadow(this)));
```
treshugart: If they call `attachShadow({ mode: 'closed' })` this would fail, right? I think the only reliable way to do this would be to override `attachShadow`, but I'm not sure we should go down that route.

treshugart:
Something that I've thought about in the past would be to add a special prop to render to that `withRender` would implement by default. The prop would be responsible for always returning a reliable root, whether it's the shadow root, the host node, or a "portal" (like in React) to somewhere else.

Failing that, the expected behaviour here could be that you can't attach your own shadow with the default implementation.

Do you have a use case that might be relevant?

src/with-render.js
matthewp: Sure, it will fail for closed shadows with or without this change. But since 99% of shadows are open, why not fix it for those?

matthewp:
My use case: I wrote a custom element that swaps out an element's light DOM and sticks it in the shadow DOM, for SSR upgrades: https://github.com/matthewp/aliens-click/blob/a861a1587f45aefbaaa403ec68d59fcf97cecb53/src/window/swap-shadow.js

matthewp:
A callback for getting the root could be ok too. That would allow people to use the light DOM for rendering (I know you don't like that :D).  Maybe `rootCallback`?

treshugart:
I'd rather it be a getter, I think, because it follows a similar convention to `shadowRoot`. Maybe `renderRoot`, or something?

src/with-render.js
treshugart: Nah, it should have been fine for closed roots before?

matthewp:
@treshugart ready to be reviewed again.

matthewp:
Ah, good point, I'll add a test where it renders to itself.

matthewp:
Test added.

PR #1163, opened 2017-05-08
Converts much of the source to Flow, and adds a way to export both Flow and TS types

treshugart:
*Work in progress*

The conversion from Vanilla JS to Flow - instead of TypeScript - is happening because:

1. We want types
2. We are collaborating on a tool to generate both Flow and TypeScript types from Flow source
3. We want simpler incremental adoption (incremental by default, and no config)

***Please no comments on the merits of TypeScript over Flow***


matthewp:
I'm out of commission until Wednesday. No objections to merge from me.

package.json
```diff
@@ -49,11 +49,12 @@
...
     "docs:build": "gitbook install && gitbook build",
     "docs:watch": "gitbook install && gitbook serve",
-    "prepublish": "nwb build",
+    "prepublish": "nwb build && npm run types",
     "release": "semantic-release pre && npm publish && semantic-release post",
     "test": "node umd/skatejs.js && semistandard && npm run test:ts && nwb test -- --ci",
     "test:ts": "tsfmt -r && tsc -p ./",
-    "test:watch": "nwb test --server"
+    "test:watch": "nwb test --server",
+    "types": "../../treshugart/get-typed/bin/get-typed.js src/index.js --dest umd/skatejs"
```
matthewp: Looks like this isn't on the github yet, does this mean only you can publish for now (no problem for me, just curious)

treshugart: That's true. This will end up being published prior this being merged. Just wanted to start getting some early feedback.

Hotell: I would like to see generated TS definitions. I highly doubt that the "new tool" will correctly generate proper mixin definition for class type extension. If they do on the other hand then kudos, high five, mic drop.

I'm back from vacation, I'll look into this and other things this week.

maybe off topic but I hope flow works OK on windows, will test that as well. ( build is not working on win at the moment )

treshugart: I don't have Windows, but they officially support it https://flow.org/blog/2016/08/01/Windows-Support/

treshugart: > I highly doubt that the "new tool" will correctly generate proper mixin definition for class type extension

Can you provide some sample code that might make this fail?

Hotell: as I've said I would like to see the generated code :) can you submit it as well pls? then I can evaluate.

I have windows, and moving to it when working from home, so I'll be testing it thoroughly.

thanks!

treshugart: @Hotell https://gist.github.com/treshugart/9bbbba9d524dd75eb2ccf6b4b21df706#file-flow-ts-definitinos-md

Hotell: where is definition of this weird `_Class` type? https://gist.github.com/treshugart/9bbbba9d524dd75eb2ccf6b4b21df706#file-skatejs-d-ts-L17

PR #1173, opened 2017-05-15
Add types back in

### LINKED ISSUE #1168: 5.x alpha npm package is missing types.
The `src` folder is not included in the package. I'd say bundle everything from src but you might keep only the `d.ts` files also.

treshugart:
Fixes #1168.


freeman:
lgtm

Thanks

Issue #1142, opened 2017-04-20
Create an es-latest build


matthewp:
This build will be to support ES2016 environments.

treshugart:
I'd say that the criteria for this is that you should technically be able to load `skatejs/es-latest` in Safari TP and everything should work. The only blocker for this is dependencies. For example, if you use `Component` or `withComponent`, it tries to `import ... from 'preact'';` which will break it. What's the best way to solve this? cc @justinfagnani (I just saw your issue in Incremental DOM :))

matthewp:
Does @justinfagnani have some super-secret build of Chromium or did he build from source himself? I had to test with this: https://github.com/matthewp/do-modules-work and not in Canary yet.  Safari though!

To answer your question @treshugart, we would need a babel plugin that rewrites the module identifiers to their full paths. So that 

`import './with-module';` becomes `import './with-module.js';`

And 

`import 'preact';` becomes `import '../node_modules/preact/preact.js';`

treshugart:
There must be a babel plugin for that.

Hotell:
Good article to es2016 raw environments

https://www.contentful.com/blog/2017/04/04/es6-modules-support-lands-in-browsers-is-it-time-to-rethink-bundling/

matthewp:
I added a build in 3984e4f064ea3005d27e907d49d79f038c5226a9. Going to check it against my fritz and see if it works. Also going to give the babel plugin idea a shot.

matthewp:
I can confirm that fritz builds and runs fine with this change. So its just a matter of seeing if we can get this to run in Safari.

Conceptually this babel plugin is very similar: https://github.com/unpkg/babel-plugin-unpkg-rewrite/blob/master/modules/index.js . Except that we'll be resolving to the module's path, instead of using a CDN. So I'll use that a basis.

matthewp:
Got it working locally!  Not quite PR ready.

![screen shot 2017-04-25 at 10 50 41 am](https://cloud.githubusercontent.com/assets/361671/25391873/3b9a14c2-29a5-11e7-872c-864436b24785.png)

treshugart:
Exciting!

matthewp:
The one downside is that it needs to copy over preact to `es-latest/preact` because preact is not web-compatible and must go through the same babel plugin. I filed https://github.com/developit/preact/issues/670 to see if they are open to having a web-compat build too.  If so then we could point skatejs to that during out own build.

[...]

PR #1175, opened 2017-05-16
#1142  - Create ES latest build

treshugart:
This creates an ES2015 build in the `es/` folder, as opposed to using NWB to do ES5 + ESM. The downside is that consumers must still transpile the source from ES2015 to their target. The upside is it reduces complexity of having `es/` and `es-latest/`.

Maybe we should consider publishing both? The main question here is how would one consume `module` from the `package.json` and transpile it with little friction. It seems they'd have to include `node_modueles/skatejs` in their babel config? This is why I liked the ES5 + ESM bundle; it worked well with the `module` field. Either way the consumer must include `node_modules/skatejs`, I guess.

Thoughts welcome.


matthewp:
Does this add `.js` to module specifiers?

treshugart:
@matthewp not yet. That still needs to be done.

matthewp:
Sure, fine by me. Just making sure that the plan is for `es/` to be for web usage as well.  Do we just need to figure out that problem you discovered with babel-plugin-modules-web-compat, right?

treshugart:
> Sure, fine by me. Just making sure that the plan is for es/ to be for web usage as well. Do we just need to figure out that problem you discovered with babel-plugin-modules-web-compat, right?

I'd like to make it consumable by both web and bundlers. Using suffixes shouldn't be an issue, but we'll need to make an opinion about peer deps like Preact and where we'll look them up. As discussed in https://github.com/matthewp/babel-plugin-modules-web-compat/pull/1 it seems as though we should assume that they'd be siblings. However that doesn't work if you have an npm org, so may have to do some org detection on both the core package and its deps.

nwb.config.js
```diff
@@ -1,19 +1,53 @@
...
+const path = require('path');
+const rimraf = require('rimraf');
+const yargs = require('yargs');
+
+const browsers = require('./test/browsers');
+
+const [ cmd ] = yargs.argv._;
+const isBuild = cmd === 'build';
+const babelConfig = {
+  plugins: ['transform-react-jsx']
```
Hotell: hmm why is this needed? we aren't using jsx in source code are we ?

nwb.config.js
```diff
@@ -1,19 +1,53 @@
...
+const fs = require('fs');
+const glob = require('glob');
+const mkdirp = require('mkdirp');
+const path = require('path');
+const rimraf = require('rimraf');
+const yargs = require('yargs');
+
+const browsers = require('./test/browsers');
+
+const [ cmd ] = yargs.argv._;
```
Hotell: huh, isn't there any sugar feature passing args like in pure webpack ?

nwb.config.js
```diff
@@ -1,19 +1,53 @@
...
+const isBuild = cmd === 'build';
+const babelConfig = {
+  plugins: ['transform-react-jsx']
+};
+
+if (isBuild) {
+  const files = glob.sync('src/**/*.js');
+  rimraf.sync('./es');
+  files.forEach(f => {
```
Hotell: isn't there some easier way how to do this? aren't there some babel-cli flags for that? sorry for dumb questions, I'm just not used to for such kind of complexity to just emit raw es6 to some other folder (  in TS this is just one liner from cli :-| )

nwb.config.js
```diff
@@ -1,19 +1,53 @@
...
+    const targetDir = path.dirname(target);
+    mkdirp.sync(targetDir);
+    fs.writeFileSync(target, code);
+  });
+}
 
 module.exports = {
   type: 'web-module',
   npm: {
+    // CJS is kinda useless if we have UMD.
```
Hotell: 👍

nwb.config.js
```diff
@@ -1,19 +1,53 @@
...
     cjs: false,
-    esModules: true,
+
+    // We create a custom ESBuild above.
```
Hotell: I'm confused, so the custom ESBuild will produce  es6 modules with transiled code to es5 ?
-> ` presets: ['stage-0', 'es2017', 'es2016']`

If so, this doesn't make any sense -> we don't want classes to be transformed to function

treshugart: Each stage converts it down to the syntax of the latest spec. Each spec preset converts it to the previous year. So ES2016 compiles down to ES2015.

nwb.config.js
treshugart: https://github.com/yargs/yargs is sugar

nwb.config.js
treshugart: We use it in tests, but don't need it for prod builds. The tests and bundles share the NWB babel config,

nwb.config.js
treshugart: > I'm just not used to for such kind of complexity to just emit raw es6 to some other folder ( in TS this is just one liner from cli :-| )

Let's set the TypeScript stuff aside, please. I think I can probably work out a simple command for it, I was just sketching off the top of my head.

treshugart: Command found and implemented.

package.json
```diff
@@ -9,7 +9,6 @@
     "url": "https://github.com/skatejs/skatejs"
   },
   "main": "umd/skatejs.js",
-  "module": "es/index.js",
```
Hotell: so with this I am unable to tree shake skate right? I would have to explicitly alias it in webpack for instance ( with which I'd have to use native-shim though )

package.json
```diff
@@ -53,7 +56,7 @@
   "scripts": {
     "docs:build": "gitbook install && gitbook build",
     "docs:watch": "gitbook install && gitbook serve",
-    "prepublish": "nwb build",
+    "prepublish": "nwb build && babel src --out-dir es --presets=stage-0,es2017,es2016",
```
Hotell: 👍

PR #1179, opened 2017-05-22
TS typings update 5x

Hotell:
TODO: 
- [x] need to tighten up mixin types ( this isn't currently possible from TS side - mixin generics )

Closes https://github.com/skatejs/skatejs/issues/1102 -> partialy from our side


Hotell:
Tests are green --> issue with karma and souce labs. I HATE KARMA PERIOD :D

PR #1180, opened 2017-05-23
build: add size script

Hotell:
- adds size script


treshugart:
What does this do?

Hotell:
just emits minified and gzipped skate size. I've noticed that this is emitted by nwb during build anyway so huh

treshugart:
Are we okay to close this since NWB already does it?

Hotell:
sure no hard feelings on this one :D 🍺

PR #1181, opened 2017-05-23
build: add back es module + transpiled output

Hotell:
as we discussed here https://github.com/skatejs/project/pull/2#discussion_r117809028

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /coverage
 /demo/dist
 /es
+/es-next
 /lib
 /node_modules
 /umd
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 /coverage
 /demo/dist
 /es
-/es-next
+/es-latest
 /lib
 /node_modules
 /umd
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -56,7 +58,7 @@
   "scripts": {
     "docs:build": "gitbook install && gitbook build",
     "docs:watch": "gitbook install && gitbook serve",
-    "prepublish": "nwb build && babel src --out-dir es --presets=stage-0,es2017,es2016",
+    "prepublish": "nwb build && babel src --out-dir es-next --presets=stage-0,es2017,es2016",
```
treshugart: 🚲 shed: es-next or es-latest?

Hotell: hmm I think `es-latest` is more appropriate, updating

treshugart:
Are we okay to close this since NWB already does it?

.gitignore
```diff
@@ -1,6 +1,7 @@
 /coverage
 /demo/dist
 /es
+/es-next
```
treshugart: `es-latest` (will probably fix the build, too)

Hotell: Whoops. Fixed

PR #1187, opened 2017-06-09
build: Expose raw source through `esnext`

alexlafroscia:
This exposes the raw source through the `esnext` key in the `package.json`, as explained here:

http://2ality.com/2017/06/pkg-esnext.html

As the blog post covers, this will allow users to detect packages that purposely offer their raw source for transpilation by the user. This is especially useful since the advent of `babel-preset-env` as it allows the user to control what their dependencies transpile, since the package author might choose to transpile things that the user could actually leave alone.

Once this change makes it into a publish Skate release, I'll update `generator-skatejs` to detect/use the raw version of the Skate source, since it's already set up to build through `babel-preset-env`. That should make it super easy to actually deliver ES6 classes to browsers, should the user's environment support it. I tried this out locally and it worked great.


README.md
```diff
@@ -109,11 +109,17 @@ If you want finer grained control about which polyfills you use, you'll have to
...
 
-When you load Skate by module name (`import { ... } from 'skatejs';` or `require('skatejs');`), you'll be getting the transpiled source. Thus even if you author your components in ES2015, you'll still be getting ES5 base-classes and the native custom elements implementation will complain. If you want to deliever native classes you have to point to the non-transpiled Skate source: `import { ... } from 'skatejs/src';`. Currently this is not supported by our API versioning but we have an [issue](#992) to work around this.
+When you load Skate by module name (`import { ... } from 'skatejs';` or `require('skatejs');`), you'll be getting the transpiled source. Thus, even if you author your components in ES2015, you'll still be getting ES5 base-classes and the native custom elements implementation will complain.
 
-More information can be found in the [webcomponents/custom-elements](https://github.com/webcomponents/custom-elements#known-issues) repo.
+If you want to deliever native classes you have to point to the non-transpiled Skate source:
+
+```
+import { ... } from 'skatejs/src';
```
treshugart: I wouldn't advocate the usage of `src` directly as that will be a moving target and we don't want it to be API.

package.json
```diff
@@ -11,6 +11,7 @@
   "main": "umd/skatejs.js",
   "module": "es/index.js",
   "es2015": "es-latest/index.js",
+  "esnext": "src/index.js",
```
treshugart: What do you think about:

1. Remove the `es2015` key above.
2. Point this to `esnext/index.js`.
3. Update the build command to use `esnext` instead of `es-latest`

?

alexlafroscia: Yeah, that works too. After I did this work, I thought that it might actually make more sense to point `esnext` at a version of the source that's transpiled to the latest JS spec (there are a few things that are used in Skate that aren't actually part of JS yet, like object spread).

If you want to remove `es2015` that's fine with me, but technically the semantics between `es2015` and `esnext` are _slightly_ different, as the blog post mentions.

treshugart: I think the main things I want are:

1. Consistency between the directory name and the package.json key.
2. Targeting the latest spec.

No point in having several versions of ES built.

alexlafroscia: Do you still want to ship an `es` directory for `pkg.module`?

README.md
alexlafroscia: 👍 updated. I removed the original mention of `skatejs/src` and just point to the blog post about configuring Webpack to pull the `esnext` version of the source.

alexlafroscia:
Not sure what's up with the build... The tests ran (and passed) locally.

treshugart:
@alexlafroscia that'd be because it's on your fork and browserstack / travis has issues with that. If you're ready for this to land, go ahead and merge it. It LGTM :)

alexlafroscia:
Cool! Ready on my end -- I'll take a look at updating the Skate generator to use the `esnext` version of the source so that it can benefit from the tree-shaking and everything, since it's already set up to use `babel-preset-env` out of the box.

PR #1188, opened 2017-06-15
Use CustomEvent when available

cary-smith:
There seems to be a problem with skatejs.emit when an emitting component is part of the shadowDOM template of another component specifically in Chrome; The event never bubbles out of the shadowDOM even with the use of {composed: true}. 

Test Case (using TypeScript): 
```JS
import { Component, h, emit } from 'skatejs';

export class Wrapper extends Component<void> {
    static get is() { return 'el-wrapper'; }

    constructor() {
        super();
        // Never reached :(
        this.addEventListener('internal-event', (event) => console.log('Event received', event));
    }

    renderCallback(){
        return (
            <div><el-emitter></el-emitter></div>
        );
    }
}
customElements.define(Wrapper.is, Wrapper);

export class Emitter extends Component<void> {
    static get is() { return 'el-emitter'; }

    connectedCallback() {
        super.connectedCallback();
        console.log('Sending event');
        emit(this, 'internal-event', { composed: true });
    }

    renderCallback(){
        return (<div>Hello World!</div>);
    }
}

declare global {
    namespace JSX {
        interface IntrinsicElements {
            'el-emitter': Partial<HTMLElement>
        }
    }
}
customElements.define(Emitter.is, Emitter);
```

### DIFF
--- a/src/emit.js
+++ b/src/emit.js
@@ -8,15 +8,6 @@ const defs: EventOptions = {
   composed: false
 };
 
-declare global{
-    interface CustomEventInit {
-        composed: boolean;
-    }
-    interface CustomEvent {
-        composed: boolean;
-    }
-}
-
 const hasNativeSupport = verifyNativeEventSupport();
 
 function verifyNativeEventSupport() {
--- a/src/emit.js
+++ b/src/emit.js
@@ -8,10 +8,36 @@ const defs: EventOptions = {
   composed: false
 };
 
+declare global{
+    interface CustomEventInit {
+        composed: boolean;
+    }
+    interface CustomEvent {
+        composed: boolean;
+    }
+}
+
+const hasNativeSupport = verifyNativeEventSupport();
+
+function verifyNativeEventSupport() {
+    try {
+        const detail = {};
+        const test = new CustomEvent('testEvent', { composed: true, detail });
+        return test.composed === true && test.detail === detail;
+    } catch (error) {
+        return false;
+    }
+}
+
 export function emit (elem: HTMLElement, name: string, opts: EventOptions): boolean {
   opts = { ...defs, ...opts };
-  const e: ComposedCustomEvent = document.createEvent('CustomEvent');
-  e.initCustomEvent(name, opts.bubbles, opts.cancelable, opts.detail);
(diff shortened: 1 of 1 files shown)


cary-smith:
I am sure the code could be cleaned up a bit more, but this patch worked for me.

src/emit.js
```diff
@@ -8,10 +8,27 @@ const defs: EventOptions = {
...
+  }
+}
+
 export function emit (elem: HTMLElement, name: string, opts: EventOptions): boolean {
   opts = { ...defs, ...opts };
-  const e: ComposedCustomEvent = document.createEvent('CustomEvent');
-  e.initCustomEvent(name, opts.bubbles, opts.cancelable, opts.detail);
-  Object.defineProperty(e, 'composed', { value: opts.composed });
+  let e: ComposedCustomEvent;
+  if (hasNativeSupport) {
```
treshugart: Instead of checking `hasNativeSupport` and using the function above, you can probably just do `'composed' in CustomEvent.prototype`.

src/emit.js
```diff
@@ -8,10 +8,27 @@ const defs: EventOptions = {
...
+    e = new CustomEvent(name, opts);
+  } else {
+    e = document.createEvent('CustomEvent');
+    e.initCustomEvent(name, opts.bubbles, opts.cancelable, opts.detail);
+    Object.defineProperty(e, 'composed', { value: opts.composed });
```
treshugart: If `composed` isn't in custom event, we're in polyfill land, so I'm not sure if we even have to define it here. Thoughts?

src/emit.js
cary-smith: Excellent point! Making the changes.

src/emit.js
cary-smith: Technically we are already in polyfill-land with the original code. I believe it is still useful for browsers that cannot catch up. Since Edge and FF are likely to be implemented soon and IE11 is sliding fast, this may not be needed for too long. It's a shame, but since its still in heavy rotation here, we have to support IE11 for the time being. :angry:

treshugart: I guess what I meant is: do you know if the polyfill fixes the custom event constructor? If so, then we can likely remove this.

Issue #1196, opened 2017-07-06
Support progressive enhancement for components rendered with @skatejs/ssr in 5.*


majo44:
The goal is to author the custom elements once and use them on server side for ssr and on client side to make them alive (eg attach event handlers). 

Currently where the page is prerendered on server side by @skatejs/ssr on client side whole inner content of each custom-element is duplicated because of on https://github.com/skatejs/skatejs/blob/5.x/src/with-component.js#L12 `this._preactDom` is `null`. 

This cause that skatejs is not universal for the server and client rendering. 
The simplest possible solution is to replace:
```
this._preactDom = render(renderCallback(), shadowRoot, this._preactDom);
```
by 
```
this._preactDom = render(renderCallback(), shadowRoot, this._preactDom || shadowRoot.firstElementChild);
```
But this will requires that renderCallback will always return one JSXElement, the array of JSXElement returned from renderCallback will be not supported. 

Unfortunately this will not work proper where rendering will be to light DOM. I described the problem here: https://github.com/developit/preact/issues/654 and created separate issue for this problem https://github.com/skatejs/skatejs/issues/1197

treshugart:
@majo44 

> But this will requires that renderCallback will always return one JSXElement, the array of JSXElement returned from renderCallback will be not supported.

Skate doesn't support this anymore out of the box because Preact doesn't support it, so I think it's okay if we apply the patch as you've described to Skate. Happy if you want to do it. Thoughts on the best way to test it?

majo44:
Testing of this should be simple. 
1. before registering ce we can setup the dom structure (eg innerHTML, attachShadow) in the way how the custom-element should render itself
2. we can register mutation observer
3. register custom-elements
Assertions:
- there should not be any mutations in dom (light&shadow)
- all proper livecycle callbacks in ce should be called 
4. apply some property change in c-e
Assertions:
- there should be related mutations in dom (light&shadow)
- all proper livecycle callbacks in ce should be called

majo44:
Unfortunately there is no way to test on current testing env proposed scenario :( But I used some gaps in that env and  I provided some simple tests which are works fine for now.

treshugart:
I checked out the tests. Would you mind annotating what the issues where in the PR? It'd be good to know what the hard parts were.

PR #1198, opened 2017-07-10
fix(render): Fix for #1196 Support progressive enhancement for components rendered with @skatejs/ssr

majo44:
Fix for #1196


treshugart:
As noted in the issue, it'd be awesome to mark the pain points in the PR so we can address the shortcomings in the future. Merged because LGTM and that is an orthogonal issue. Thanks again for the PR!

majo44:
@treshugart 
"to mark the pain points in the PR so we can address the shortcomings in the future"
You mean in code or in the PR description ? More details about why this change is needed then just referring issue ?

treshugart:
You could just raise an issue, too. I just want to make sure we have record of something to improve upon in the future. Nothing urgent.

Issue #1197, opened 2017-07-06
Rendering to light DOM is not works properly. When parent custom-element is rerendering, child custom-element content is lost.


majo44:
Example: https://www.webpackbin.com/bins/-KoOzIcTUxPE7oDkLAWq
I described this problem in https://github.com/developit/preact/issues/654
Please vote for updates in preact :) and apply this changes to skatejs

majo44:
The problem do not appear if I switch the render to https://github.com/snabbdom/snabbdom + https://github.com/Swizz/snabbdom-pragma. So I wait for #1217 :)

treshugart:
Interesting! There's a branch with some of the work already done. It's at the point where we've prepped it to move. @chrisdarroch knows the details but if you see something you might be able to help on, happy to discuss. I opened #1218 to discuss etc there.

treshugart:
Hey @majo44, we just merged moving out the preact renderer. Can you see if there's anything else we need to do here? I can't tell if there is or not. Thanks!

chrisdarroch:
This appears to be fixed when using Skate 5. Here's the updated example, @majo44  https://www.webpackbin.com/bins/-KuYJlTdy2SWIN_FVu7J

majo44:
@chrisdarroch 
Thank you. 
But unfortunately  as I see this example not working properly, the attribute change is not triggering the second render. I will look at this a bit later.

chrisdarroch:
hmm, I'll have to leave that part with you and @treshugart. Given that the child is rendering static text and isn't invoking the `h()` function, I'm not sure whether I'd expect the child to re-render after changing the attribute on the parent. I'm pretty new to react/preact, so I'm missing a load of context to understand what should be expected behaviour here.

treshugart:
Updated bin: https://www.webpackbin.com/bins/-KuYl9Fid-xH_0M81njR.

Ok, so this really isn't an issue with Skate. Sure, the content is lost but this is due to Preact's diffing algorithm. Why not just use Shadow DOM and polyfill it? 

Alternatively, you *could* use something like this: https://gist.github.com/treshugart/e19c2be255448cd31d261f2e8db2127c, but YMMV with it. I'm ambivalent about making it "a thing", as in giving it a repo because of how close we really are to having shadow DOM everywhere.

[...]

PR #1201, opened 2017-07-12
[WIP] fix(render): Fix for #1197

majo44:
Fix for #1197 but this depends on changes for https://github.com/developit/preact/issues/654

### DIFF
--- a/src/with-component.js
+++ b/src/with-component.js
@@ -1,9 +1,19 @@
 // @flow
 
-import { h, render } from 'preact';
+import { h, render, options } from 'preact';
 import { withProps } from './with-props';
 import { withRender } from './with-render';
 import { withUnique } from './with-unique';
+import { isDefined } from './define';
+
+let oldVnodeHook = options.vnode;
+options.vnode = (vnode) => {
+  if (oldVnodeHook) oldVnodeHook(vnode);
+  vnode.skipChildren =
+    typeof vnode.nodeName === 'string' &&
+    vnode.nodeName.indexOf('-') > 0 &&
+    isDefined(vnode.nodeName);
+};
 
 export const withComponent = (Base?: Class<any> = HTMLElement): Class<HTMLElement> =>
   class extends withRender(withUnique(withProps(Base))) {
--- a/src/define.js
+++ b/src/define.js
@@ -10,3 +10,8 @@ export function define (Ctor: Class<CustomElement>): Class<CustomElement> {
   }
   return Ctor;
 }
+
+export function isDefined (name) {
+  const registry: FixedCustomElementRegistry = customElements;
+  return !!registry.get(name);
+}
--- a/test/unit/with-render.spec.js
+++ b/test/unit/with-render.spec.js
@@ -229,4 +229,36 @@ describe('withRender', () => {
       return mount(elem).waitFor(() => elem.innerHTML === '<x-c><div>Hello2</div></x-c>');
     });
   });
+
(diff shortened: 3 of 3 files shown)


src/with-component.js
```diff
@@ -1,9 +1,19 @@
...
 import { withUnique } from './with-unique';
+import { isDefined } from './define';
+
+let oldVnodeHook = options.vnode;
+options.vnode = (vnode) => {
+  if (oldVnodeHook) oldVnodeHook(vnode);
+  vnode.skipChildren =
+    typeof vnode.nodeName === 'string' &&
+    vnode.nodeName.indexOf('-') > 0 &&
+    isDefined(vnode.nodeName);
```
treshugart: Why not just use `customElements.get(vnode.nodeName)`?

treshugart:
While I can see a need to do something like this, I don't think it belongs in core Skate as the core focuses on the happy path of Preact rendering to a shadow root. If you want to have a custom `renderRoot`, then it should be up to the implementation (or non-core extension) to do what is being done here.

Thoughts?

majo44:
@treshugart
 Sorry I do not understand this point of view.
Library is exposing some API (Component, renderCallback, renderRooot,....)
I used this core API and figure out that this API is not work proper.
I even created the test case which is exactly shows that the core API is not working as expected.
So there are two possibilities: 
1. API is not designed well, (and eg renderRoot should be removed) or 
2. this is just a bug, which have to be fixed (if is not fixable or you do not want to fix it please look point 1.).

treshugart:
The core of the problem here is that Preact will blow away anything it doesn't know about. This is a side-effect of the diff / patch algorithm of most virtual DOM libraries and is not a bug. I realise you probably already know this, but I want to be clear for the sake of posterity.

Skate provides `renderRoot` as an escape hatch for you to render to whatever you want but does not make any assumptions as to what that is because it could be anything. Skate is not misbehaving here; this design is intentional. The API line is drawn here.

Modifying Preact `options` here can also have side-effects if the same Preact instance is shared in other places. We had problems when doing this with Incremental DOM and the workarounds were hard to understand and difficult to maintain.

I like the idea of documenting how to do this and even going as far to provide a non-core way of doing it as a separate implementation but I don't want to maintain it as part of the core. The focus of this part of the API is to try and coerce the consumer to use Shadow DOM because this is the ideal path, but give them an escape hatch to do whatever they want, at their own risk because sometimes that is necessary. This is where the API line is drawn.

If this makes the API "not designed well," then so be it.

majo44:
@treshugart 
" This is a side-effect of the diff / patch algorithm of most virtual DOM libraries and is not a bug. "
It is not full true. Incrementl-dom have a skip/skipNode which address this problem. There are also https://github.com/Matt-Esch/virtual-dom/blob/master/docs/thunk.md, https://github.com/snabbdom/snabbdom#thunks. In other VDOMs the live cycle interceptors can be used for the same purpose eg: https://github.com/patrick-steele-idem/morphdom#morphdomfromnode-tonode-options--node. Even Preact for own Components supports this by the shouldComponentUpdate, but is to hi level in this case. 

I agree that approach like in this PR it is not perfect. And I just realized that such solution in case of mix shadow & light DOM components on one page will cause a next bug (the slots will be not working in ShadowDOM, because light DOM will be not updated so shadow DOM will not update the slots).

When I said that Api is not designed well I had in mind that maybe we should consider different approach.

So lets think forward. I will look for other options.

PR #1203, opened 2017-07-16
feat: Web compatible ES Module build

### LINKED ISSUE #1189: Make esnext output web compat
References #1142, https://github.com/matthewp/babel-plugin-modules-web-compat/pull/4.

The `exnext` build works in well in Node and bundlers, but won't work in a browser because they're not relative paths.

matthewp:
This fixes #1189 

screen shot 2017-07-16 at 1 50 02 pm


test/esm-sanity.html
```diff
@@ -0,0 +1,16 @@
+<!doctype html>
```
treshugart: Can we automated this (not now, but would be good to have a plan)?

matthewp: Would be good to be able to run the tests with this build perhaps.

PR #1213, opened 2017-08-10
Docs - start process for deploying ssr site to netlify

treshugart:
This deletes the gitbook stuff (which we can still access on 5.x). This can be merged to 5.x prior to having a 1:1 docs ratio to Gitbooks (because master contains the existing docs), but 5.x shouldn't be merged until docs can sufficiently replace the gitbook.

The reason we are replacing Gitbook is because we want to dogfood the SSR lib and to have a custom design (as opposed to the gitbook layout). Gitbook really helped us get to this point.


matthewp:
:+1

treshugart:
To preview what the site will look like all we need to do is click "details" on the Netlify build. This is so awesome!

treshugart:
@matthewp I've started adding a couple components and pages just to flesh out how this is going to work. Not trying to step on your toes.

matthewp:
I see you're using the `<shadow-root>` trick... looks good to me!

PR #1214, opened 2017-08-11
fix: Prevent esnext from being transpiled to commonjs

matthewp:
Adding the --no-babelrc flag fixes it.

Need to add testing for the esnext build eventually..


matthewp:
I guess netlify will fail until your PR is merged @treshugart ?

treshugart:
Merged the PR then merged master with this. Will merge if green.

Issue #903, opened 2016-11-03
Improve docs examples: simple to complex


treshugart:
A comment on hacker news pointed out that the docs should start out a little bit simpler and go to more complex. Currently, there's a couple examples but they are very similar. Coming from web components and seeing Skate, one might not know what `render()` means right off the bat and may need some explanation. While we don't want to detract those coming from a React (i.e. `render()` world, examples starting simple and employing some of the spec'd stuff before introducing it might fare better.

More info: https://news.ycombinator.com/item?id=12862025

Hotell:
As I've suggested by person in London 😉  I think it would be very nice to port docs to gitbook and have similar experience like "readers" of [redux](http://redux.js.org/) or [ngMetadata](https://hotell.gitbooks.io/ng-metadata/content/) have. I'm here to help with this. just lemme know if you think it's good idea and wanna proceed.

treshugart:
Yeah, more than happy to do that. I think it would be simpler than maintaining a website, though, the website is a great as a suite of example components. But dev speed > website IMO, so go for it!

haoxins:
Personally, I'd like to add an example of `rxjs and skatejs (stateless components)`.

treshugart:
@haoxins that's a great idea. I haven't had a chance to use RXJS yet, but if anyone has, I'd be really interested how this might work out.

pyros2097:
I think integrations with redux, mobx or even relay(v2) might help new users to use skatejs as a complete frontend solution as a replacement to react or angular.

jeznag:
Would be great to add docs on the toolchain required. I can't figure out how to get this example going: https://hackernoon.com/building-a-custom-tag-input-with-skate-js-fbd4cdf744f#.5pmuikday

I get this error:
![image](https://cloud.githubusercontent.com/assets/1034890/22001424/373a301e-dc99-11e6-827d-efc742475f96.png)

There's probably some obvious webpack/babel plugin to use. Is it just the JSX loader?

treshugart:
@jeznag replied via Twitter. See https://skatejs.gitbooks.io/skatejs/content/docs/api/#jsx. You don't need any special tools if you're doing raw JS. If you want to use JSX, you have to enable it.

Issue #860, opened 2016-10-13
Add docs explaining what slots are and how to use them


treshugart:
See #859.

micahscopes:
feel free to use [this one](https://jsbin.com/wababu/edit?html,js,output) as an example if you'd like

treshugart:
We should reference other articles as well as the spec when documenting it. I don't think we should put much effort into complex examples for it as it's very well documented other places; Skate just uses the platform here.

PR #1230, opened 2017-09-24
Doc changes

chrisdarroch:
## Requirements

- [x] The commit message follows [our guidelines](https://github.com/skatej/skatej/blob/master/docs/CONTRIBUTING.md#commit)
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)
- [x] Updated TypeScript definitions, if applicable
- [x] Update docs/migrating.md for breaking changes

## Related issues

#860, #903 

## Details

This is an attempt at demonstrating a few ways in which Skate can be utilised in the README, going from "simple" through to "complex", and making the "site" code orthogonal to it at the same time.

The two primary challenges with the docs are (1) the flow, and (2) their simplicity.

1. It's challenging to construct a linear flow through what Skate "is" and incremental examples of web components authored via Skate. The difficulty lies in requisite knowledge (i.e., what can we expect people to know prior to looking at Skate) and relevance (i.e., when people want to author web components, what will pique their interest the most?). This leads in to the second problem...
2. Simple examples of using Skate aren't that "simple" -- by that, I mean they end up looking less terse than I'd hoped.
    * If you compare a base Skate component to a base React/Preact one, the comparison is unfavourable, since there's more boilerplate for the base usage.
    * If you compare to a basic no-library web component, Skate is favourable, but only if you get in to the nitty-gritty of Skate's helper functions.

I'm not sure I've struck the right balance between a linear flow, terseness, and relevance yet. Comments and discussion are more than welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,34 +19,105 @@ Skate is a functional abstraction over [the web component standards](https://git
 - Adds several lifecycle callbacks for responding to prop updates, rendering and more
 - Provides a base set of [mixins](http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/) that hook into renderers such as [@skatejs/renderer-preact](https://github.com/skatejs/renderer-preact).
 
-## Installing
+## Anatomy of a Skate web component
 
-```sh
-npm install skatejs
-```
+At its core, Skate is about creating [Custom Elements](https://w3c.github.io/webcomponents/spec/custom/).
+Skate provides a series of [mixin functions](http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/)
+that enable you to control what your component can do.
 
-To use Skate with a renderer, you'll want to install one of them:
-
-```sh
-npm install skatejs @skatejs/renderer-[renderer]
-```
-
-Where `[renderer]` is one of:
-
-- [@skatejs/renderer-lit-html](https://github.com/skatejs/renderer-lit-html)
-- [@skatejs/renderer-preact](https://github.com/skatejs/renderer-preact)
-- [@skatejs/renderer-react](https://github.com/skatejs/renderer-react)
-- Or any custom renderer!
-
-## Basic usage
-
-HTML
+For example, to create a simple greeting component:
 
 ```html
-<x-hello name="Bob"></x-hello>
+<x-hello>Bob</x-hello>
 ```
 
-JavaScript (using the Preact renderer)
+We could define a Skate component that renders the contents of our Custom Element:
+
+```js
+import { withComponent, withRenderer } from 'skatejs';
+
+const Component = withComponent(withRenderer());
+
+class GreetingComponent extends Component {
+  rendererCallback (renderRoot, renderCallback) {
+    renderRoot.innerHtml = '';
+    renderRoot.appendChild(renderCallback());
+  }
+  renderCallback () {
+      let el = document.createElement('span');
+      el.innerHTML = 'Hello, <slot></slot>!';
+      return el;
+  }
+});
+
+customElements.define('x-hello', GreetingComponent);
+```
+
+When this element is rendered, the end-user will see `Hello, Bob!`,
+though the DOM will look something like the following:
+
+```html
+<x-hello>
+  #shadow-root
+    <span>Hello, <slot></slot>!</span>
+  Bob
+</x-hello>
+```
+
+This is the utility that web components provide when using Custom Elements and the Shadow DOM.
+
+### Watching element properties and attributes
+
+We can create a Skate component that watches for HTML attribute changes on itself:
+
+```js
+import { props, withComponent, withRenderer, withProps } from 'skatejs';
+
+const Component = withComponent(withProps(withRenderer()));
+
+class GreetingComponent extends Component {
+  static props = {
+    name: props.string
+  }
+  rendererCallback (renderRoot, renderCallback) {
+    renderRoot.innerHtml = '';
+    renderRoot.appendChild(renderCallback());
+  }
+  renderCallback ({ name }) {
+    return <span>Hello, {name}!</span>;
+  }
+});
+
+customElements.define('x-hello', GreetingComponent);
+```
+
+The resulting HTML when the element is rendered would look like this:
+
+```html
+<x-hello name="Bob">
+  #shadow-root
+    <span>Hello, Bob!</span>
+</x-hello>
+```
+
+Now, whenever the `name` property or attribute on the greeting component changes,
+the component will re-render.
+
+### Rendering using other front-end libraries
+
+Because Skate provides a hook for the renderer, it can support just about
+every modern component-based front-end library &mdash; React, Preact, Vue...
+just provide a `renderCallback` to stamp out your component's HTML,
+a `rendererCallback` to update the DOM with your HTML, and then it's all the same to Skate!
+
+The Skate team have provided a few renderers for popular front-end libraries;
+check the [Installing](#installing) section.
+
+#### Using Skate with Preact
+
+Instead of writing our own `rendererCallback`, we could use a library like
+[Preact]() to do the work for us. Skate provides a ready-made renderer for Preact;
+here's how we would update our previous greeting component to use it:
 
 ```js
 /** @jsx h */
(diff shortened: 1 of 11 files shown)


chrisdarroch:
Something else I'm interested in...

After the changes @treshugart made in #1213, the `docs/` directory is just... gone. Is the plan to rewrite the content for docs as a series of skated pages -- i.e., document skate using skate itself? I'm keen to help re-write the documentation, but the plan isn't clear to me; it's particularly confusing given I know that @matthewp is working on "the site", where I don't know what the overlap between "docs" and "site" is, so don't know if I'd be treading on toes.

matthewp:
Is the intention to have docs on the site? If so we'll need to do some static site generation I suppose.

chrisdarroch:
@matthewp I have no intentions other than writing docs and examples of using Skate. I don't really mind where the docs live. They could live in this repo or another; they could be deployed to the skate website or another. My problem is that the previous docs were deleted in #1213 and I don't know if there's a grand plan for replacing them.

README.md
```diff
@@ -12,50 +12,121 @@
...
+just provide a `renderCallback` to stamp out your component's HTML,
+a `rendererCallback` to update the DOM with your HTML, and then it's all the same to Skate!
+
+The Skate team have provided a few renderers for popular front-end libraries;
+check the [Installing](#installing) section.
+
+#### Using Skate with Preact
+
+Instead of writing our own `rendererCallback`, we could use a library like
+[Preact]() to do the work for us. Skate provides a ready-made renderer for Preact;
```
treshugart: Link to Preact.

README.md
```diff
@@ -12,50 +12,121 @@
...
+<x-hello>Bob</x-hello>
 ```
 
-To use Skate with a renderer, you'll want to install one of them:
+We could define a Skate component that renders the contents of our Custom Element:
 
-```sh
-npm install skatejs @skatejs/renderer-[renderer]
+```js
+import { withComponent, withRenderer } from 'skatejs';
```
treshugart: All you need is `withComponent()`, the `withRenderer()` part is included with it and has no effect when composed aside from adding another prototype to the chain.

README.md
```diff
@@ -12,50 +12,121 @@
...
+
+const Component = withComponent(withRenderer());
+
+class GreetingComponent extends Component {
+  rendererCallback (renderRoot, renderCallback) {
```
treshugart: I like how this colocates both the `renderer` and the `render`. It might make sense to iterate on this in another example to show how the renderer would be split out into a mixin. Then go into custom built renderers.

treshugart:
@matthewp, @chrisdarroch, to reiterate re the docs / static-site generation, `npm run dev` and `npm run site` do this for you. Everything is in the `site/` directory and already does all of this. There's a Netlify build to preview on PRs, too. For example, [this](https://deploy-preview-1230--housing-agent-ostritch-68781.netlify.com/) is the link in the "Details" check for Netlify.

treshugart:
@chrisdarroch I like how you've structured the tests. Something I've been liking lately is to co-locate tests with code, as opposed to separate directory. If we do this, we can run tests within the `site/` directory on the code examples, while having them co-located with where they're being used. I'm happy to not do this for now because we'd have to shave a couple yaks and get the babel config working with that setup. Just a general comment and a LGTM for what you're doing!

treshugart:
Re docs being gone, I deleted those because I think it's a bad idea to try and port them to the website. There'd be a lot of massaging and we'll miss something. Better to rewrite from scratch, I think. We can, however, retrieve them from a previous commit if we need to. They're also on master.

test/samples/simple/index.js
```diff
@@ -0,0 +1,17 @@
+import { withComponent, withRenderer } from '../../../src';
+
+let Component = withComponent(withRenderer());
```
Hotell: any particular reason why are you using withRenderer ? it's already within withComponent

test/samples/simple/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+class MyComponent extends Component {
+    rendererCallback (renderRoot, renderCallback) {
```
Hotell: it would be better to use this as a showcase of creating renderer maybe ?

test/samples/with-preact/index.js
```diff
@@ -0,0 +1,16 @@
+/** @jsx h */
+
+import { props, withProps, withComponent } from '../../../src';
+import withRenderer from '@skatejs/renderer-preact';
+import { h } from 'preact';
+
+const Component = withComponent(withProps(withRenderer()));
```
Hotell: any particular reason why are you using withProps ? it's already within withComponent

matthewp:
@treshugart Maybe I'm misunderstanding, are docs written in Markdown or in JSX? if they are written in Markdown then I think we do need some build process because there isn't one for markdown > html/jsx, right?

treshugart:
I want to just rewrite everything in JSX or lit (now that we've released the lit renderer). The SSR lib should support everything it needs, too.

test/samples/simple/index.js
chrisdarroch: None; I'll remove it.

test/samples/simple/index.js
chrisdarroch: I wanted to think about it from the perspective of someone new to web components and Skate. I wanted to create a barebones component that didn't have extra dependencies. 

The tradeoff here is "simplicity" versus "magic". I can either...

* Lean on one of the pre-made renderers in the "simple" example, which means that we believe the "simplest" Skate can be is a minimum of three top-level dependencies: `skate`, `@skatejs/renderer-x` and `x`.
* Use what Skate itself ships, which no longer includes a renderer, meaning if I want to demonstrate a component that renders, I need to write my own.

Maybe the "simplest" web component you can create is one that doesn't render at all? What would the use-case be for such a "simple" component?

If you think the demo with Preact is simpler than this one, I'm happy to use the preact one as the first demonstration of Skate. It changes my approach for how I was "building up" the explanation of what Skate is and does, but maybe it's for the better.

test/samples/simple/index.spec.js
```diff
@@ -0,0 +1,24 @@
...
+            done();
+        });
+    });
+
+    xit('renders slot contents', (done) => {
+        let el = document.createElement('hello-simple');
+        el.textContent = 'World';
+        mount( el ).wait(e => {
+            expect(e.shadowRoot.firstChild.textContent).toBe('Hello, !');
+            expect(e.textContent).toBe('World');
```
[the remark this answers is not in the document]
chrisdarroch: @treshugart do you have thoughts on why this test doesn't pass? Is my expectation wrong for what should happen here? I want to know if this is me being wrong versus there being a bug or shortcoming in one of the tools used here.

chrisdarroch:
Re: writing the docs + examples in JSX or lit-html, my vote would be on lit-html. The main reason is that while JSX is cool and all, I believe the pragma stuff is a mindfuck to newbies. I'd rather not give rise to questions that wonder what part of the example is "Skate" versus some other tech. lit-html has a clearer distinction there.

test/samples/simple/index.spec.js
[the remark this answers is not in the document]
treshugart: `e` is actually a wrapper. You'll want to do `e.node`. Both should be failing.

treshugart:
+1 for lit

test/samples/with-preact/index.spec.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-env jest */
+
+import { mount } from '@skatejs/bore';
+import './';
+
+describe('samples/with-preact', () => {
+    xit('renders what we expect', (done) => {
+        let el = document.createElement('hello-withpreact');
```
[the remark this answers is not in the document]
treshugart: Where is the element definition?

test/samples/with-preact/index.js
treshugart: You don't need `withProps()` here. The `withComponent()` mixin already brings that in.

test/samples/with-preact/index.spec.js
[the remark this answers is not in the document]
treshugart: I think the issue here is that Jest transpires to ES5 and when importing `@skatejs/renderer-preact`, it points to the `/node` source. To force ES5 you'll probably want to do `@skatejs/renderer-preact/umd`.

chrisdarroch:
Okay, I've fixed the test expectations and site generation. Only two things before I'd say this is merge-worthy:

* The netlify deploy is failing, though, and I'm not sure why. 
* I'm not sold on how I've done the readme... it feels more like a step-by-step tutorial for using Skate. I'm not sure what I'd put instead, but in terms of describing what Skate is all about, it doesn't feel "concise" enough to me.

treshugart:
You can do whatever you want re squashing. I'll fix the netlify deploy rn.

treshugart:
Looking at the deploy, it looks like you might want to pull from master and then retry.

PR #1235, opened 2017-10-12
fix: Symbol named props

deini:
**Please check if the PR fulfills these requirements**
- [X] The commit message follows [our guidelines](https://github.com/skatej/skatej/blob/master/docs/CONTRIBUTING.md#commit)
- [X] Tests for the changes have been added (for bug fixes / features)

**Does this PR introduce a breaking change?** (check one with "x")
```
[ ] Yes
[X] No
```
## Details
Fixes a bug where you can't use symbols as name of props if using the props helpers.

I'm not sure if this is the right approach, since I'm not entirely sure of the side effects of not having a dashCased string name.

### DIFF
--- a/src/util/with-props.js
+++ b/src/util/with-props.js
@@ -13,7 +13,7 @@ interface CanDefineProps extends HTMLElement {
 }
 
 export function normaliseAttributeDefinition(
-  name: string,
+  name: string | Symbol,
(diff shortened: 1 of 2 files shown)


treshugart:
We can fix the MetLife build later.

PR #1238, opened 2017-10-12
fix: flow errors. Refactor some types and code.

treshugart:
Minor refactors to fix some flow / prettier errors. Much of the PR is also prettier formatting.


src/util/index.js
```diff
@@ -1,34 +1,39 @@
...
+        }
+      };
 
-export function dashCase (str: string): string {
-  return str.split(/([_A-Z])/).reduce((one, two, idx) => {
-    const dash = !one || idx % 2 === 0 ? '' : '-';
-    two = two === '_' ? '' : two;
-    return `${one}${dash}${two.toLowerCase()}`;
-  });
+export function dashCase<T>(str: T): T {
```
Hotell: why generic? I guess it tightly set to be `string | symbol` right?

src/util/index.js
```diff
@@ -48,24 +53,29 @@ export function debounce (cbFunc: () => void): Function {
   };
 }
 
-export const empty = <T> (val: T): boolean => val == null;
+export const empty = <T>(val: T): boolean => val == null;
```
Hotell: again any reason why is this generic? I gues `mixed` should be sufficient in this case ?

src/util/with-props.js
```diff
@@ -18,17 +18,17 @@ export function normaliseAttributeDefinition(
...
-    typeof attribute === "object"
+    typeof attribute === 'object'
       ? { ...attribute }
       : {
           source: attribute,
           target: attribute
         };
   if (obj.source === true) {
-    obj.source = typeof name === 'symbol' ? name : dashCase(name);
+    obj.source = dashCase(name);
```
Hotell: 👌🏻 wanted to do this exactly after latest PR which fixes symbols as props

src/with-children.js
```diff
@@ -4,14 +4,18 @@ export const withChildren = (
   Base: Class<HTMLElement> = HTMLElement
 ): Class<HTMLElement> =>
   class extends Base {
+    childrenChangedCallback: Function | void;
```
Hotell: doesn't flow support something more robust [for mixin definition like typescript](https://github.com/Hotell/reactive-typesafe-webcomponents/blob/vanilla-mixins/02-vanilla-lit-html/src/mixins/withRender.ts#L10) has? hey just asking, no flame war ✌️

src/util/index.js
treshugart: Oh that, yeah, I saw that but didn't change that in this PR. Happy for you to do a fix / push if you want.

src/util/index.js
treshugart: Yeah, it is. So if I make it `string | Symbol`, then I get an error on the first and last lines [here](https://github.com/skatejs/skatejs/blob/3807c71ab54738d1703ded15bd04c21640b07a3c/src/util/with-unique.js#L20-L23) saying "Symbol This type is incompatible with the expected param type of string". Not really sure what to do here to fix that because `Ctor.name` is a string and we've typed it as such. Making it a generic at least made it have the same output as the input.

src/with-children.js
treshugart: How would this help `childrenChangedCallback`? All you'd need to do is declare an interface and apply it to the `Class<>` annotation. I don't think that's entirely necessary, though, is it?

Issue #1237, opened 2017-10-12
Concept of state as an internal props


treshugart:
I've spoken to a few people lately and it seems that it'd be worthwhile to add the concept of internal "state" to components.

Since state is tied closely to the concept of props, I'm thinking we could just add this functionality to the `withProps()` mixin.

## Usage

The following describes the current state of `props` along with the new `state`.

```js
import { state, withState } from 'skatejs';

export default class extends withState() {
  constructor() {
    super();

    // This represents the public API.
    this.name = 'Trey';

    // This is identical to the above.
    this.props = { name: 'Trey' };

    // These are all the same.
    this.props = { [state]: { name: 'Trey' } };
    this.props[state] = { name: 'Trey' };
    this[state] = { name: 'Trey' };

    // These have no effect in terms of triggering a propsChangedCallback().
    this.props.name = 'Trey';
    this.props[state].name = 'Trey';
    this[state].name = 'Trey';
  }
}
```

## Implementation

The implementation can simply be a symbolised prop. Consumers can do this by themselves quite easily, but it requires them to provide boilerplate and have their own opinion. We have an opportunity to streamline this for them so they don't have to think about it.

```js
import { props } from './with-props';

// We don't use symbols yet, so we'd have to account for older browsers.
function sym(id) {
  return typeof Symbol === 'function' ? Symbol(id) : `__skate_${id}`;
}

// Should we prefix this with something (i.e. $/_)?
export const state = sym('state');

export const withState = (
  Base?: Class<HTMLElement> = HTMLElement
): Class<HTMLElement> =>
  class extends Base {
    static props = {
      [state]: props.object
    }
  }
```

It's likely that this isn't useful without the `withProps()` mixin anyways, so it might be better to simply include it as part of that.

## Related

We introduced a `props` setter because that used to be the way to update state all at once. While you could have done an `Object.assign(elem.props, {})` somewhere, it sort of made sense to have a counterpart to the getter. However, with the introduction of `state`, it doesn't make as much sense.

The `props` getter is definitely useful because then you don't have to merge prototype chains, however, I don't think the setter has much use if we introduce state because public props are for external APIs. Maybe it's worth doing:

- `props` becomes a symbol and has only a getter (maybe not?)
- `state` has both a getter and a setter

Thoughts welcome. This isn't a super drastic change, but I'm wary of making a breaking change on a beta.

## Bikeshedding

- Should we use a symbol?
- If so, should we have a naming convention?
[...]

PR #1240, opened 2017-10-13
#1237 - Implement state

treshugart:
## Implementation

#1237

This implements state using a separate mixin. The mixin has a getter and setter that calls `triggerUpdate` if it's defined on the instance.

## Tests

The tests test both setting and that `triggerUpdate` is called. All other functionality is assumed ot be borrowed from something like `withProps` that implements a `triggerUpdate` method.

## Other

Theres some other stuff here:

- Moved the `_connected` check to `withRender`. No use in doing this in the props mixin unless we really want to block any props reactions when not connected. We should be doing this check in `withRender` anyways even if it's done in props.
- Added tests to `withProps` to ensure `triggerUpdate` and `triggerUpdateBatched` are called.

## Open questions

- Should we mix this in with `withProps`? I feel like too many mixins are going to get us into perf trouble.
- Do we want to continue checking `_connected` in `withProps` before calling `propsChangedCallback` or are we okay with calling that even when it's not connected? I feel this would break people's expectations, but not with rendering (or maybe at some point have an option for that for rendering).

### DIFF
--- a/src/with-component.js
+++ b/src/with-component.js
@@ -3,7 +3,10 @@
 import { withChildren } from './with-children';
 import { withProps } from './with-props';
 import { withRenderer } from './with-renderer';
+import { withState } from './with-state';
 import { withUnique } from './with-unique';
 
-export const withComponent = (Base?: Class<any>) =>
-  withChildren(withProps(withRenderer(withUnique(Base || HTMLElement))));
+export const withComponent = (Base?: Class<HTMLElement>) =>
+  withChildren(
+    withProps(withRenderer(withState(withUnique(Base || HTMLElement))))
+  );
--- a/src/with-props.js
+++ b/src/with-props.js
@@ -69,8 +69,6 @@ export const withProps = (
     static _observedAttributes: Array<string>;
     static _props: Object;
 
-    _connected: boolean;
-    _constructed: boolean;
     _prevProps: Object;
     _syncingAttributeToProperty: null | string;
     _syncingPropertyToAttribute: boolean;
@@ -120,24 +118,15 @@ export const withProps = (
 
     constructor() {
       super();
-      if (this._constructed) return;
-      this._constructed = true;
       this.triggerUpdateBatched = debounce(this.triggerUpdate.bind(this));
     }
 
     connectedCallback() {
-      if (this._connected) return;
-      this._connected = true;
-      // $FlowFixMe - HTMLElement doesn't implement connectedCallback.
-      if (super.connectedCallback) super.connectedCallback();
+      if (super.connectedCallback) {
+        // $FlowFixMe - not in HTMLElement.
+        super.connectedCallback();
+      }
       this.triggerUpdateBatched();
-    }
-
-    disconnectedCallback() {
-      if (!this._connected) return;
-      this._connected = false;
-      // $FlowFixMe - HTMLElement doesn't implement disconnectedCallback.
-      if (super.disconnectedCallback) super.disconnectedCallback();
     }
 
     // Called to see if the props changed.
@@ -158,7 +147,7 @@ export const withProps = (
 
     // Invokes the complete render lifecycle.
     triggerUpdate() {
-      if (this._updating || !this._connected) {
+      if (this._updating) {
         return;
       }
 
--- a/src/with-props.js
+++ b/src/with-props.js
@@ -49,7 +49,7 @@ export function prop(definition: PropOptions | void): Function {
           normalised.serialize,
           val
         );
-        this._updateDebounced();
+        this.triggerUpdateBatched();
       }
     });
   };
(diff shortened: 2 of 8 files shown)


src/with-component.js
```diff
@@ -3,7 +3,10 @@
 import { withChildren } from './with-children';
 import { withProps } from './with-props';
 import { withRenderer } from './with-renderer';
+import { withState } from './with-state';
 import { withUnique } from './with-unique';
 
-export const withComponent = (Base?: Class<any>) =>
-  withChildren(withProps(withRenderer(withUnique(Base || HTMLElement))));
+export const withComponent = (Base?: Class<HTMLElement>) =>
+  withChildren(
```
treshugart: This is growing... and growing... and growing. Perf?

src/with-props.js
```diff
@@ -49,7 +49,7 @@ export function prop(definition: PropOptions | void): Function {
           normalised.serialize,
           val
         );
-        this._updateDebounced();
+        this.triggerUpdateBatched();
```
treshugart: This is now explicitly public. I feel that if people are going to author mixins they're going to need this stuff. For example, if someone were to have written `withState`.

src/with-props.js
```diff
@@ -156,8 +146,8 @@ export const withProps = (
     }
 
     // Invokes the complete render lifecycle.
-    _updateCallback = () => {
-      if (this._updating || !this._connected) {
+    triggerUpdate() {
+      if (this._updating) {
```
treshugart: We now trigger an update even if it's not connected. Is this okay?

src/with-renderer.js
```diff
@@ -22,14 +24,28 @@ export const withRenderer = (
...
+      }
+      this._connected = true;
+    }
+
     propsChangedCallback() {
+      if (super.propsChangedCallback) {
+        // $FlowFixMe - not in HTMLElement.
+        super.propsChangedCallback();
+      }
+      if (!this._connected) {
```
treshugart: This doesn't break behaviour because this is what it was explicitly intended for.

src/with-state.js
```diff
@@ -0,0 +1,15 @@
+export const withState = (
```
treshugart: The new state implementation.

test/unit/with-state.spec.js
```diff
@@ -0,0 +1,28 @@
+import { define, withState, withUnique } from '../../src';
```
treshugart: The new tests.

src/with-component.js
matthewp: I wouldn't think so, why would it?

treshugart: I’m just wary of long prototype chains. Maybe it’s unfounded, though?

src/with-props.js
```diff
@@ -69,23 +69,22 @@ export const withProps = (
     static _observedAttributes: Array<string>;
     static _props: Object;
 
-    _connected: boolean;
-    _constructed: boolean;
```
treshugart: This was to ensure the constructor wasn’t called twice. The polyfill seemed to do this awhile back. Removed because there’s noooo way we should havrle to test for this.

src/with-component.js
matthewp: i'm assuming some inlining magic business or something.. worth testing though! :D

src/with-props.js
matthewp: I'd leave only the async version for now, would want very good reasons to allow sync.

src/with-component.js
treshugart: I'm happy to just keep it like it is. If there's a problem, we can probably do something around flattening the chain since they don't rely on `super`. I talked to some at length about this at Polymer Summit and, at least Chrome, optimises this.

src/with-props.js
treshugart: I (we?) chatted to Rich Harris about this awhile back. Things that affect layout want sync. Unless we offer a way out from our default async / batched, there's going to be times when some want to force a sync render so consumers can inspect layout props.

src/with-component.js
matthewp: Yeah, most elements already have a long prototype chain anyways, I don't think a half dozen extras will move the needle.

src/with-props.js
matthewp: Since shadow DOM is an implementation detail, consumers shouldn't have an assumption of what it looks like at any given point, IMO. What they *should* rely on is the element's own properties/events, and those should be updated synchronously.

treshugart: What if you set a prop that causes a re-render. It "looks" different, and as a consumer you know that, and it's an implementation detail you have to be aware of, and now you want to recalc the offset height.

```js
elem.somePropThatRenders = true;

// Wrong.
elem.offsetHeight;

// Right.
Promise.resolve().then(() => elem.offsetHeight);
```

I don't think it's fair for them to need to call `triggerUpdate()` sync, but if there were a mixin that fixed this behaviour for async, they'll want to call the sync version inside a getter for the accessors to ensure the render queue is flushed.

src/with-state.js
```diff
@@ -0,0 +1,15 @@
...
+): Class<HTMLElement> =>
+  class extends Base {
+    triggerUpdate: Function | void;
+    get state() {
+      return this._state || {};
+    }
+    set state(state: Object) {
+      this._state = state;
+      if (this.triggerUpdate) {
+        this.triggerUpdate();
```
treshugart: This is a sync re-render FYI

PR #1242, opened 2017-10-23
Add polyfills to docs site

bwhitty:
The docs site currently doesn't run in browsers which don't support
custom elements natively.

- [x] Bug
- [ ] Feature

## Requirements

- [ ] Read the [contribution guidelines](https://github.com/skatejs/skatejs/blob/5.x/CONTRIBUTING.md).
- [ ] Wrote tests.
- [ ] Updated docs and upgrade instructions, if necessary.
- [ ] Updated TS definitions, if necessary.

## Rationale

Having the docs site completely not load for a good portion of first-time users is probably not a good look.

## Implementation

Added `skatejs-web-components` as a dep and loaded it in the main bundle.
I have *no idea* what npm is doing with that diff, but all I did was `npm i --save skatejs-web-components`. I thought npm 5 fixed stuff like this, but doesn't appear so.

## Open questions

Do we want to do this? 
Is this still the best way to do it?
Should we implement a "canonical example" of polyfill loading (lazy, user-agent-based?) on the docs site as a type of best practice guide?

### DIFF
--- a/site/index.js
+++ b/site/index.js
@@ -1,4 +1,5 @@
 import 'file-loader!./index.html';
+import 'skatejs-web-components';
 import App from './components/app';
 
 document.getElementById('app').appendChild(new App());
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -17,6 +17,18 @@
       "resolved": "https://registry.npmjs.org/@skatejs/val/-/val-0.3.1.tgz",
       "integrity": "sha512-WejvTgQT9baY0hwb+VYE66fkKm8SekGdjEczneZvpUoepBsFlmeN2uQynCt2H0h2KTQppvgz0yPvbd4NGHCHYg=="
     },
+    "@webcomponents/custom-elements": {
+      "version": "https://github.com/webcomponents/custom-elements/archive/f4b6f1f.tar.gz",
+      "integrity": "sha1-b4BQGPw+Rwf/g06o6+GDOKBxfzg="
+    },
+    "@webcomponents/shadycss": {
+      "version": "https://github.com/webcomponents/shadycss/archive/bea402d.tar.gz",
+      "integrity": "sha1-QhhxxoFhCJm6EKQz5MkgyaMfRmY="
+    },
+    "@webcomponents/template": {
+      "version": "https://github.com/webcomponents/template/archive/60d743e.tar.gz",
+      "integrity": "sha1-BFX3CS0xyGbl7ptsAHLxo/Ay4xQ="
+    },
     "abab": {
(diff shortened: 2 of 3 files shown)


treshugart:
Thanks @bwhitty. The skatejs-web-components package has been deprecated in favour of just using the https://github.com/webcomponents/webcomponentsjs#webcomponents-loaderjs. I've planned on doing this in the future at some point, but it'd be awesome if you wanted to get this in. The only issue I see with it is finding a way to defer the loading of everything until that's done. Maybe with dynamic imports that's easier.

bwhitty:
I'll take a look at trying to use the standard polyfills instead. I think deference of loading the bundle can be achieved via some sort of bootstrap/main function?

Also do you know why the netlify build fails? Is that just an access issue? Tough to test changes without that build working.

treshugart:
I think this is an issue with Netlify and having nested `node_modules` and their `.bin` folders. If you retry and clear the build cache, it usually works.

treshugart:
Hey @bwhitty we've resolved the Netlify issue by moving all deps to the root `node_modules`. You should be good to merge and resolve conflicts. FYI we don't need ShadyCSS as we're using CSS in JS, but do need the shim (for native), CE and SD fills.

bwhitty:
Hey Trey, I'm not sure what else there is to do here except for a bug I think I see. I've taken some time to look at your updated site code and I'm really not sure why the sd-ce fill you've configured [here](https://github.com/skatejs/skatejs/blob/5.x/site/index.js#L23) doesn't work. It loads correctly, but then on Firefox the `import()` call within `mount` never resolves and so the app is never mounted.

firefox

Note there aren't any errors and the polyfill loads fine, but that import promise never resolves for now reason.

treshugart:
Whoah, ok. I'll have a look. I noticed some weirdness, too.

treshugart:
I've updated master to work on FF. The problem was an exception being swallowed by the promise. There was a `DOMException` being thrown by `insertRule` because FF doesn't yet understand the `:host` selector. I had to work around that by hitting style directly on the host in `connecting()`.

Try your branch now with latest master. The styles look a bit funky because they still need to be scoped for some of the dependencies. The parts that are using `yocss` should be fine.

treshugart:
K, styles are fixed in FF.

bwhitty:
I'm going to close this as your work appears to be doing just fine on FF now. 👍 

Thanks, Trey!

treshugart:
Thanks, @bwhitty. Sorry if that work made this redundant. Thanks for submitting this :)

PR #1250, opened 2017-11-19
docs: fix css and util links, update Readme, Contribution, prettier

Hotell:
## Rationale

*Why is this PR necessary?*

- lack of cross platform `code` styling
- wrong utils links
- update prettier + apply
- update markdown docs + contribution

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,33 +12,47 @@
 
 [![Sauce Test Status](https://saucelabs.com/browser-matrix/skatejs.svg)](https://saucelabs.com/u/skatejs)
 
-Skate is a functional abstraction over [the web component standards](https://github.com/w3c/webcomponents) that:
+Skate is a functional abstraction over
+[the web component standards](https://github.com/w3c/webcomponents) that:
 
-- Produces cross-framework compatible components
-- Abstracts away common attribute / property semantics via `props`, such as attribute reflection and coercion
-- Adds several lifecycle callbacks for responding to prop updates, rendering and more
-- Provides a base set of [mixins](http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/) that hook into renderers such as [@skatejs/renderer-preact](https://github.com/skatejs/renderer-preact).
+* Produces cross-framework compatible components
+* Abstracts away common attribute / property semantics via `props`, such as
+  attribute reflection and coercion
+* Adds several lifecycle callbacks for responding to prop updates, rendering and
+  more
+* Provides a base set of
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -82,6 +84,52 @@ When this element is rendered, the DOM will look something like the following:
 
 This is the utility that web components provide when using Custom Elements and the Shadow DOM.
 
+Skate also allows **turning off Shadow DOM** if you don't wanna use it for various particular reasons. You can turn it off via `get renderRoot()` override:
+
+> NOTE: by turning off Shadow DOM you cannot use <slot/> content projection anymore by default, further tweaks needs to be applied
```
Hotell: It would be great to provide also solution how to achieve "slotting" without Shadow DOM. You mentioned that you have a solution for this, can you provide it please ?

README.md
```diff
@@ -82,6 +84,52 @@ When this element is rendered, the DOM will look something like the following:
...
+
+```js
+import { withComponent, props } from 'skatejs';
+
+// define base class withouth Shadow DOM
+class NoShadowComponent = class extends withComponent() {
```
Hotell: just thinking out loud here, maybe we can add `withoutShadomDOM()` mixin to skate API, so then user can do just
`const Component = withComponent(withoutShadomDOM(withCustomRenderer()))` 
WDYT ?

treshugart: I don't think we need an extra proto chain just for that and shadow dom is only specific to the `withRenderer` mixin. I say just leave it, but happy for others to weight in. @chrisdarroch may have an opinion here.

treshugart:
Added a comment about `renderRoot` but otherwise LGTM. I'll merge like this and we can discuss that later if need be.

README.md
chrisdarroch: I think it's reasonable for people who don't want the shadow DOM to opt out themselves. Having the override point should be sufficient. My general thoughts are: make the most common thing easy, make the less common thing possible; the people who know what they're doing will manage the additional complexity.

README.md
```diff
@@ -80,11 +96,61 @@ When this element is rendered, the DOM will look something like the following:
...
+
+> NOTE: by turning off Shadow DOM you cannot use <slot/> content projection
+> anymore by default, further tweaks needs to be applied
+
+```js
+import { withComponent, props } from 'skatejs';
+
+// define base class withouth Shadow DOM
+class NoShadowComponent = class extends withComponent() {
+  // you need to return where you want to render your content, in our case we wanna render directly to our custom element children
```
chrisdarroch: The remainder of the lines in this diff fit in the 80 char width, except for this one </nitpick>

Issue #1253, opened 2017-11-28
types are not included in the latest release


cronon:
## Bug report

* Skate version: _5.0.0-beta.4_

package.json file says that types should sit in `src/index.d.ts`, but this file is not included in the latest release

treshugart:
Bah! You're right. I'll see what I can do about that tonight. Are you able to ignore it for now?

cronon:
Yes, I can wait a bit, thanks

treshugart:
I'm wondering if we should be publishing types to a separate `types/` folder, or something. We'll have to do this for flow, too.

treshugart:
I've pushed things into a `ts/` folder. Tests and `index.d.ts` files are in there for now.

treshugart:
Hey @cronon, I've decided to remove the TS types for the foreseeable future. It's a pain having to maintain them along side a source not written with them. I'm happy if someone wants to take the types from git history and maintain them in definitely-typed, though. Unfortunately, I don't have time to continue maintaining them. If you have any questions, feel free to ask away.

PR #1255, opened 2017-11-28
fix(release): explicitly include src folder so we get types to npm

Hotell:

* [x] Bug
* [ ] Feature

Closes https://github.com/skatejs/skatejs/issues/1253

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "yocss": "^0.5.1"
   },
   "files": [
+    "src/",
     "es/",
     "esnext/",
     "node/",


treshugart:
I've moved all TS stuff into a `ts/` folder (tests and d.ts files) and updated the `tsconfig.json`. The monorepo now copies the tsconfig to all ts projects (that have a `ts` folder specified in the `files` array) and then runs it. TS tests are now run in CI and I've ensured they're passing.

The only thing to do now might be to have a separate folder for the TS tests? They were in `tests/` but that will be moving into `__tests__/` folders in `src/` soon, so I don't think they should go there. Let's tackle that later, though.

Sorry for making this PR redundant, but thank you for raising it :)

PR #1267, opened 2017-12-02
Fix typos and broken link in readme

blikblum:
## Rationale

Update readme to:
 * fix typos
 * fix broken installing link
 * fix template string expression in custom mixin example
 * update possible values of [renderer] to the correct ones


treshugart:
Thanks @blikblum!

PR #1271, opened 2017-12-12
Clarify lit html renderer

trescenzi:
* [ ] Bug
* [x] Feature

## Requirements

* [x] Read the
* [ ] Wrote tests.

## Rationale

I ran into the problem of attributes no longer rendering when migrating from direct usage of `lit-html`'s `render` in `connectedCallback` to `withLitHtml`. This lead me to realizing that the `withLitHtml` uses the extended render from `lit-html`. I was surprised since the extended renderer's exist isn't really loudly documented by `lit-html`'s readme. This change points to the extended renderer's docs and hopefully prevents others from going on the debugging journey I went on.

I wasn't sure if it made sense to update both the readme and the site but I did both in case that's what you'd want.

## Open Questions

It's possible it makes more sense to link to the file within lit-html's repo and not a specific line since that line reference is liable to become dead. If you know of a place this renderer is documented outside of the code I think it would also make sense to link there but I couldn't find one.

### DIFF
--- a/packages/renderer-lit-html/README.md
+++ b/packages/renderer-lit-html/README.md
@@ -2,6 +2,10 @@
 
 SkateJS renderer for [Lit HTML](https://github.com/PolymerLabs/lit-html).
 
+Please note `renderer-lit-html` uses Lit HTML's extended renderer by default. This provides slightly different
+than default, and enhanced, functionality as described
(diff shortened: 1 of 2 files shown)


treshugart:
Thanks for the contrib!

PR #1277, opened 2017-12-16
renderer-react: wrap via asCustomElement

bedeoverend:
Per [this twitter thread](https://twitter.com/BedeOverend/status/941975889348214785), this PR explores wrapping React components via a separate function exported with `renderer-react`.

Instead of:

```js
import ReactComponent from 'a-react-component';
import withReact from '@skatejs/renderer-react';

export default class extends withReact() {
  static component = ReactComponent
}
```

usage would be:

```js
import ReactComponent from 'a-react-component';
import { asCustomElement } from '@skatejs/renderer-react';

export default asCustomElement(ReactComponent);
```

Currently it doesn't (as this PR just moves from static to an external function), but might be worth having `asCustomElement` export a fully fledged CustomElement that'll mimic the React component's lifecycle methods. I'm guessing for general support of this it'll need to export a class which uses `withUpdate` and `withRenderer`, that way when props update it'll actually trigger a render.

cc @treshugart thoughts?

### DIFF
--- a/packages/renderer-react/src/__tests__/__snapshots__/index.js.snap
+++ b/packages/renderer-react/src/__tests__/__snapshots__/index.js.snap
@@ -6,8 +6,6 @@ exports[`renders 2`] = `"<div data-reactroot=\\"\\"><!-- react-text: 2 -->Hello,
 
 exports[`renders 3`] = `"<div data-reactroot=\\"\\"><!-- react-text: 2 -->Hello, <!-- /react-text --><!-- react-text: 3 -->Bob<!-- /react-text --><!-- react-text: 4 -->!<!-- /react-text --></div>"`;
 
-exports[`wrapper - empty 1`] = `"<span data-reactroot=\\"\\"></span>"`;
-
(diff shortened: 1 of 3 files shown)


treshugart:
> but might be worth having asCustomElement export a fully fledged CustomElement that'll mimic the React component's lifecycle methods

While I'm totally on board with something like this, I think for now, just exporting a custom element for the specified React component is good to start with.

> asCustomElement

Is it worth bike-shedding this name? I don't have any strong opinions, but names like `wrap` and `convert` came to mind. It feels implied that you'd be creating a custom element from a react component.

bedeoverend:
>just exporting a custom element for the specified React component is good to start with.

Sounds good 👍 

And I think `wrap` would be a good choice - you're right, it's strongly implied it will be a custom element that you get back. Also `wrap` feels slightly more like what it gives back than `convert`.

bedeoverend:
If you're happy with all that, feel free to merge. Not sure what's going on with travis..?

treshugart:
I've merged and we'll see about the build. I've raised #1281 to update the docs and release. Thanks!

bedeoverend:
Absolute pleasure! About time I made a contribution to core 😉

treshugart:
> Absolute pleasure! About time I made a contribution to core

Well, it is a monorepo now :p

PR #1283, opened 2017-12-22
Add to with-react renderer showing wrap usage. Fixes #1281

### LINKED ISSUE #1281: Update docs for changes in PR and release
The API changed in: #1277 and we need to updated the docs and release once done. cc @bedeoverend

bedeoverend:
Adds documentation about the `wrap` function - it's pretty minimal, just showing that it exists and a _very_ basic example.

cc @treshugart

### DIFF
--- a/site/pages/renderers/__samples__/wrap-react.js
+++ b/site/pages/renderers/__samples__/wrap-react.js
@@ -1,0 +1,21 @@
+// @jsx React.createElement
+
+import { props, withComponent } from 'skatejs';
+import { wrap } from '@skatejs/renderer-react';
+import React from 'react';
+
+class HelloReact extends React.Component {
+  render() {
+    return <span>Hello, {this.props.name}!</span>;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


site/pages/renderers/__samples__/wrap-react.js
```diff
@@ -0,0 +1,21 @@
...
+import React from 'react';
+
+class HelloReact extends React.Component {
+  render() {
+    return <span>Hello, {this.props.name}!</span>;
+  }
+}
+
+class WrapReact extends withComponent(wrap(HelloReact)) {
+  static get props() {
```
treshugart: This is more a discussion point / note for myself:

We could have another example that shows sharing Flow types for both components and you wouldn't even have to have a `props` getter here.

bedeoverend: 👍 sounds good - would be happy to do, though I have no experience with Flow, so until I've gotten some experience wouldn't be able to contribute on this

Issue #1285, opened 2017-12-27
@skatejs/ssr doesn't work with Jest@22


trescenzi:
## Bug report

@skatejs/ssr doesn't work with Jest@22 even with the addition of `setup` and `teardown` due to the dependency on `jest-enviroment-node@21`. This is because `expect(<mock>).toHaveBeenCalled` now calls `getMockName` which isn't defined in Jest@21. You can see this call [here](https://github.com/facebook/jest/blob/efe65d4d7b265e328f069df57661ba67594d5f9f/packages/expect/src/spy_matchers.js#L64).

* Skate version: @skatejs/ssr@0.18.0

### Current behaviour

`expect(<mock>).toHaveBeenCalled` throws `received.getMockName is not a function` when using with Jest@22.

### Expected behaviour

Honestly not sure. I'm fine with `@skatejs/ssr` not working with Jest@22 but currently it seems like it's half supported so I wanted to document that. Willing to contibute whatever is decided here but I'm certain if sticking with Jest@21 was the plan or upgrading was.

treshugart:
Thanks for reporting. We should make it compatible with Jest 22, for sure. Happy to accept a PR if you have one in flight.

trescenzi:
There's a few different ways it could be done not sure if you have a preference. We could update the dep on `jest-enviroment-node` to be `@22` which would probably be a breaking change if that matters. Or we could update `jest-environment-node` to be `^21 || ^22`. That probably makes the most sense if we want to keep compatibility with `21`. I _think_ that would work as expected.

I've been working with the `ssr` package to add `cloneNode`(for templates) and am close to a PR. I could add whatever you think makes sense for this to that PR which I hope to have up tomorrow morning.

treshugart:
> I've been working with the ssr package to add cloneNode(for templates) and am close to a PR. I could add whatever you think makes sense for this to that PR which I hope to have up tomorrow morning.

That sounds sensible to me :). Thanks for hacking on it!

PR #1286, opened 2018-01-02
Ssr improvements

trescenzi:
* [x] Fixes:#1285
* [x] Feature: Add `cloneNode`
* [x] Fixes broken `content`

## Requirements

* [x] Read the
* [x] Wrote tests.
* [x] Updated TS definitions, if necessary.

## Rationale

This PR does a few things let me know if you'd like me to break it up.

- Firstly this enables usage of either jest 21 or jest 22. This required an update to the `jest-environment-node` dep as well as working around a [breaking change](https://github.com/facebook/jest/blob/master/CHANGELOG.md#features-3).
- Secondly this adds a thin `cloneNode` implementation. This implementation *should* be sufficient for 99% of cases. I'm sure there are some edge cases where the duplication won't be sufficient however I don't know what those might be without doing more serious testing.
- Thirdly this fixes a bug with `.content` which was causing it to return an incomplete `documentFragment`. This is because `appendChild` disconnects the child from the parent. We don't want this behavior for 2 reasons:
  1. If you disconnect a child while iterating the list you'll fail to iterate the whole list.
  1. After using `content` the element will no longer have children.

## Implementation

`cloneNode` was implemented as faithfully as I could given the simplicity of undom. As mentioned above there's a few known oddities such as:

- `innerText` is copied as well as any actual text nodes. This is because `innerText` isn't actually handled in any particular way aside from being a static property. Unlike in a real DOM where `innerText` creates a text node child and replaces the content. Possibly this is something that should be added either in this or another PR.
- Given the few number of node types it's likely that there are specializations that are being missed here.

## Tasks

I've still got to update docs and the TS definitions.

### DIFF
--- a/packages/ssr/jest.js
+++ b/packages/ssr/jest.js
@@ -12,7 +12,14 @@ module.exports = class extends NodeEnvironment {
   setup() {
     return Promise.resolve();
   }
+
+  // Jest@21 uses `dispose` Jest@22 uses `teardown`.
+  // To support both we provide both.
+  // See the Jest@22 Changelog: https://github.com/facebook/jest/blob/master/CHANGELOG.md#features-3
   teardown() {
     return Promise.resolve();
+  }
+  dispose() {
+    return this.teardown();
   }
 };
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "babel-register": "^6.26.0",
     "glob": "^7.1.2",
     "highlight.js": "^9.12.0",
-    "jest-environment-node": "^21.1.0",
+    "jest-environment-node": "^21.1.0 || ^22",
     "js-beautify": "^1.7.3",
     "lit-html": "^0.7.1",
     "marked": "^0.3.7",
--- a/packages/ssr/package.json
+++ b/packages/ssr/package.json
@@ -3,7 +3,7 @@
   "dependencies": {
     "babel-register": "^6.26.0",
     "glob": "^7.1.2",
-    "jest-environment-node": "^21.1.0",
+    "jest-environment-node": "^21.1.0 || ^22",
     "js-beautify": "^1.7.3",
     "mkdirp": "^0.5.1",
     "outdent": "^0.4.1",
--- a/packages/ssr/register/Node.js
+++ b/packages/ssr/register/Node.js
@@ -69,6 +69,34 @@ Object.defineProperty(NodeProto, 'textContent', {
   }
 });
 
+NodeProto.cloneNode = function(deep) {
+  let clone;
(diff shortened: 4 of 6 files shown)


packages/ssr/jest.js
```diff
@@ -12,7 +12,14 @@ module.exports = class extends NodeEnvironment {
...
     return Promise.resolve();
   }
+
+  // Jest@21 uses `dispose` Jest@22 uses `teardown`.
+  // To support both we provide both.
+  // See the Jest@22 Changelog: https://github.com/facebook/jest/blob/master/CHANGELOG.md#features-3
   teardown() {
     return Promise.resolve();
   }
+  dispose() {
```
treshugart: Do you think it would be better to call `this.dispose()` from `teardown()`? It makes it more clear that `dispose()` is the future implementation, and if anything changes, we only need to change it there.

treshugart:
It looks like we need to bump jest-environment-node because of the dependency mismatch. See: https://travis-ci.org/skatejs/skatejs/builds/323971638. The monorepo tool ([Bolt](https://github.com/boltpkg/bolt)) enforces a strict dependency match across all packages.

Do you still think that's a breaking change for us so long as we export both v21 and v22 callbacks?

packages/ssr/jest.js
trescenzi: Yea I think I like that. 👍

trescenzi:
I don't think this is breaking. I'm pretty confident that a user should be able to upgrade to this version and stick with either jest 21 or jest 22.

Two questions:
- Are there ts definitions for ssr? I can't find them.
- What's your opinion on documentation for these changes? Since ssr doesn't currently document what parts of the DOM API it implements I'm unsure anything needs to be added here.

packages/ssr/jest.js
trescenzi: Oh so I made this change and then realized that I think your comment is reversed. Future is `teardown` so call it from `dispose`.

treshugart:
No TS defs to update. Need to remove that from the GH template.

Nothing to add to docs in terms of support, although we should add a support table eventually.

packages/ssr/jest.js
treshugart: Put the logic in whatever jest 22 uses I reckon.

trescenzi: 👍

trescenzi:
Cool then I think this is good to good assuming you're good with the rest of it.

treshugart:
Quality! Thanks, @trescenzi!

treshugart:
Published 0.19.0!

Issue #1318, opened 2018-01-21
How to know which props changed with withUpdated?


trusktr:
I'm trying to use `withUpdated`, and like with `attributeChangedCallback`, I'd like to know which prop was updated.

Should I just use attributeChangedCallback then?

I think if I do,

```js
attributeChangedCallback(attr, oldVal, newVal) {
  super.attributeChangedCallback(attr, oldVal, newVal) // triggers withUpdate sync

  // check `attr` here
}
```

then I can check which prop was changed, and I think `updated` should've already been called at that point right?

But then I feel like I'm side stepping the API a little. It'd be nice if we knew what changed in the updated API.

treshugart:
The `attributeChangedCallback` will sync props, thus queuing a call to `updating`, `shouldUpdate` and `updated`, in that order. They are invoked once in a micro task to batch changes. I recommend using `updated` for capturing all changes and `updated` for rendering to the DOM. All methods receive `prevProps` and `prevState` as arguments.

trusktr:
> All methods receive prevProps and prevState as arguments.

This means I have to diff it myself to find what changed.

I'm making that tool I mentioned in another thread to generate elements from Vue components, and I want to update only the Vue instance's props that have changed. This is important because it means the implementation can withstand future updates to Vue APIs (f.e. re-rendering only parts of a template based on which props have changed).

This is simple to do with attributeChangedCallback, because I know which prop to update.

I'll post the tool here when finished, so you can see what I mean, almost done...

---

Maybe the changed attributes can he recorded in a `Set`, and passed to `updated`

treshugart:
> This means I have to diff it myself to find what changed.

There's plenty of libraries out there that can easily do this and then you can just do `diff(this.props, prevProps)`. This is really all you'd need to do, though:

```js
const diff = Object.keys(this.props).reduce((prev, curr) => {
  if (this.props[curr] !== prevProps[curr]) prev.push(curr);
  return prev;
}, []);
```

If Skate did this on every render just in case you needed the, it would cause an unnecessary performance hit. It's better if users opt-in to doing things like this.

trusktr:
We could make it easier with something like a list created without a loop:

> the changed attributes can he recorded in a `Set` [inside `attributeChangedCallback`], and passed to `updated`

and the idea is here: https://github.com/skatejs/skatejs/pull/1321 (not tested yet).

It doesn't involve a loop, the user can optionally iterate on it. What do you think about this?

[...]

PR #1321, opened 2018-01-22
Let users know which props changed (without looping)

trusktr:

* [ ] Bug
* [x] Feature, https://github.com/skatejs/skatejs/issues/1318

## Requirements

* [x] Read the
* [ ] Wrote tests.

## Rationale

_Why is this PR necessary?_

It gives the user a simple array to check which props were changed. This may prevent someone from instead relying on `attributeChangedCallback` which is not batched.

The user can then _optionally_ loop on the changed attributes.

## Implementation

I user an `Array` isntead of a `Set` because I didn't want to worry about TypeScript types, and plus it's probably faster. I'm setting `length = 0` instead of making a new array (avoiding GC).

## Open questions

Not yet.

## Tasks

_List any tasks you need to do here, if any. Delete this section if you don't
need it._

* [ ] If you're okay with the change, then I can finish the Requirements section (docs, tests, etc).

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -82,6 +82,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
     static _props: Object;
 
     _prevProps: Object;
+    _changedProps: Array<string>;
     _prevState: Object;
     _state: Object;
     _syncingAttributeToProperty: null | string;
@@ -96,6 +97,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
 
     static _observedAttributes = [];
     _prevProps = {};
+    _changedProps = [];
     _prevState = {};
     _state = {};
 
@@ -147,6 +149,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       if (super.attributeChangedCallback) {
         super.attributeChangedCallback(name, oldValue, newValue);
       }
+      if (!this._changedProps.includes(name)) this._changedProps.push(name)
       syncAttributeToProperty(this, name, newValue);
     }
 
@@ -167,15 +170,16 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       }
       this._updating = true;
       delay(() => {
-        const { _prevProps, _prevState } = this;
+        const { _prevProps, _prevState, _changedProps } = this;
         if (this.updating) {
-          this.updating(_prevProps, _prevState);
+          this.updating(_prevProps, _prevState, _changedProps);
         }
-        if (this.updated && this.shouldUpdate(_prevProps, _prevState)) {
-          this.updated(_prevProps, _prevState);
+        if (this.updated && this.shouldUpdate(_prevProps, _prevState, _changedProps)) {
+          this.updated(_prevProps, _prevState, _changedProps);
         }
         this._prevProps = this.props;
         this._prevState = this.state;
+        this._changedProps.length = 0;
         this._updating = false;
       });
     }
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -82,7 +82,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
     static _props: Object;
 
     _prevProps: Object;
-    _changedProps: Array<string>;
+    _changedProps: Set<string>;
     _prevState: Object;
     _state: Object;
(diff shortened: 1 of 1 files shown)


treshugart:
> This may prevent someone from instead relying on attributeChangedCallback which is not batched.

It is batched in terms of only happening after all set attribute calls happen in a microtask, but is called for every set.

You say:

> without looping

However, `array.includes()` is essentially a `array.indexOf() > -1` alias, and the algorithmic complexity of `indexOf()`, I believe, is O(log n). It's not crazy, but the code here is doing it on every batched update regardless of if the user needs to know about it.

I really don't think I want to include this in core because it would be simple enough to install / write a utility that does this for you when you do need it. Is it too much to install an object diff function from NPM and use it with `this.props` and `prevProps`?

trusktr:
> It is batched in terms of only happening after all set attribute calls happen in a microtask, but is called for every set.

Yep, exactly, 99.99999% of the time we only care about the final state, f.e. then we use the final state to determine what to render in the next animation frame in case of custom elements for rendering WebGL scenes.

> `array.includes()`

Oops, you're right. I originally meant to use of `Set`, which is mostly O(1) in v8 according to [this](https://stackoverflow.com/a/33614512/454780). Updated the PR. It's MUCH faster. Thought on this?

Here's an interesting benchmark you can run in the console that shows how much faster it is that Array with `.includes()`:

```js
var s = new Set
var NUM_ITERATIONS = 10000
var totalTime = 0
var begin

for (let i=0; i<NUM_ITERATIONS; i+=1) {
  begin = performance.now()
  if (!s.has(i)) s.add(i)
  totalTime += performance.now() - begin
}

console.log('total time', totalTime)
console.log('average time', totalTime/NUM_ITERATIONS)

totalTime = 0
s.clear()

for (let i=0; i<NUM_ITERATIONS; i+=1) {
  begin = performance.now()
  s.add(i)
  totalTime += performance.now() - begin
}

console.log('total time', totalTime)
console.log('average time', totalTime/NUM_ITERATIONS)

totalTime = 0
var a = []

for (let i=0; i<NUM_ITERATIONS; i+=1) {
  begin = performance.now()
  if (!a.includes(i)) a.push(i)
  totalTime += performance.now() - begin
}

console.log('total time', totalTime)
console.log('average time', totalTime/NUM_ITERATIONS)

totalTime = 0
a.length = 0

for (let i=0; i<NUM_ITERATIONS; i+=1) {
  begin = performance.now()
  a.push(i)
  totalTime += performance.now() - begin
}

console.log('total time', totalTime)
console.log('average time', totalTime/NUM_ITERATIONS)
```

> Is it too much to install an object diff function from NPM and use it with this.props and prevProps?

I didn't need to because writing one was easy. So my updated method looks like:

```js
        updated( prevProps, prevState ) {
            super.updated( prevProps, prevState )

            forChangedProps( this.props, prevProps, name => {
                this.vue_instance[ name ] = this[ name ]
            })
        }
```

Using the PR's new `Set` implementation, it'd look like this:

```js
        updated( prevProps, prevState, changedProps ) {
            super.updated( prevProps, prevState, changedProps )

            for (const prop of changedProps) {
                this.vue_instance[ prop ] = this[ prop ]
            }
        }
```

and it'll iterate quickly on only the number of props that were modified! Woot woot!!

trusktr:
Here's a test that shows iterating on a Set is almost the same performance as iterating on an Array:

```js
var a = []
var s = new Set
var NUM = 1000000

for (let i=0; i<NUM; i+=1) a.push(i)
for (let i=0; i<NUM; i+=1) s.add(i)

var totalTime = 0

begin = performance.now()
for (let n of a) {
  n
}
totalTime += performance.now() - begin

console.log('total time', totalTime)

var totalTime = 0

begin = performance.now()
for (let n of s) {
  n
}
totalTime += performance.now() - begin

console.log('total time', totalTime)
```

So the Set implementation is very light for the convenience it offers. Still don't like it?

trusktr:
Interesting, changing to `Set` somehow broke the lit renderer test?

treshugart:
I fixed the lit-html issue causing the build failure last night. Pull and see if that fixes things for you.

trusktr:
Looks like checks passed!

treshugart:
> So the Set implementation is very light for the convenience it offers. Still don't like it?

It's not that I don't like it, it's more that I just don't think it's something that Skate should be concerned with. There's plenty of libraries out there that can give you a list of properties that are different between two objects.

I apologise and 100% value the ideas you bring to the table but I don't feel comfortable adding this feature without more people wanting it who can justify why this is hard to do out side of Skate. Cheers for the PR, though.

Issue #1320, opened 2018-01-21
[feature] reactive props


trusktr:
## Feature request

When using `withUpdate`, make `this.props` reactive similar to properties on Vue instances.

### Use case(s)

I can currently trigger an update by soing `el.setAttribute('foo', 45)`, and if the prop type for foo is props.number, then it sets `this.props.foo` to the number `45` as expected.

But I was also hoping that I could do something like `this.props.foo = 123` and that that would also trigger an update.

If you have an instance of a Vue component, you can do `instance.foo = 123` and it triggers a re-render of the component.

I haven't tried this with `this.state`, but I am only testing `this.props` because that is the public interface, and in Vue props are also the public interface and they are reactive, which I really like.  (Vue doesn't have an additional `state`, only `props` which are directly on the instance).

This would be cool feature!

trusktr:
I see in the docs:

> Attribute binding is one-way: attribute updates affect the property, but not the other way around.

I think that confirms that using `setAttribute` is the only way to supply new values via JavaScript (which includes the [`number-to-string-to-number` problem](https://github.com/skatejs/skatejs/issues/1317)).

trusktr:
Hmmm, the "Guide" section of the docs says

> `withUpdate` -- the generated element will react to changes on their props or HTML attributes.

This seems to imply that I can change either the attribute **or** the props, but in my experience changing values inside of `elem.props` isn't doing anything, only `setAttribute` is working, so I'm not sure if this is a feature request or a bug report. Which one might it be?

treshugart:
If you have:

```js
import { props, withComponent } from 'skatejs';

class MyComponent extends withComponent() {
  static props = {
    foo: props.number
  }
}
```

You can:

1. `element.foo = 123;`
2. `element.props = { foo: 123 };`
3. `element.setAttribute('foo', '123');`

Where it says:

> Attribute binding is one-way: attribute updates affect the property, but not the other way around.

This means that property updates don't affect attributes. This is intentional because if you mutate an attribute, it might mess with any state a parent renderer has over the host element. This is consistent with DOM props that maintain internal state such as `value` on inputs.

You can enable two-way binding if you want by specifying `{ attribute: true }` to your `props`:

```js
import { props, withComponent } from 'skatejs';

class MyComponent extends withComponent() {
  static props = {
    foo: { ...props.number, attribute: true }
  }
}
```

[...]

PR #1332, opened 2018-01-23
Remove return value from sample, it's not doing anything

trusktr:
I was looking at where to add [this detail](https://github.com/skatejs/skatejs/issues/1320#issuecomment-359957003) to the docs, but I'm a little confused by the first example. The update method returns a value, but it's not clear why anything needs to be returned.

Looking at with-update.js implementation, it doesn't do anything with a return value:

https://github.com/skatejs/skatejs/blob/master/packages/skatejs/src/with-update.js#L175

Maybe it's better to remove the return?


treshugart:
Good call. Cheers!

PR #1355, opened 2018-02-06
IE11 fix for preact renderer

klipstein:
## Requirements

* [x] Read the
* [ ] Wrote tests.

## Rationale

Preact Renderer does not work in IE11.

## Implementation

`children` is not supported in IE11, falling back to `childNodes` instead.

### DIFF
--- a/packages/renderer-preact/src/index.js
+++ b/packages/renderer-preact/src/index.js
@@ -13,8 +13,6 @@ export default (Base = HTMLElement) =>
       };
     }
     renderer(root, call) {
-      const children = root.children ||
-        [].map.call(root.childNodes, node => node.nodeType === 1);
-      this._preactDom = render(call(), root, this._preactDom || children[0]);
+      this._preactDom = render(call(), root, this._preactDom || root.childNodes[0]);
     }
   };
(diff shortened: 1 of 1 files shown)


packages/renderer-preact/src/index.js
```diff
@@ -13,10 +13,8 @@ export default (Base = HTMLElement) =>
       };
     }
     renderer(root, call) {
-      this._preactDom = render(
-        call(),
-        root,
-        this._preactDom || root.children[0]
-      );
+      const children = root.children ||
```
treshugart: Do you think it would be safe to just do `childNodes[0]` in place of `children[0]`, instead of having to do the map? Also, do you mean to `filter` here instead?

treshugart:
Thanks for the PR! I just left a comment. Would like to discuss prior to merging, if that's alright.

packages/renderer-preact/src/index.js
klipstein: Ups, meant `filter`. I guess using `childNodes[0]` makes the most sense here. Will change it.

treshugart:
So weird. I could have sworn children was available and it seems it is: https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/children. I'll merge this for now to get you working and push a release but it would be good to see what the core issue was.

treshugart:
pushed 0.2.1

klipstein:
I guess it was because of this weird issue: https://stackoverflow.com/questions/39288373/document-children-vs-document-childnodes

PR #1367, opened 2018-02-12
Compatibility with babel-transform-builtin-classes

trusktr:
This is an incomplete pull request. It just shows how to make the lib compatible with babel-plugin-transform-builtin-classes (and therefore not require `webcomponentsjs`'s adapter).

This would need to be applied anywhere else that a native classes is being used.

What this does is enable transform-builtin-classes to work (it reads the class syntax), so that it will be able to patch the native extension.

End users will still have to configure the plugin to look for `HTMLElement`, and other native classes. We could also supply an example babelrc in this repo.

I haven't checked, but if all uses of the native class can be supplied from the end user, then it might not be necessary.

So, just opening this here in case it is helpful, but not sure if you actually want to do all these changes.

---

Also note, Babel 7 will [include this transform builtin](https://github.com/babel/babel/pull/7020), but at the moment it [doesn't work](https://github.com/babel/babel/pull/7020#issuecomment-362113864), or at least, there's no indication of what I have to do to make it work.

### DIFF
--- a/packages/renderer-lit-html/src/index.js
+++ b/packages/renderer-lit-html/src/index.js
@@ -1,6 +1,6 @@
 import { render } from 'lit-html/lib/lit-extended';
 
-export default (Base = HTMLElement) =>
+export default (Base = class extends HTMLElement {}) =>
   class extends Base {
     renderer(root, call) {
       render(call(), root);


rzymek01:
I make it work in the following browsers:
Chrome 65.0.3325.181
FF 57.0b12
FF 60.0b10
Safari Version 11.1 (13605.1.33.1.2)
Edge 16
IE 11 seems it should work but run into another issue: https://github.com/webcomponents/custom-elements/issues/135


`.babelrc` config:

```
{
  "presets": [
    [
      "env", {
        "modules": false,
        "debug": true
      }
    ],
    "react",
    "flow"
  ],
  "plugins": [
    "transform-class-properties",
    "transform-object-rest-spread",
    [
      "babel-plugin-transform-builtin-classes", {
        "globals": ["HTMLElement"],
        "logIfPatched": true
      }
    ],
    "syntax-dynamic-import",
    [
      "babel-plugin-styled-components", {
        "minify": false,
        "uglifyPure": false
      }
    ]
  ],
  "env": {}
]
```

In package.json:

    "babel-plugin-transform-builtin-classes": "^0.6.1",

In component source:

     export default define(
       class extends withComponent(class extends HTMLElement {}) { 


And including only https://github.com/webcomponents/custom-elements polyfill (w/o any adapters etc).

treshugart:
With custom elements on the horizon, and class support everywhere except IE11, I think I'd prefer to close this for now but it'd be good to document it and link to the PR. Happy to repopen if you disagree and we can discuss. Somethings I'd like if we were to accept this:

1. Docs on Babel config (v6 and v7)
2. Doing this everywhere we use `HTMLElement`

trusktr:
At the moment, I get around this issue by using [`lowclass`](https://github.com/trusktr/lowclass) at runtime which is an alternative to using `babel-plugin-transform-builtin-classes` at build time (which adds runtime output that does a similar thing).

treshugart:
Since my last comment, things are now in TypeScript. I'm not sure how they handle this but hope to look into it soon. Definitely prior to releasing.

PR #1397, opened 2018-02-22
Add a withInitializer mixin to allow one-time setup of components after they are defined 

chrisdarroch:
Based on a twitter exchange: https://twitter.com/treshugart/status/965907794556678144

* [ ] Bug
* [x] Feature

## Requirements

* [x] Read the
* [x] Wrote tests.

## Rationale

A change like this allows component authors to perform one-time setup work once their components have been defined by the browser, without having to know the element's name ahead of time.

## Implementation

We're exploiting the nature of how the `observedAttributes` method is invoked by browsers upon custom elements being registered.

Future implementations could be made more robust if we switched behaviour based on the presence of an `is` attribute; in that case we could call `customElements.whenDefined` directly and avoid this little exploit.

## Open questions

I'm not sure how sensible this is. I have a few conversation points:

* [ ] Is `withInitializer` and `initializeCallback` a sensible name? What others might make more sense?
* [x] Decide how `withUpdate` should interact with the return value of the `observedAttributes` method defined by `withInitializer`, if it needs to at all.
* [ ] Decide whether an addition of a naked `observedAttributes` method to a custom element class extended from `withInitializer` should work if they don't call `super.observedAttributes()`.

## Tasks

* [ ] Review the way elements are defined in tests; decide if they reasonably cover the real world.
* [ ] Add a working example to the docs, ideally based off of a real-world use-case.

### DIFF
--- a/packages/skatejs/src/with-initializer.js
+++ b/packages/skatejs/src/with-initializer.js
@@ -3,8 +3,9 @@
 export const withInitializer = (Base: Class<any> = HTMLElement): Class<any> =>
   class extends Base {
     static get observedAttributes() {
-      super.observedAttributes && super.observedAttributes();
+      let observed = super.observedAttributes;
       this.initializeCallback();
+      return observed;
     }
     static initializeCallback() {
       super.initializeCallback && super.initializeCallback();
--- a/packages/skatejs/src/with-initializer.js
+++ b/packages/skatejs/src/with-initializer.js
@@ -2,11 +2,11 @@
 
 export const withInitializer = (Base: Class<any> = HTMLElement): Class<any> =>
   class extends Base {
-    constructor() {
-      super();
+    static get observedAttributes() {
+      super.observedAttributes && super.observedAttributes();
       this.initializeCallback();
     }
-    initializeCallback() {
+    static initializeCallback() {
       super.initializeCallback && super.initializeCallback();
     }
   };
--- a/packages/skatejs/src/with-initializer.js
+++ b/packages/skatejs/src/with-initializer.js
@@ -1,0 +1,12 @@
+// @flow
+
+export const withInitializer = (Base: Class<any> = HTMLElement): Class<any> =>
+  class extends Base {
+    constructor() {
+      super();
+      this.initializeCallback();
+    }
+    initializeCallback() {
+      super.initializeCallback && super.initializeCallback();
+    }
+  };
--- a/packages/skatejs/src/index.js
+++ b/packages/skatejs/src/index.js
@@ -8,6 +8,7 @@ export * from './shadow.js';
 export * from './with-children.js';
 export * from './with-context.js';
 export * from './with-component.js';
+export * from './with-initializer.js';
 export * from './with-lifecycle.js';
 export * from './with-update.js';
 export * from './with-renderer.js';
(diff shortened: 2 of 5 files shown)


packages/skatejs/src/with-initializer.js
```diff
@@ -3,8 +3,9 @@
 export const withInitializer = (Base: Class<any> = HTMLElement): Class<any> =>
   class extends Base {
     static get observedAttributes() {
-      super.observedAttributes && super.observedAttributes();
+      let observed = super.observedAttributes;
```
treshugart: Something I just noticed: what happens when this is retrieved explicitly? i.e. not via `customElements.define()` - either before or after?

treshugart:
Commented to see if we can flesh out what happens if it's not invoked via `customElements.define()`. I'd love for this to go in but we might want to have a good story for that.

In terms of naming, we've gone a way from the `*Callback()` convention mostly. Maybe `static init()`? Or if we want to reflect the timing, maybe `static defined()`?

packages/skatejs/src/with-initializer.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+export const withInitializer = (Base: Class<any> = HTMLElement): Class<any> =>
+  class extends Base {
+    static get observedAttributes() {
+      let observed = super.observedAttributes;
+      this.initializeCallback();
```
treshugart: I'd like to pull in the comment used in the `observedAttributes` implementation and put it somewhere in here (up to you where it goes).

packages/skatejs/src/with-initializer.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+export const withInitializer = (Base: Class<any> = HTMLElement): Class<any> =>
```
treshugart: I think we should remove the implementation of this in `with-update` and explicitly use this mixin in it to do the props initialisation.

packages/skatejs/src/with-initializer.js
```diff
@@ -0,0 +1,13 @@
...
+  class extends Base {
+    static get observedAttributes() {
+      let observed = super.observedAttributes;
+      this.initializeCallback();
+      return observed;
+    }
+    static initializeCallback() {
```
treshugart: I've been thinking about names and I've undid the `Callback` suffix on all Skate specific stuff. I'd also like it to correspond to the `whenDefined()` functionality. Given that, I was going to suggest we call this `static defined()`. Thoughts?

chrisdarroch: `defined` has meaning in the basic web component spec, right? Unless this meaning is the same as that meaning, I don't think they should be named similarly.

chrisdarroch:
my ability to action your feedback is limited right now. I don’t have any special attachment to the code I wrote, so you’re most welcome to polish off what I started and merge it if you’re happy with it @treshugart :)

packages/skatejs/src/with-initializer.js
treshugart: Defined is specifically what we're trying to hook into without having to know the name and using `whenDefined()`.

treshugart:
That's fair. Nobody is screaming for this (we are the only use case, and we've already solved it within the `updated` mixin). For those reasons, I'm unsure if it's worth adding this feature. It's definitely the right way to abstract it, but not sure if adding to the api surface area is the right thing to do.

treshugart:
I think I'll close this until we have more than one use case for hooking into defined (without knowing the name).

PR #1410, opened 2018-02-28
Add missing file extensions

leofavre:
* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests. (*not necessary, the current tests apply*)

## Rationale

Adding file extensions to javascript imports makes them compatible with native implementations on modern browsers. This way, one can directly import `dist/esnext/index.js` in their project without running Babel ou Rollup.


leofavre:
Sorry about that.

PR #1432, opened 2018-04-17
Unmount React component when disconnected. Closes #1260

### LINKED ISSUE #1260: Unmount react component when disconnected
Do the same thing as: https://github.com/skatejs/renderer-preact/issues/5

NMinhNguyen:
Closes #1260.

* [x] Bug
* [ ] Feature

## Requirements

* [x] Read the
* [x] Wrote tests.

## Rationale

_Why is this PR necessary?_

## Implementation

I based the implementation on skatejs/renderer-preact#5.
For the test, I looked at [how](https://github.com/facebook/react/blob/a442d9bc082878ccf66872b1eeed3465927801b6/packages/react-dom/src/__tests__/ReactMountDestruction-test.js#L34-L36) React tests `unmountComponentAtNode` and [how](https://github.com/facebook/react/blob/a442d9bc082878ccf66872b1eeed3465927801b6/packages/react-dom/src/__tests__/ReactMount-test.js#L80-L106) they spy on lifecycle hooks.

## Open questions

* Is it safe to store `renderRoot` on `this._renderRoot`?
* inside `disconnectedCallback`, should `super.disconnectedCallback()` be called first (the way it is in skatejs/renderer-preact#5) or last?


treshugart:
> Is it safe to store renderRoot on this._renderRoot? I didn't want to rely on this.renderRoot being provided via withRenderer, but perhaps it's fine to require withReact to always be used with withRenderer? It does mean that I'd either have to mock it in tests, or couple it to withRenderer though.

Happy for you to rely on `renderRoot`.

> I've done some googling and seems like the super call is always first.

👍 

> is there any reason why the current renderer-preact in the monorepo does not unmount, unlike the old repo?

Hmm, I don't remember and can't find anything on a quick search of issues. Would you mind reimplementing this? It should just be a copy pasta, which is okay for now.

I'll leave this PR as is and await your response to discuss how we should proceed.

NMinhNguyen:
Thanks for your review!

> Happy for you to rely on `renderRoot`.

What's the best way to expose that in the test though? Should I use it with `withRenderer`? Or give it a mock class that has a dummy `get renderRoot()`?

> t should just be a copy pasta

I guess you'd want me to also use `renderRoot` there too? But sure, no problem :)

NMinhNguyen:
Right, so I cherry-picked the original change from the Preact renderer (to preserve authorship). I then tried to naively change `this._renderRoot` to `this.renderRoot`, but obviously without `withRenderer` or `withComponent` it is `undefined`. I then tried to wrap everything in `withRenderer` in the tests, but that meant having to change the first argument to `el.renderer(el` to be `el.renderer(el.renderRoot`, as well as changing all `el.innerHTML` references to be `el.renderRoot.innerHTML`. I've pushed the commit (9bd696b) anyway, to see what you think, but I'm not sure I like it 😕

packages/renderer-react/src/index.js
```diff
@@ -1,6 +1,5 @@
 import React from 'react';
 import { render, unmountComponentAtNode } from 'react-dom';
-import { withRenderer } from 'skatejs';
```
NMinhNguyen: This `skatejs` didn't seem to be used in this module, so I decided to remove it, along with the dependency on `skatejs` from `package.json`

packages/renderer-preact/src/__tests__/index.js
```diff
@@ -1,9 +1,10 @@
 /** @jsx h */
 
 import { Component, h } from 'preact';
-import withRenderer from '..';
+import { withRenderer } from 'skatejs';
+import withPreact from '..';
 
-class MyElement extends withRenderer() {
+class MyElement extends withRenderer(withPreact()) {
```
treshugart: I don't mind using `withRenderer` here in the tests. It simplifies it a bit to do so. Another way is, like you said, to just define it on the class and return a value it can use.

treshugart: Why didn't this work before? Or did it? If you remove `withRenderer`, and the accessing of `renderRoot`, everything works, right?

NMinhNguyen: You mean, just use the original code where you store `root` on `this._renderRoot`? I only updated it for consistency with the React renderer really, but I'm happy to revert.

NMinhNguyen: Interestingly, the current tests still pass with the accessing of `renderRoot`, would you rather I reverted these last changes to the Preact renderer?

NMinhNguyen: I've reverted the `renderRoot` changes to `renderer-preact`

treshugart:
Hey @NMinhNguyen I've updated the tests in c8b0006b to be as consistent as possible. Hopefully this simplifies the work you have to do. I don't think I originally grokked your intention and probably responded slightly out of context.

treshugart:
> Or just store it on an underscore-prefixed instance property like in the Preact renderer, so that it can be referenced when unmounting?

Right, this is the part I misunderstood, apologies. I thought you meant in the tests. For the unmounting you can just store it in the underscore prefixed like the Preact renderer. Reusing `renderRoot` isn't possible because, like you said, it would assume the usage of `withRenderer()`.

Thanks again, and sorry for the run-around.

NMinhNguyen:
@treshugart ah I think I might know where your confusion's coming from - it doesn't help that in the renderer tests, each renderer is called `withRenderer`, rather than `withReact` or `withPreact` 😅 Should in theory be as simple as reverting that last commit.

packages/renderer-preact/package.json
```diff
@@ -3,7 +3,8 @@
   "browser": "dist/es/index.js",
   "description": "A SkateJS renderer for Preact.",
   "devDependencies": {
-    "preact": "8.2.7"
+    "preact": "8.2.7",
+    "skatejs": "^5.1.2"
```
NMinhNguyen: Added this since the tests are now pulling in `skatejs` in https://github.com/skatejs/skatejs/commit/c8b0006bf70eab9790f2e7169c68f5468b800618#diff-61ecfdc8a7d04b1bf1fbefe49953d7feR4

packages/renderer-react/src/index.js
```diff
@@ -13,8 +12,14 @@ const withReact = (Base = HTMLElement) =>
       };
     }
     renderer(root, call) {
+      this._renderRoot = root;
       render(call(), root);
     }
+    disconnectedCallback() {
+      super.disconnectedCallback && super.disconnectedCallback();
+      unmountComponentAtNode(this._renderRoot);
+      this._renderRoot = null;
```
NMinhNguyen: kinda got this habit from React where `ref`s are nulled out when unmounting - prob a good idea to prevent any memory leaks. Could also add this line to the Preact renderer if you want?

NMinhNguyen:
Ready for another review. Will happily address any further comments.

packages/renderer-react/src/index.js
treshugart: Yeah, that'd be great. I'll merge this to get it in and we can do that as a separate change. Should we test that it's nulled out you think?

NMinhNguyen: Sure, can do. It feels almost like testing an implementation detail (`expect(el._renderRoot).toBeNull()`), but perhaps that's not so bad

packages/renderer-preact/src/index.js
```diff
@@ -13,10 +13,17 @@ export default (Base = HTMLElement) =>
...
       this._preactDom = render(
         call(),
         root,
         this._preactDom || root.childNodes[0]
       );
     }
+    disconnectedCallback() {
+      super.disconnectedCallback && super.disconnectedCallback();
+      // Preact hack https://github.com/developit/preact/issues/53
+      const Nothing = () => null;
```
developit: I believe Preact will happily accept `null` as the first argument to `render()` these days, the component should no longer be needed.

Here's an example: https://jsfiddle.net/developit/xngr7pm1/

PR #1434, opened 2018-04-23
Ensure the (p)react renderers do not hold on to renderRoot post disco…

NMinhNguyen:
…nnect.

See https://github.com/skatejs/skatejs/pull/1432#discussion_r183269151

Supersedes #1433 (didn't realise force pushing prevented PRs from reopening)

* [x] Bug
* [ ] Feature

## Requirements

* [x] Read the
* [x] Wrote tests.

## Rationale

_Why is this PR necessary?_

I think it's generally a good idea to null out any references after you're done using them, especially if they're DOM element references, to avoid any memory leaks.

## Implementation

Had to add a guard in the React renderer - otherwise the new `re-connecting & disconnecting without rendering` test case fails.

## Open questions

Decided to experiment with re-mounting using the Preact renderer and ran into issues (see https://github.com/NMinhNguyen/skatejs/commit/cc577bc382d93bd62543e02f818c7ecfbb3207ae#r28691709) whereby `el.innerHTML` doesn't update after a re-mount 😕 It is very likely to do with the "unmounting hack" introduced in https://github.com/skatejs/skatejs/commit/2de8a547ffb15c1c1987a2c37e5d7a7e85652886. That said, similar tests for the React renderer seem to pass. Gonna have to investigate what's going on in the Preact case.

### DIFF
--- a/packages/renderer-preact/src/__tests__/index.js
+++ b/packages/renderer-preact/src/__tests__/index.js
@@ -74,5 +74,6 @@ test('wrappers cleanup', () => {
   root.removeChild(el);
 
   expect(el.innerHTML).toMatchSnapshot();
+  expect(el._renderRoot).toBeNull();
   expect(willUnmountSpy).toHaveBeenCalled();
 });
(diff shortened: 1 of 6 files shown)


treshugart:
I'll merge this and we can work on the preact one separately.

PR #1438, opened 2018-05-01
Preact autodefine vnode

treshugart:
Implements #1437


packages/renderer-preact/src/index.js
```diff
@@ -1,7 +1,27 @@
 /** @jsx h */
 
+import { define } from 'skatejs';
 import { h, render } from 'preact';
 
+let oldVnode;
+
+function setupPreact() {
+  oldVnode = preact.options.vnode;
+  preact.options.vnode = vnode => {
```
treshugart: This should be extracted out of this function scope so it's not re-created every time.

PR #1440, opened 2018-05-12
ssr: fix setting textContent when node has child nodes

blikblum:
* [x] Bug
* [ ] Feature

## Requirements

* [x] Read the
* [x] Wrote tests.

## Rationale

When a html node already has content (child nodes), textContent is appending the text instead of replacing

## Implementation

It removes the child nodes before adding the new one


treshugart:
So good! Thanks for the fix!

treshugart:
Released in 0.19.3!

PR #1441, opened 2018-05-12
Add snabbdom-renderer

blikblum:
* [ ] Bug
* [x] Feature

## Requirements

* [x] Read the
* [x] Wrote tests.

## Rationale

Adds snabbdom renderer as an option to skatejs users

## Implementation

It exports a factory function that creates a renderer mixin. It accepts as argument an array
of snabbdom modules, the same signature of snabbdom init function

This is necessary to allow the user to configure the used modules

The alternative would be hardcode the default modules which would prevent usage of custom modules or even discarding some default ones 

## Other

While testing i spotted a bug when setting textContent in ssr
The snapshot tests are created with the fix applied so, if test is run before #1440 is merged, the snapshot will not match

## Tasks

* [ ] Define if setting a children prop by default is necessary.
   The v5 renderers set a children prop (basically a slot element) by default. In v6 is not set anymore
   In the current implementation, i kept the children prop to align in v5, but since this is the first snabbdom version there's no risk to break compatibility if children is not set
   Personally i'm inclined to not set a children prop by default since most of the times is not used and when necessary is simple enough


treshugart:
I'm going to merge this as is because I just merged the ssr pr you put through (thanks!!!) and it should pass. If not we can address then.

Thanks again for the fantastic contribution!

treshugart:
Re the children prop, I have to get back into the v6 branch and feel around. It's been awhile but it might make sense to keep them in there.

As a side note, if you have time, feel free to poke around in the site and add a docs page for the snabdom renderer. No pressure, though!

treshugart:
Published: https://unpkg.com/@skatejs/renderer-snabbdom

PR #1451, opened 2018-06-15
Allow to define prop type using native types constructors

blikblum:
* [ ] Bug
* [x] Feature

## Requirements

* [x] Wrote tests.

## Rationale

Currently is necessary to import props from skatejs package to define simple props types like string, number etc

Given the skatejs design based on mixins, most of the times, this is the only import from skatejs when defining a component class

By allowing to use native constructors like String, Number the props import could be avoided. It's also pretty intuitive and align with other libraries like [lit-element](https://github.com/Polymer/lit-element/)

With this change is possible to do

```
class Test extends MyComponent {
  static props = {
     str: String,
     obj: Object
  }
}
```

## Implementation

Used a Map instance to store the props definitions mapped to correspondent native constructor

## Open questions

Should props.any be mapped to null?
Should be added a function `registerPropType` to allow  sharing custom properties? 

## Other

To test if props.* match to constructors needed to access private field _propsNormalised in tests
In normaliseAttributeDefinition i created a identity function to be used when necessary instead of creating a anonymous function on the fly  

## Tasks

* [ ] Document

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Skate is a functional abstraction over
 [the web component standards](https://github.com/w3c/webcomponents) that:
 
 * Produces cross-framework compatible components.
-* Abstracts away common attribute / property semantics via `props`, such as
+* Abstracts away common attribute / property semantics via `props` or native types, such as
   attribute reflection and coercion.
 * Adds several lifecycle callbacks for responding to prop updates, rendering and
   updating, as well as a way to manage internal component state.
@@ -60,7 +60,7 @@ import { h } from 'preact';
 class WithPreact extends withComponent(withPreact()) {
   static get props() {
     return {
-      name: props.string
+      name: props.string // String could be used also to define the prop type
     };
   }
   render({ name }) {
--- a/packages/skatejs/test/unit/with-update.spec.js
+++ b/packages/skatejs/test/unit/with-update.spec.js
@@ -86,6 +86,41 @@ test('static get props() {} should define observedAttributes', () => {
   expect(Test.observedAttributes.indexOf('test')).toBe(0);
 });
 
+function comparePropTypes(type1, type2) {
+  expect(type1.coerce).toEqual(type2.coerce);
+  expect(type1.serialize).toEqual(type2.serialize);
+  expect(type1.deserialize).toEqual(type2.deserialize);
+  expect(type1.default).toEqual(type2.default);
+}
+
+test('prop types could be defined with native types constructors', () => {
+  @define
+  class Test extends withUpdate() {
+    static get props() {
+      return {
+        propsString: props.string,
+        nativeString: String,
+        propsBoolean: props.boolean,
+        nativeBoolean: Boolean,
+        propsNumber: props.number,
+        nativeNumber: Number,
+        propsArray: props.array,
+        nativeArray: Array,
+        propsObject: props.object,
+        nativeObject: Object
+      };
+    }
+  }
+  const testEl = new Test();
+  const propsNormalised = testEl.constructor._propsNormalised;
+
+  comparePropTypes(propsNormalised.nativeString, propsNormalised.propsString);
+  comparePropTypes(propsNormalised.nativeBoolean, propsNormalised.propsBoolean);
+  comparePropTypes(propsNormalised.nativeNumber, propsNormalised.propsNumber);
+  comparePropTypes(propsNormalised.nativeObject, propsNormalised.propsObject);
+  comparePropTypes(propsNormalised.nativeArray, propsNormalised.propsArray);
+});
+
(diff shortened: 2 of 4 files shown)


treshugart:
> Should props.any be mapped to null?

Yeah, that's probably a good idea.

> Should be added a function registerPropType to allow sharing custom properties?

How would this work?

blikblum:
I implemented null/undefined map to props.any using it as a fallback in defineProps:
`let func = props[name] || props.any;`

Let me know if you prefer to do the mapping through defaultTypes instance


>> Should be added a function registerPropType to allow sharing custom properties?

> How would this work?

```javascript
// implementation
registerPropType (typeId, definition) {
  defaultTypes.set(typeId, definition)
}

// setup
import { registerPropType } from 'skatejs'

registerPropType('mytype', {...TypeDef})

//usage
class Comp extends Component {
  static props = {
     myProp: 'mytype'
  }
}
```
I think if aggred to implemented, this should be done in a separated PR

packages/skatejs/test/unit/with-update.spec.js
```diff
@@ -98,6 +98,9 @@ test('prop types could be defined with native types constructors', () => {
   class Test extends withUpdate() {
     static get props() {
       return {
+        propsAny: props.any,
```
treshugart: I'm actually fine if we assert something like:

```js
expect(Component.props.nativeNull).toMatchObject(props.any);
```

And remove direct testing of `props.any` in withUpdate (and of course continue testing them separately as a unit).

treshugart: I'd also like to start advocating the usage of `null`, `String`, `Array`, `Object`, etc over the `props.*` as it's definitely closer to the metal in how one might expect it to behave. Custom usage would involve the `props.*` API.

treshugart: This might be out of the scope of this PR and we can do it in 6, though. For now, I think I'm happy to see this as is, but your thoughts here would be useful.

treshugart:
I'm not sure if we need that API. I think we should be covering all built-ins with pre-configured prop types (like we are). Any edge cases, consumers can just write custom prop definitions, as they can now. I'm generally not a fan of having a function that mutates the global configuration of a library.

packages/skatejs/test/unit/with-update.spec.js
blikblum: > expect(Component.props.nativeNull).toMatchObject(props.any);

As far as i remember this did not work because of the prop name saved in prop definition. I can try again and put the error here

> And remove direct testing of props.any in withUpdate (and of course continue testing them separately as a unit).

Do you mean creating a separated spec file?

> This might be out of the scope of this PR and we can do it in 6, though. For now, I think I'm happy to see this as is, but your thoughts here would be useful.

Agree

treshugart: I'm happy to keep as-is for the short-term.

treshugart:
What work is left here? It still looks like you need to implement the mapping between native / `props.*`?

blikblum:
> What work is left here? It still looks like you need to implement the mapping between native / props.*?

The mapping (Array -> props.array etc) is already done. 

Needs to update the docs

blikblum:
Updated docs

treshugart:
> Also there's a misspelling normalised -> normalized but initially i think better todo in a separated PR

Oh, ha! That's "proper" English (UK / AU). I moved from California to Sydney, so I switch between the two without knowing it.

README.md
```diff
@@ -60,7 +60,7 @@ import { h } from 'preact';
 class WithPreact extends withComponent(withPreact()) {
   static get props() {
     return {
-      name: props.string
+      name: props.string // String could be used also to define the prop type
```
treshugart: I'd actually prefer we advocate native usage instead of built-in, and introduce custom usage separately. We can do that in a separate PR.

treshugart:
This is awesome! Last thing to do is to pull latest and run `bolt` to ensure deps are latest. That should green the build.

blikblum:
I updated the repo, rebased and run yarn / bolt but the skatejs dependency on site package was not updated automatically. I had to update it manually.

With this the tests fail at MutationObserver tests. I checked and found that the errors were introduced by  https://github.com/skatejs/skatejs/commit/932bd95a9d141dce30b6324df573b91e32c10c74

treshugart:
Okay, I'll see what's going on there.

blikblum:
I rebased and now CI/Travis is passing

treshugart:
Nice work! Thanks!

treshugart:
Released in 5.2.0.

Issue #1448, opened 2018-06-12
Decorators like `props.string` do not works as expected. Attribute is not reflected to property.


majo44:
## Bug report

* Skate version: 5.1.3
* Affected browsers (and versions): any

### Current behaviour

Please look at: 
[![Edit 0qowkq1pk0](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/0qowkq1pk0)

When properties are described by usage of `static get props` everything is works fine.
But for the same scenario and almost same code where property is described as `@props.string name: string;` the value from attribute is not reflected as property.

### Expected behaviour

There should not be any differences between usage of decorators and static props property.

majo44:
Seems that:

1. [here](https://github.com/skatejs/skatejs/blob/master/packages/skatejs/src/with-update.js#L207) propertyDefinition is undefinied because 
2. property definition is only [reflected in _propsNormalised](https://github.com/skatejs/skatejs/blob/master/packages/skatejs/src/with-update.js#L78) and never in props constructor property

majo44:
Probably the simplest solution for this problem will be to [take the property Definition](https://github.com/skatejs/skatejs/blob/master/packages/skatejs/src/with-update.js#L207) from [_propsNormalised instead of props ](https://github.com/skatejs/skatejs/blob/master/packages/skatejs/src/with-update.js#L198)

P.S.
I tried to create PR but I used Windows and Ubuntu, on both I was not able install deps (npm which downloads 300MB/33K files (!sic) fails on post install) and the tests in master branch fails (some 6 tests fails).

treshugart:
Interesting. I wonder what is causing that to fail on your end. If you want to push a blind PR I'm happy to take it from there.

PR #1452, opened 2018-06-27
Fix for #1448

majo44:
Fix for #1448 
This is "blind" PR :)


majo44:
I have probably same problem as in #1450 
@treshugart are you able to handle it ?

treshugart:
Raised a separate PR (#1455) from master to pull these commits in. I'll make you a collaborator as it'll make collaborating on branches easier (and you've had several contributions in the past).

PR #1455, opened 2018-07-03
Fixes #1448

### LINKED ISSUE #1448: Decorators like `props.string` do not works as expected. Attribute is not reflected to property.

treshugart:
Fixes #1448. References #1452. cc @majo44 can you give this a quick thumbsup?

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -195,7 +195,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       const {
         _attributeToAttributeMap,
         _attributeToPropertyMap,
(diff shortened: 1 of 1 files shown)


treshugart:
Released in 5.2.1!

PR #1458, opened 2018-07-07
sk-marked: fix baseline indentation stripping

### LINKED ISSUE #1457: sk_marked strips first char of each line
## Bug report

* Skate version: _5.2.1_ (`npm info skatejs version`)
* Affected browsers (and versions): all
* Running on windows

### Current behavior

sk_marked strips first char of each line. 

### Expected behavior

Should not strip first char

I've got this error when building the site under a windows environment. Since windows handles new lines differently the bug is not visible on linux

Consider the following sk_market [code](https://github.com/skatejs/skatejs/blob/master/packages/sk-marked/src/index.js#L11):
```
 src = src.replace(/"/g, '&quot;');

  // Remove leading newlines.
  src = src.split('\n');

  // Get the initial indent so we can remove it from subsequent lines.
  const indent = src[1] ? src[1].match(/^\s*/)[0].length : 0;
```
at this point src is an array so `src[1]` matches the second line and not first char. Probably the bug is only apparent because on windows `src[1]` points to an empty line (when parsing skatejs README)

I did not tried a fix because i don't get the reason of this procedure. It's supposed to check the first text char and than if the char is blank, remove the some number of chars for each line. But what if next lines does not have the same indentation?

blikblum:
Fixes #1457 

* [X] Bug 
* [ ] Feature

## Requirements

* [x] Wrote tests.

## Rationale

Previous implementation was stripping more than necessary

## Implementation

Its straight forward: check indentation of each line to detect baseline indentation. If no indentation is found break loop and dont strip lines

### DIFF
--- a/packages/sk-marked/src/__tests__/utils.spec.js
+++ b/packages/sk-marked/src/__tests__/utils.spec.js
@@ -1,61 +1,59 @@
 import { format } from '../utils';
 
-describe('format', () => {
-  test('text without indentation', () => {
-    const text = `(line without indentantation)
+test('format text without indentation', () => {
+  const text = `(line without indentantation)
 (line without indentantation)
 
 (line without indentantation)
 
 `;
-    const formatted = format(text);
-    expect(formatted).toBe(text);
-  });
+  const formatted = format(text);
+  expect(formatted).toBe(text);
+});
 
-  test('text with mixed indentation but no baseline indentation', () => {
-    const text = `  (line indented with 2 spaces)
+test('format text with mixed indentation but no baseline indentation', () => {
+  const text = `  (line indented with 2 spaces)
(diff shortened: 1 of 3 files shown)


packages/sk-marked/src/__tests__/utils.spec.js
```diff
@@ -0,0 +1,61 @@
+import { format } from '../utils';
+
+describe('format', () => {
```
treshugart: Last time I tried a test() inside describe() it didn't work as expected. Could be wrong now, but you don't really even need the describe here. Long-story short, I've been moving everything to just use test() and separate things into files as much as possible, mostly matching the src structure.

treshugart:
Commented about describe() but other than that lgtm.

Issue #1459, opened 2018-07-09
withChildren: Children may not be upgraded when childrenUpdated() is called.


trusktr:
https://github.com/skatejs/skatejs/blob/31b776c6fd9031a2bb43e005d1ed8b93d0b1b001/packages/skatejs/src/with-children.js#L16-L24

I have some code similar to that one, calling a custom `childConnectedCallback` method inside of `connectedCallback`, but the problem with it is that `connectedCallback` can be called during parsing of the HTML page. If the custom element is already defined before parsing, then when `connectedCallback` is fired and therefore `fn()` is called, the `children` will be not-yet-upgraded if they are also custom elements. This will break code that relies on certain APIs existing on the child custom elements.

During parsing, parent elements are upgraded and their `connectedCallback` methods fired before their children are upgraded.

What I did to solve it in my case was to defer it like this:

```js
        connectedCallback() {
            super.connectedCallback && super.connectedCallback()

            // wait for children to have been upgraded
            // (in my case, children are from the same library: infamous)
            documentReady().then(() => {

                // if we're connected for the first time, we will not yet have a MutationObserver
                if (!Private(this).childObserver) {

                    const children = this.childNodes
                    for (let l=children.length, i=0; i<l; i+=1) {
                        this.childConnectedCallback(children[i])
                    }

                    // this creates a MutationObserver
                    Private(this).childObserver = observeChildren(this, this.childConnectedCallback, this.childDisconnectedCallback)
                }
            })
        }
```

In that sample `documentReady()` is a promise, so even if the document is already ready, the behavior is consistent (async).

I did this so that different orderings of loading `infamous` by script tag would always work. For example, compare the following two possibilities:

```html
<script src="infamous.js"></script>
<i-scene ...></i-scene>
```

```html
<i-scene ...></i-scene>
<script src="infamous.js"></script>
```

---

At some point, someone using `withChildren` will encounter this.

I wonder what's the best solution? Is waiting for document ready the best way to do it? What about on pages with lots of HTML to load, and the document takes long to load? Is there ever a chance to execute the callback earlier than document ready but after children are upgraded?

trusktr:
I made a PR to show the minimal idea: https://github.com/skatejs/skatejs/pull/1460

treshugart:
If children have a definition already loaded, during parsing, they should already be defined, IIRC. So in this instance, you could use `customElements.whenDefined()` to defer any work until after they're defined, or load the definitions before (or as dependencies of the host.

[...]

PR #1460, opened 2018-07-10
Fire childrenUpdated on document ready

trusktr:
* [x] Bug (could be, sometimes)
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

See https://github.com/skatejs/skatejs/issues/1459

## Implementation

Because it works best so far. I don't know if there's a better way.

## Open questions

Is there a better way? (see the linked issue for reasoning)

## Tasks

* [ ] add test

### DIFF
--- a/packages/skatejs/src/with-children.js
+++ b/packages/skatejs/src/with-children.js
@@ -19,7 +19,7 @@ export const withChildren = (Base: Class<any> = HTMLElement): Class<any> =>
         const fn = this.childrenUpdated.bind(this);
         this._mo = new MutationObserver(fn);
         this._mo.observe(this, { childList: true });
-        fn();
+        documentReady().then(fn);
       }
     }
 
@@ -30,3 +30,12 @@ export const withChildren = (Base: Class<any> = HTMLElement): Class<any> =>
       }
     }
   };
+
+function documentReady() {
+  if (document.readyState === 'loading') {
+    return new Promise(resolve => {
+      document.addEventListener('DOMContentLoaded', resolve);
+    })
+  }
+  return Promise.resolve();
+}


treshugart:
See my comment. I'm open to discussing this, but not sure if it's necessary to layer this into Skate yet.

blikblum:
I dont have opinions about the need of it or not (maybe create a proof of concept to show what is failing?) but implementation wise could be simplified: instead of returning a promise, pass the `fn` to `documentReady` and call it directly or attach it to 'DOMContentLoaded' event

trusktr:
If we determine we need it, I can simplify it. 👍 from me.

packages/skatejs/src/with-children.js
```diff
@@ -19,7 +19,7 @@ export const withChildren = (Base: Class<any> = HTMLElement): Class<any> =>
         const fn = this.childrenUpdated.bind(this);
         this._mo = new MutationObserver(fn);
         this._mo.observe(this, { childList: true });
-        fn();
+        documentReady().then(fn);
```
treshugart: Circling back to the discussion, I'm happy for this to go in. Looks like this could just be `document.addEventListener('DOMContentLoaded', fn)` though.

packages/skatejs/src/with-children.js
```diff
@@ -30,3 +30,12 @@ export const withChildren = (Base: Class<any> = HTMLElement): Class<any> =>
       }
     }
   };
+
+function documentReady() {
```
treshugart: Don't need this if doing `DOMContentLoaded`.

treshugart:
Let's go ahead with this. Commented re `DOMContentLoaded`. I'll port this to the `v6` branch.

treshugart:
Ported here: bbe91ed04cb63e7b209f7a94db996b4f7b464631.

packages/skatejs/src/with-children.js
treshugart: Cool, cheers. Might be worth porting that to the v6-trey branch, too.

trusktr:
@treshugart Hey hey, this still needs the conditional check

```js
  if (document.readyState === 'loading') {
    document.addEventListener('DOMContentLoaded', fn);
  } else {
    fn();
  }
```

otherwise in some cases the `childrenUpdated` won't be called with the initial children.

For example:

```js
// ... suppose right here the document is already loaded ("ready" in jQuery terms) ...

const el = document.createElement('my-element')
el.appendChild(document.createElement('other-element'))

// fires connected, but not childrenUpdated
document.body.appendChild(el)
```

because what will currently happen is that it will add an event listener for `DOMContentLoaded` but that event already passed, so it will never trigger. That's why I had that check in there.

treshugart:
Closing as this is in master and I don't have the capacity to maintain older versions right now. Thanks again for the PR, though! It helped flesh out the solution that went into skatejs/element :)

Issue #1453, opened 2018-07-02
Idea: a DOM node can contain contexts, and subnodes can get them by unique reference


trusktr:
I was looking at

https://github.com/skatejs/skatejs/blob/6d9fd2c1f6380ee1497e4286fe627660175e80d0/packages/skatejs/src/with-context.js#L10-L14

It looks like a sub-context can block one that is higher up. So if I stick a node in the middle of the tree with a context, it'll prevent a leaf node from getting a context from the top of the tree.

What does React do?

What if we change `set context` so that contexts inherit:

```js
    set context(context: Object): void {
      if (!this._context)
        this._context = Object.create(this.context);
          
      Object.assign(this._context, context); // naive copy, could be better
    }
```

This reminds me of `$scope` in Angular 1, which inherit from parent scopes.

The problem with this would be that, if we re-parent a node that has a context to another parent, the inheritance of the context would still be based on the old location which might not make sense. To solve this we could use connected/disconnected callbacks to change the prototype of the moved node's context.

treshugart:
I'd prefer to stick closer to what React would do here. They don't inherit, but the way they expose the provider / consumer API, it allows you to wrap them to inherit if you want:

```js
<Consumer>
  {value => (
    <Provider value={{...value, something: 'else' }}>{children}</Provider>
  )}
</Consumer>
```

I'm wondering if we can change our context implementation to behave more like this? One drawback of using the DOM like this is that updates to context don't trigger updates to subscribers. It'd be neat if we could do this.

blikblum:
I've been looking into how implement context for webcomponents and found the following approaches:

- skatejs, that is simple implementation but has drawback, as pointed by OP, of child contexts overriding completely parent ones
- [wc-context](https://github.com/blikblum/wc-context) by blikblum (me) that implements a sort of inheritance where child contexts only overrides the properties it exports. Currently the context properties are exported declaratively by a `contextRequests` hash, but i'm considering changing for a childContext / setChildContext approach
- [wc-context](https://github.com/askbeka/wc-context) by askbeka that uses a custom element to hold the context trying to mimic the new React API

Each one have pros and cons.

All in all, IMO there's not a design that is good enough, so in any case, keeping the current implementation or changing it, i think the context mixin should be opt-in, i.e., not included in withComponent

treshugart:
[...]

PR #1461, opened 2018-07-11
context implementation that doesn't rely on the DOM tree

trusktr:
This is an initial implementation of the idea in #1453. It's not tested yet, but paints the picture.

* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

See https://github.com/skatejs/skatejs/issues/1453

## Implementation

This way it doesn't rely on the DOM tree structure, and there's no overriding of contexts by other contexts (as there would be in the named-context and inheriting-context ideas also mentioned in #1453).

## Tasks

* [ ] docs
* [ ] tests

### DIFF
--- a/packages/skatejs/src/createContext.js
+++ b/packages/skatejs/src/createContext.js
@@ -1,0 +1,88 @@
+// @flow
+
+class Group extends Map {
+  set(...itemsInGroup) {
+  }
+}
+
+export function createContext(props: any): Context {
+
+  class Context {
+
+    constructor() {
+      this._callbacks = new Map
+      this._updatedProps = new Set
+    }
+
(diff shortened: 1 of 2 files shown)


trusktr:
This is interesting, because with this foundation someone could implement a more complex action/store system like Redux, or some sort of Rest API Model lib that auto-updates elements, etc.

trusktr:
No prop. For reference, here's the [full working example](https://codepen.io/trusktr/project/editor/XbEOMk) on Codepen.

I my self haven't needed the feature in any of my projects yet, but I imagine I will once I move to more app coding instead of library coding. Planning to make some games soon...

PR #1462, opened 2018-07-12
a DOM node can contain contexts, and subnodes can get them by unique reference

trusktr:
This is an initial implementation of the idea in #1453. It's not tested yet. Some parts might be incomplete, but it shows the idea.

* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

See https://github.com/skatejs/skatejs/issues/1453

## Implementation

This is similar to the original idea of contexts being relative to a parent element, but without them being overridden by intermediary contexts.

## Basic usage

Working demo: https://codepen.io/trusktr/project/editor/XbEOMk

```html
<el-with-context>
  <el-that-consumes>
  </el-that-consumes>
</el-with-context>
```

```js
// ./ElWithContext.js
import {withComponent} from 'skatejs'
import {SomeContext} from './SomeContext'

class ElWithContext extends withComponent() {
  static get provides() { return [ SomeContext ] }
  // ...
}

customElements.define('el-with-context', ElWithContext)
```

```js
// ./ElThatConsumes.js
import {withComponent} from 'skatejs'
import {SomeContext} from './SomeContext'

class ElThatConsumes extends withComponent() {
  static get consumes() { return [
    [ SomeContext, 'count' ] // subscribe to the `count` prop of `SomeContext`
  ] }

  connectedCallback() {
    super.connectedCallback()
    this.contexts.get(SomeContext).startCounter()
  }

  render() {
    const {count} = this.contexts.get(SomeContext)
    return <div>The count is: {count}</div>
  }
}

customElements.define('el-that-consumes', ElThatConsumes)
```

```js
// ./SomeContext.js
import {defineContext} from 'skatejs'

export const SomeContext = defineContext({
  count: 0

  startCounter() {
    setInterval(() => {
      someContext.count++
    }, 1000)
  },
})
```

## Tasks

* [ ] docs
* [ ] tests

### DIFF
--- a/packages/skatejs/src/createContext.js
+++ b/packages/skatejs/src/createContext.js
@@ -1,17 +1,19 @@
 // @flow
 
-class Group extends Map {
-  set(...itemsInGroup) {
-  }
-}
-
-export function createContext(initialProps: any): Context {
+export function defineContext(initialProps: any): Context {
+  initialProps = { ...initialProps }
 
   class Context {
 
     constructor() {
       this._callbacks = new Map
       this._updatedProps = new Set
+
+      this._propValues = {}
+
+      for (let prop in initialProps) {
+        this._propValues[prop] = initialProps[prop]
+      }
     }
 
     observe(props, callback) {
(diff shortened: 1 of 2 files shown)


trusktr:
Working demo: https://codepen.io/trusktr/project/editor/XbEOMk

(has a few minor changes different from the current PR, but gets the idea across. Would have to update the PR)

PR #1471, opened 2018-07-20
snabbdom: use a constant tag name instead of root.tagName

blikblum:

* [x] Bug
* [ ] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

root.tagName  is undefined for shadow root. This crashes snabbdom

## Implementation

Use a constant instead


blikblum:
> In general, I think we can simplify the code and let the user decide which root element to render.
In other words, we can avoid having Skate make a root element, and avoid do any checking like

The root element can be anything as returned by render method, and skatejs does not create any root element. 

The reason to create a root vnode is to prevent snabbdom to try to patch the shadow root or the element

trusktr:
> The root element can be anything as returned by render method, and skatejs does not create any root element.

SkateJS does create a root vnode inside of `renderer`, then it places the content returned from `render` (`call`) inside the root.

Snabbdom does not seem to allow to render an Array of child nodes using the `patch` function. The `patch` function accepts only root nodes.

Skate is allowing users to return an array from `render`, for example letting them write:

```
render() {
  return [
    <div>...</div>,
    <ul>...</ul>,
    <p>...</p>,
  ]
}
```

then it uses the check `Array.isArray(newVTree) ? newVTree : [newVTree]` to detect an array and make a new array.

(Why does it need to make a new array when it already got an array? cc @treshugart)

Maybe we can be simplify it to something like this: https://github.com/skatejs/skatejs/pull/1478

blikblum:
> SkateJS does create a root vnode inside of renderer, then it places the content returned from render (call) inside the root.

This root vnode maps to the element itself or shadow root. It does not create any element at all

> Snabbdom does not seem to allow to render an Array of child nodes using the patch function. The patch function accepts only root nodes.

It allows. Did you tested?

blikblum:
@treshugart can you have a look at this PR?

Currently snnabdom does not works with shadow dom

trusktr:
> It allows. Did you tested?

I guess if it does then no need for the array handling in Skate?

treshugart:
Released in 0.0.1.

Issue #1466, opened 2018-07-15
withChildren crashes with "Cannot set property props... " message


blikblum:
## Bug report

* Skate version: 5.2.1 (`npm info skatejs version`)
* Affected browsers (and versions): Chrome 67, Opera

### Current behaviour

Example created for https://github.com/skatejs/skatejs/issues/1186 (https://codesandbox.io/s/oqy3oozx9z) crashes with message "Cannot set property props of #<_class2> which has only a getter"

Really weird. Could not figure why

### Expected behavior

Should not crash

blikblum:
The crash reason is that preact renderer (and React/Snabbdom renderers) override `props` property getter but do not override `props` property setter. 

BTW: the children key set by withChildren is override by the children key returned by the renderers, which makes its useless

Anyway, more reasons to [not set children in props](https://github.com/skatejs/skatejs/issues/1443)

blikblum:
I confirm that this bug occurs also with snabbdom renderer and esnext builds (not only in codesandbox)

trusktr:
1. If you add your own setter, that seems to make it work: https://codesandbox.io/s/4qlzy1853x
2. Is there another way to fix it other than removing it? I'm not familiar with snabbdom though.
3. If we did remove that rom snabbdom, would we want to remove it from all the renderers?
4. Another question, do we want to let `this.props.children` be just the native children? Or is providing a `slot` element or vnode actually useful?
  - Note the user can also make their own slot element.
  - Note also that `this.children` also always returns the native children
5. `withChildren` does not seem to set the `children` in `static props`, so is that even reactive? Will it call `triggerUpdate` if that changes? (I haven't tested tried using `withChildren` yet.)

trusktr:
About point 6, this test shows that methods `childrenUpdated`, `updated`, or `render` do not get called when children are updated:

https://codesandbox.io/s/qq5z251px9

After the timeout, why does `childrenUpdated` not get called again?

trusktr:
Oops, I forgot to call `super.childrenUpdated()`, but the behavior is the same: https://codesandbox.io/s/2x1jlv7o50

trusktr:
Doh! I was forgetting to call `super.connectedCallback`. It's easy to overlook that. Sorry for the spam.

So this works: https://codesandbox.io/s/xl5q8vj78w

And if we call `updated()` inside of `childrenUpdated` then we get the desired updates: https://codesandbox.io/s/nrzpmx8lnl

[...]

PR #1472, opened 2018-07-20
snabbdom: do not override props to return children (a slot vnode)

blikblum:
#1466

* [x] Bug
* [ ] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

The props override does not works with withChildren mixin which is part of withComponent #1466 . 

Since any component built with snabbdom renderer is broken it cannot considered a breaking change. Also this renderer is still not documented

The consistency between renderers also is not broke because lit-html renderer already do not return children

### DIFF
--- a/packages/renderer-snabbdom/src/index.js
+++ b/packages/renderer-snabbdom/src/index.js
@@ -5,14 +5,6 @@ export default function createMixin(modules) {
   var patch = init(modules);
   return (Base = HTMLElement) =>
     class extends Base {
-      get props() {
-        // We override props so that we can satisfy most use
-        // cases for children by using a slot.
-        return {
-          ...super.props,
-          ...{ children: vnode('slot', {}, [], undefined, undefined) }
-        };
-      }
       renderer(root, call) {
         this._renderRoot = root;
         let newVTree = call();


trusktr:
Based on https://github.com/skatejs/skatejs/issues/1466#issuecomment-406945633, IMO let's

- get rid of `this.props.children` being a `<slot>`, as we can do that already in the markup just fine, and so then `this.children` and `this.props.children` will be consistent, as we're dealing with a DOM API here, and I think people using Skate will be aware of that.
- make child updates hook into `updated` or `render` lifecycle methods (similar to other frameworks like React or Preact).
- double check all definitions of `props` have both a `set` and `get`.
- update docs

Thoughts on that @blikblum ?

blikblum:
> get rid of this.props.children being a <slot>, as we can do that already in the markup just fine, and so then this.children and this.props.children will be consistent, as we're dealing with a DOM API here, and I think people using Skate will be aware of that.

Yes. For the other renderers it should be done at v6 since technically is a breaking change.
For snabbdom should be no problem because is undocumented and in current state is not working anyway

> make child updates hook into updated or render lifecycle methods (similar to other frameworks like React or Preact).

I dont understand how would work. Do you mean call updated or trigger render if children changes?

blikblum:
> Yeah, just a call to updated (which also calls render, etc, if using withRenderer),

This would cause at least double render: render changes children -> children changed triggers render. Depending what is done with children could loop infinitely 

> This makes the API more similar to existing patterns of other frameworks, f.e. in React and Preact changes to children are reactive and we can use this.props.children in lifecycle methods like componentDidUpdate and render.

children has a different meaning in React / Preact (passed vnodes) which can translate in different number of nodes.

trusktr:
> This would cause at least double render: render changes children -> children changed triggers render. Depending what is done with children could loop infinitely

Not the case when using a shadow root (standard in `withComponent`). `this.children` refers to children in the "light tree", not the children in the "shadow tree". The `render` method will render the content for the inner "shadow tree", and the outside children from the "light tree" are only modified by the end user of the element.

For example, if we take my example above,

```js
import {withComponent} from 'skatejs'

class AwesomeEl extends withComponent() {
    
    // this should work out of the box when children change:
    render({ children }) {
        return <div>
            <div>Number of children: {children.length}.</div>
            <div>The children: <slot></slot></div>
        </div>
    }
    
}

customElements.define('awesome-el', AwesomeEl)
```

then the end user would use it like this:

```html
<awesome-el>
  <h1>Hello</h1>
  <h2>World</h2>
</awesome-el>
```

which will render as

> Number of children: 2.</div>
> The children:
> # Hello
> ## World

Then, the user can remove the `h2` dynamically by doing

```js
const el = document.querySelector('awesome-el')
el.removeChild(el.querySelector('h2'))
```

and this will change the output the _user's_ DOM to

```html
<awesome-el>
  <h1>Hello</h1>
</awesome-el>
```

which will render as

> Number of children: 1.</div>
> The children:
> # Hello

trusktr:
The ShadowDOM root with `this.children` and `<slot>` is the same concept as `this.props.children` in React, and modifying the children of the element (not the children of the element's shadow root) is what is happening here, and `this.children` contains the element's children (not the element's root's children).

trusktr:
In any case, the `render` should render a slot, while it is more useful for the component if `this.children` is the list of DOM children so that component itself (regardless of which renderer it uses) can inspect those children.

blikblum:
Did you tested?

Please create a fiddle or codesandbox showing the example above

blikblum:
> What's the status of this? Is it ready for merging?

IMO, yes

treshugart:
This is now in the latest release compatible with `@skatejs/element`.

PR #1475, opened 2018-07-22
Do not destructure arguments in withRenderer.updated

blikblum:

* [ ] Bug
* [ ] Feature
* [x] Refactor

## Requirements

* [ ] Wrote tests.

## Rationale

`updated` method is always called with fixed number number of arguments. By avoiding the destructuration of arguments, we get a optimized code

## Implementation

Use named arguments instead of destructuration


trusktr:
👍 for optimization. The only downside is that if we update the `updated` signature in `with-update`, then we may have to remember to change it here too.

I'm good to merge though, because I don't think the `updated` signature is changing any time soon.

PR #1476, opened 2018-07-23
fix: inherit `observedAttributes`, if they exist in the base class.

trusktr:
* [x] Bug
* [ ] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

I ran into an issue when using `withUpdate` on my own class hierarchies: `withUpdate` was not inheriting `observedAttributes` from a base class if the base class had them. This caused some things not to work.

## Implementation

This inherits the `observedAttributes` so that other non-Skate behavior will still register with the Custom Element API and things will still work.

## Tasks

* [ ] write a test.

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -159,7 +159,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       // only once when the custom element is defined. If we did this only in
       // the constructor, then props would not link to attributes.
       defineProps(this);
-      return this._observedAttributes;
+      return this._observedAttributes.concat(super.observedAttributes || []);
     }
 
     static get props(): PropTypesNormalized {
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -159,7 +159,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       // only once when the custom element is defined. If we did this only in
       // the constructor, then props would not link to attributes.
       defineProps(this);
-      return this._observedAttributes.concat(super.observedAttributes || []);
+      return [...(super.observedAttributes || []), this._observedAttributes];
     }
 
     static get props(): PropTypesNormalized {
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -159,7 +159,9 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       // only once when the custom element is defined. If we did this only in
       // the constructor, then props would not link to attributes.
       defineProps(this);
-      return [...(super.observedAttributes || []), this._observedAttributes];
+      return Array.from(
+        new Set([...(super.observedAttributes || []), this._observedAttributes])
+      );
     }
 
     static get props(): PropTypesNormalized {


trusktr:
Any idea why travis-ci/push fails?

packages/skatejs/src/with-update.js
```diff
@@ -159,7 +159,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       // only once when the custom element is defined. If we did this only in
       // the constructor, then props would not link to attributes.
       defineProps(this);
-      return this._observedAttributes;
+      return this._observedAttributes.concat(super.observedAttributes || []);
```
trusktr: Another way to write it could be

```js
      return [...super.observedAttributes || [], this._observedAttributes];
```

trusktr:
I'll write tests by the weekend. No docs needed, this makes stuff work as expected when people define class hierarchies.

trusktr:
Updated, so it removes duplicates. Not sure if that's necessary though. Thoughts?

packages/skatejs/src/with-update.js
```diff
@@ -159,7 +159,9 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       // only once when the custom element is defined. If we did this only in
       // the constructor, then props would not link to attributes.
       defineProps(this);
-      return this._observedAttributes;
+      return Array.from(
```
treshugart: IE doesn't support `Array.from()` and doesn't support an iterable being passed to `Set`.

treshugart: You should just be able to write this as: `return this._observedAttributes.concat(super.observedAttributes)`, or similar.

treshugart:
> Currently is already possible to do something like:

True. I remember, now, that's why I never implemented the inheritance automatically. You can opt-in (using what @blikblum suggested) but you cannot opt-out if you inherit automatically.

treshugart:
I think it's best to close this for now given the original reasoning and the available workaround.

trusktr:
@blikblum You didn't understand my use case, let me try to describe it better:

What you both are talking about is components that extend from a _base_ class that has `withUpdate` mixed into it: I know I can easily do `super.observedAttributes.concat([ 'myprop' ])` in this case.

But my case is different: I want to use an existing class that _already has_ `observedAttributes` but it doesn't use SkateJS, then I want to extend from this new class using `withUpdate`.

This breaks and is unexpected (and very much a "bug").

This is what I mean:

```js
import ExistingClass from 'somewhere' // for example, from the ones listed on webcomponents.org

// ... ExistingClass already has an `observedAttributes` array ...

// Now, I want to extend a new class from ExistingClass, and use the benefits of SkateJS.

class NewClass extends withUpdate(ExistingClass) {
  get observedAttributes() {

    // this won't work.
    return super.observedAttributes.concat([ 'my-prop' ])

  }
}
```

It doesn't work because `super.observedAttributes` returns `this._observedAttributes` defined by SkateJS, which does not contain the `observedAttributes` from `ExistingClass`.

Or, is `class NewClass extends withUpdate(ExistingClass) {` not a supported use of SkateJS?

trusktr:
Here's a workaround for the issue:

```js
import ExistingClass from 'somewhere'

class NewClass extends withUpdate(ExistingClass) {
  get observedAttributes() {

    // this will work, but it is unexpected to have to do this:
    return ExistingClass.observedAttributes.concat([ 'my-prop' ])

  }
}

// Now SkateJS will properly create `_observedAttributes`:
customElements.define('my-new-element', NewClass)
```

trusktr:
Oh crap, wait, that breaks SkateJS, because we're not calling `super.observedAttributes`. So here's the final _working_ workaround:

```js
import ExistingClass from 'somewhere'

class NewClass extends withUpdate(ExistingClass) {
  get observedAttributes() {

    // this will work, but it is unexpected to have to do this:
    return super.observedAttributes.concat( ExistingClass.observedAttributes.concat([ 'my-prop' ]) )

  }
}

// Now SkateJS will properly create `_observedAttributes`:
customElements.define('my-new-element', NewClass)
```

trusktr:
`super.observedAttributes.concat( ExistingClass.observedAttributes.concat([ 'my-prop' ]) )` isn't intuitive. People will intuitively expect to be able to do `super.observedAttributes.concat([ 'my-prop' ])`.

trusktr:
Oops, forgot to put the `static` keywords in the samples. Updated ^

trusktr:
Made a new PR, as I've updated this one to show examples of what's exactly wrong: https://github.com/skatejs/skatejs/pull/1494

treshugart:
Looking back at your rationale I'm happy merging this.

PR #1478, opened 2018-07-23
Simplify Snabbdom renderer

trusktr:
* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

Makes Snabbdom simpler, based on discussion at https://github.com/skatejs/skatejs/pull/1471.

## Implementation

It's the way the Snabbdom README prescribes rendering to DOM.

`disconnectedCallback` is missing the part to remove it from the DOM (waiting for a reply at https://github.com/snabbdom/snabbdom/issues/386).

But maybe, we don't need to worry about removing the DOM. If the SkateJS custom element is itself removed from the DOM, I think everything will be garbage collected because all Snabbdom VNode information is stored on the nodes themselves.

## Open questions

Should we unmount the rendered Snabbdom output?

### DIFF
--- a/packages/renderer-snabbdom/src/index.js
+++ b/packages/renderer-snabbdom/src/index.js
@@ -5,45 +5,33 @@ export default function createMixin(modules) {
   var patch = init(modules);
   return (Base = HTMLElement) =>
     class extends Base {
-      get props() {
-        // We override props so that we can satisfy most use
-        // cases for children by using a slot.
-        return {
-          ...super.props,
-          ...{ children: vnode('slot', {}, [], undefined, undefined) }
-        };
-      }
+      _lastVTree = null;
+        
       renderer(root, call) {
         this._renderRoot = root;
         let newVTree = call();
-        newVTree = vnode(
-          root.tagName,
-          {},
-          Array.isArray(newVTree) ? newVTree : [newVTree],
-          undefined,
-          root
-        );
-        if (!this._vTree) {
-          // small cheat to allow rendering root el
-          // creates an empty vnode with the same sel as the rendered vtree
-          // this ensure the view element will be properly patched
-          const emptyVTree = vnode(root.tagName, {}, [], undefined, root);
-          patch(emptyVTree, newVTree);
-        } else {
-          patch(this._vTree, newVTree);
+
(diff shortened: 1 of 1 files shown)


trusktr:
> I dont want to be rude, but seems you are not familiarized with snabbdom.

It's all good! Their docs aren't very good (yet), and I prefer not to learn strictly from reading source code. If a lib is not documented well, then it is a no go for me.

> > Should we unmount the rendered Snabbdom output?
> Yes of course. Otherwise lifecycle (destroy/remove) events wont be fired

I had to ask: https://github.com/snabbdom/snabbdom/issues/386

Of course that answer is simple, but it was not obvious. I updated the PR to unmount the snab from the dom using `null`.

> Your implementation just creates a temp element that is internally converted to a vnode, just creating more overhead.

So how can we do it? Want to add that to the PR?

trusktr:
Even their most basic example in the README says to do it that way like I did:

```js
var container = document.getElementById('container');

var vnode = h('div#container.two.classes', {on: {click: someFn}}, [
  h('span', {style: {fontWeight: 'bold'}}, 'This is bold'),
  ' and this is just normal text',
  h('a', {props: {href: '/foo'}}, 'I\'ll take you places!')
]);

// Patch into empty DOM element – this modifies the DOM as a side effect
patch(container, vnode);
```

Notice they pass `container` to `patch`. That's the same thing I did with `tmp`.

trusktr:
Okay, look at the diff now. That's short and easy to understand, and no _hacks_.

If you want to hack, maybe this will work instead:

```js
        if (!this._lastVTree) {
          !root.tagName && (root.tagName = 'ROOT')
          this._lastVTree = patch(root, newVTree);
        } else {
```

but then the `root` does not have a `parentNode`, so I'm not sure if that'll break anything, because see:

https://github.com/snabbdom/snabbdom/blob/b7f90368ee771a7e57def6f261f3d267b5f3f210/src/snabbdom.ts#L299-L306

treshugart:
It seems like this might be similar to #1471 which was just merged?

treshugart:
These changes should have made their way into the latest renderer release.

PR #1488, opened 2018-07-30
allow a getter for the default value

trusktr:
* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

Makes it easy to do define a getter for the default value, for example to generate random defaults like in this custom prop definition:

```js
import { Color } from 'three'
import { props } from 'skatejs'

Object.assign(props, {
    THREE: {
        Color: {
            attribute: { source: true }, // get the value from an attribute (but don't mirror it back)
            coerce: val => new Color( val ), // Color accepts integers or {r: _, g: _, b _} objects
 /*HERE->*/ get default() { return new Color( Math.random(), Math.random(), Math.random() ) }, // random default color
            deserialize: val => new Color( val ), // Color accepts CSS strings
            serialize: val => new Color( val ).getStyle(), // returns CSS string like "rbg(220, 130, 50)"
        },
    },
}

export { props }
```

Now every time I make a new instance, it has a random new default color if I didn't specify one via attribute. 👍 😄 

## Implementation

Currently, the default getter is read only once, during prop normalization. This allows it to always work after normalization, which is super useful for default value that should be dynamically generated.

## Open questions

## Tasks

* [ ] tests
* [ ] docs

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -37,14 +37,16 @@ export function normalizePropertyDefinition(
   name: string,
   prop: PropType
 ): Object {
-  const { coerce, default: def, deserialize, serialize } = prop;
-  return {
+  const { coerce, deserialize, serialize } = prop;
+  const defaultDescriptor = Object.getOwnPropertyDescriptor(prop, 'default');
+  const result = {
     attribute: normalizeAttributeDefinition(name, prop),
     coerce: coerce || identity,
-    default: def,
     deserialize: deserialize || identity,
     serialize: serialize || identity
   };
+  Object.defineProperty(result, 'default', defaultDescriptor);
+  return result;
 }
 
 const defaultTypesMap = new Map();


trusktr:
I'll write tests and docs by the weekend.

blikblum:
In my opinion, the better way to achieve the desired result is using the primitive for default values in ES classes:

```
class X extends Component {
   static props = {x: Object}
   x = new Color( Math.random(), Math.random(), Math.random() )
}
```

treshugart:
I agree with @blikblum here. This is how you’d do it in React and other, similar libs.

treshugart:
Closing since it looks like we can do this already.

trusktr:
Setting the default on the class is interesting.

However, there might be a bunch of classes in non-related hierarchies that want to use a prop with the same default value, in which case it becomes no longer DRY if we have to repeat it in all the classes.

Having the default in a shared prop type makes it more DRY.

Looking at how to propose this in `v6`.

treshugart:
I can see this being useful for some more complex properties. This has been tentatively added in v6-trey as a function that is called if `get()` returns `== null`.

PR #1490, opened 2018-08-05
Use lit-html over the deprecated lit-html-extended

MichaReiser:
* [ ] Bug
* [x] Feature?

## Requirements

* [x] Read the [contribution guidelines]
* [ ] Wrote tests.
* [x] Updated docs 
* [ ] upgrade instructions: Unclear where I can find these

## Rationale

[lit-extended is deprecated](https://github.com/Polymer/lit-html/blob/master/src/lib/lit-extended.ts#L57) and does no longer support all the features it did before (e.g., on-click binding).

## Implementation

Replaced the render call from lit-extended with the call to the default render function. 

## Open questions

- This is a breaking change. The semver of the renderer-lit-HTML version needs to be bumped to a new major version. Should we change the version to a beta version as long as the lit-HTML dependency points to a dev version?

## Other

This is a breaking change. All lit-HTML component needs to be updated to use the new syntax.

## Tasks

* [ ] Upgrade guide
* [ ] Bump version

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "highlight.js": "9.12.0",
     "jest-environment-node": "22.4.1",
     "js-beautify": "1.7.5",
-    "lit-html": "0.9.0",
+    "lit-html": "^0.11.0-dev.1",
     "marked": "0.3.14",
     "mkdirp": "0.5.1",
     "outdent": "^0.5.0",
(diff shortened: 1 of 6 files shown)


treshugart:
> Should we change the version to a beta version as long as the lit-HTML dependency points to a dev version?

I'm okay with just bumping. Lit currently is in dev releases anyways.

PR #1492, opened 2018-08-05
Fix extending builtin classes

### LINKED ISSUE #1465: `es` build doesn't work with native Custom Elements, can't call HTMLElement without `new`
## Bug report

* Skate version: 5.2.1 (`npm info skatejs version`)
* Affected browsers (and versions): ones that have native Custom Elements

### Current behaviour

demo: https://codesandbox.io/s/vyy3pklr15

### Expected behaviour

For the transpiled code to be able to call HTMLElement classes properly.

### Solution:

Build the `es` version using @WebReflection's [`babel-plugin-transform-builtin-classes`](https://github.com/WebReflection/babel-plugin-transform-builtin-classes/). It solves the problem at build-time.

I suggest to avoid Polymer's native-shim.js, which works at runtime and introduces build complexity when you want to produce only a single bundle that works everywhere.

### Workarounds

- Use [`lowclass`](https://github.com/trusktr/lowclass/) or [`newless`](https://github.com/Mr0grog/newless/) to define classes that extend builtins. This will always work regardless of the build system.
- If you're using a tool like Webpack and a modern browser, tell it to use the `esnext` version by specifying an alias.

### TODO

- [ ] A pull request

MichaReiser:
Use the babel-plugin-transform-builtin-classes in the es build to properly call the super constructor when extending HTMLElement.

Fixes #1465 

* [x] Bug
* [ ] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

See #1465

## Implementation

Uses the [babel-plugin-transform-builtin-classes](https://github.com/WebReflection/babel-plugin-transform-builtin-classes/) to transform extends of built-in classes. The plugin only performs a static analysis to match base class names that might require the fix. Since skatejs use dynamic base classes it was needed to match the identifier `Base` that can potentially refer to a built-in class. 

I preferred the compilation step over using a library for inheritance since this is only needed when older browsers without class support are targeted. 

## Open questions

## Tasks

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,12 @@
           "flow",
           "react",
           "stage-0"
+        ],
+        "plugins": [
+          ["babel-plugin-transform-builtin-classes", {
+            "globals": ["HTMLElement", "Base"],
+            "logIfPatched": true
+          }]
         ]
       },
       "esnext": {
@@ -104,6 +110,7 @@
     "babel-plugin-dynamic-import-node": "1.2.0",
     "babel-plugin-modules-map": "1.0.0",
     "babel-plugin-modules-web-compat": "1.1.1",
+    "babel-plugin-transform-builtin-classes": "^0.6.1",
     "babel-plugin-transform-decorators-legacy": "1.3.4",
     "babel-plugin-transform-skate-flow-props": "0.1.0",
     "babel-preset-env": "1.6.1",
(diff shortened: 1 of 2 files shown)


treshugart:
Awesome! Thanks.

trusktr:
@MichaReiser Good call! Note, this is supposed to be fixed in Babel 7 (still Beta), but last I checked it wasn't working. Gotta check the new `rc` releases.

trusktr:
Here's the Babel 7 thread: https://github.com/babel/babel/pull/7020

PR #1506, opened 2018-08-18
simplify the props instance getter

trusktr:
* [ ] Bug
* [ ] Feature
* [x] refactor

## Requirements

* [ ] Wrote tests.

## Rationale

Seems like that does the same thing.

## Implementation

_Why have you implemented it this way?_

To make it simpler

## Open questions

Do we need to use `keys()` for a specific reason?

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -171,13 +171,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
     }
 
     get props(): Object {
-      return keys(this.constructor.props).reduce(
-        (prev: Object, curr: string) => {
-          prev[curr] = (this: any)[curr];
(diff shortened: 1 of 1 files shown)


treshugart:
Looks like the tests are red. Happy to land this in v5. In v6 I'm probably removing `props` (as a getter / setter) altogether in favour of just using `elem.yourProp` instead of `elem.props.yourProp`. I think it simplifies things a lot.

trusktr:
Yeah, I think just have `this.foo` and not `this.props.foo` is ideal, we don't need to be able to get the same prop from two places. The only potential issues are with props that clash with builtin ones, like `this.children`. In my case, I've used `this.subnodes` to avoid clashing.

trusktr:
I'll just leave this as is, because it works, and we can focus on `v6`.

Issue #1487, opened 2018-07-30
commit hook commits all unstaged changes


trusktr:
I noticed that a commit hook will pick up any un-staged changes and add them to the commit I just made, which isn't normal expected git behavior. 

Example:

```
❯ git status
On branch default-value-getter
Your branch is behind 'origin/default-value-getter' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        modified:   packages/skatejs/src/with-update.js

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   packages/skatejs/src/with-update.js

❯ git commit -m "foo"
husky > npm run -s precommit (node v9.11.2)

 ✔ Running tasks for *.{js,json,md}
[default-value-getter 30c4dadc] foo
 1 file changed, 6 insertions(+), 4 deletions(-)

❯ git status
On branch default-value-getter
Your branch and 'origin/default-value-getter' have diverged,
and have 1 and 1 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)

nothing to commit, working tree clean
```

It'd be great if we could fix this, because it complicates command line usage, requires more commands to save unstaged changes elsewhere.

trusktr:
If I have no unstaged changes, the output of the commit command is different:

```
skatejs+skatejs git:default-value-getter ❯ git commit -m "foo"
husky > npm run -s precommit (node v9.11.2)

 ↓ Running tasks for *.{js,json,md} [skipped]
   → No staged files match *.{js,json,md}
On branch default-value-getter
Your branch is behind 'origin/default-value-getter' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

nothing to commit, working tree clean
```

treshugart:
It ensures things meet code formatting standards. We could possibly change it so that it creates a separate commit, but I'm not sure if we can do that.

trusktr:
> meet code formatting standards

That part is nice. It's just unexpected that when I try to commit staged changes it also commits unstaged changes.

Can we stash the unstaged changes, run the code formatting, then after all is done, pop the last stash?

trusktr:
For now I can use `git stash -k` before every commit, so that it stashed only unstaged changes.

treshugart:
[...]

PR #1507, opened 2018-08-18
the default property of a prop definition is a function, so we can do…

trusktr:
The default property of a prop definition can be a function, so we can do useful things like call `Math.random` or use `this` to do instance-specific things.

* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

This makes it easy to make dynamic default values. For example,

```js
const THREEColorProp = {
    attribute: { source: true, target: false }, // get the value from an attribute (but don't mirror it back)
    coerce: val => val instanceof Color ? val : new Color(val),
    deserialize: val => new Color(val), // f.e. accepts a hex "#af3eea" color string
    serialize: val => new Color( val ).getStyle(), // returns a CSS "rbg()" string

    // Let the default Color be random for each instance that has this prop
    default: () => new Color( Math.random(), Math.random(), Math.random() ), // R, G, B
}
```

## Implementation

This way the value can be a default static value, or a function that returns a value. If a function, it is called every time a default value is needed, and also called with `this` which can be useful in some cases where props are designed to work with certain classes that handle their own default values. 

Basically, it gives the end user more flexibility.

The only caveat (which we'd put in the docs) is that if the value is meant to be a function, then the function should be returned from the function:

```js
const someFunction = () => { /* ... */ }

class Foo extends withUpdate(HTMLElement) {
    static props = {
        foo: {
            // ...
            default: () => someFunction
        }
    }
}
```

## Tasks

* [ ] add to docs
* [ ] add a test

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -41,7 +41,7 @@ export function normalizePropertyDefinition(
   return {
     attribute: normalizeAttributeDefinition(name, prop),
     coerce: coerce || identity,
-    default: def,
+    default: typeof def === 'function' ? def : () => def,
     deserialize: deserialize || identity,
     serialize: serialize || identity
   };
@@ -97,7 +97,7 @@ export function prop(definition: PropType | void): Function {
       configurable: true,
       get() {
         const val = this._props[name];
-        return val == null ? normalized.default : val;
+        return val == null ? normalized.default.call(this) : val;
       },
       set(val) {
         const { attribute: { target }, serialize } = normalized;
@@ -216,7 +216,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
           const { default: defaultValue, deserialize } = propertyDefinition;
           const propertyValue = deserialize ? deserialize(newValue) : newValue;
           this._props[propertyName] =
-            propertyValue == null ? defaultValue : propertyValue;
+            propertyValue == null ? defaultValue.call(this) : propertyValue;
           this.triggerUpdate();
         }
       }


trusktr:
Oops, the commit hook committed my other unstaged changes I was playing with (https://github.com/skatejs/skatejs/issues/1487).

For now just pay attention to the stuff regarding `default`, and I can remove the other stuff.

treshugart:
If the prop default is a function, then this would be a breaking change no?

Looking down the road at v6, default props are specified as class properties or in the constructor, so you could just do this in the constructor. You can probably just do that now in v5. I know it's nice for certain props but I have a few problems with it:

- I really want prop types to do no dynamic transforms on the value.
- The only ones we really *must* do are for property serialization and deserialization; there's nothing we can do about that.
- I'm not sure how you'd type the return values and instance properties. It feels like, even if it was possible, it'd be a nasty mess of generics.

I've found, that most of the time, for default props on a pre-defined property, you're looking for a value that just "doesn't error" (tm) when being passed through the functions. I'd prefer it if types on the custom element instance were leveraged here instead and prop types were built to be resilient where values are handled internally (or typed).

trusktr:
> If the prop default is a function, then this would be a breaking change no?

In this case no, because the normalization process converts a static value into a function that return the static value, so defining `default: 123` still works.

> Looking down the road at v6, default props are specified as class properties or in the constructor, so you could just do this in the constructor. 

Ah, interesting, I didn't realize this yet... Let me check this out...

trusktr:
> I'd prefer it if types on the custom element instance were leveraged here instead and prop types were built to be resilient where values are handled internally (or typed).

That sounds good; props for typing, and classes defining their defaults.

trusktr:
In my cases, the default values aren't for "not erroring", but more for nice default visuals. For example, rendering WebGL objects with default colors, default sizes, default fog value, etc, so that out of the box things look nice.

treshugart:
This ended up making its way to `@skatejs/element` so closing.

Issue #1493, opened 2018-08-06
[idea] Allowing the use of `this` in prop definitions, as well as passing in the prop name.


trusktr:
I mentioned it in https://github.com/skatejs/skatejs/issues/1385, but this is a specific issue to track the idea (I will circle back to my old ones and close them or leave them open with intent for a PR).

## Feature request

After experimenting with the `props` API and defining a bunch of prop types lately, I've come to wish for and desire being able to use `this` inside the prop definition functions.

trusktr:
### Use case(s)

In my example at https://github.com/skatejs/skatejs/issues/1481#issuecomment-410622063, I'd like to change my prop-defining function from

```js
function createPropType(Type, override = {}) {
    return {
        attribute: { source: true, target: false }, // get the value from an attribute (but don't mirror it back)
        coerce: val => val instanceof Type ? val : new Type(val),
        default: new Type,
        deserialize: val => new Type(val),
        serialize: val => val.toString(),
        ...override,
    }
}
```

to

```js
function createPropType(Type, override = {}) {
    return {
        attribute: { source: true, target: false },
        coerce(prop, val) { return val instanceof Type ? val : this[prop].from(val) }, // knowing that `this[prop]` is already an instance of Type
        default: new Type,
        deserialize(prop, val) { return this[prop].from(val) },
        serialize(prop, val) { return this[prop].toString() },
        ...override,
    }
}
```

Basically, overriding how values are cached.

I'm not sure if this is the best solution, but basically the problem I have with the former is that it can only ever create a new instance, because there's nowhere to cache it. But if I have access to `this`, I can make a custom way of caching values.

The reason I'd like to keep instances the same (besides for performance) is because other parts of the app observe the thing that is stored in `this[prop]` indefinitely (direct notifications), and I would like to avoid having to modify that part of the code (introducing diffing).

This would be a none-breaking change, as users can still return values just like before.

trusktr:
Would it make sense to add `get` and `set` options to `props`, which do have access to `this`? This circles back to https://github.com/skatejs/skatejs/issues/1385.

Then we could leave `coerce`/`serialize`/`deserialize` as is, responsible for their job, and let `set` and `get` determine where to set/get values to/from.

[...]

PR #1508, opened 2018-08-19
Access `this` and property names in prop functions

trusktr:
This expands on PR https://github.com/skatejs/skatejs/pull/1507. Related issue: https://github.com/skatejs/skatejs/issues/1493

* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

This makes the API more flexible for certain use cases while maintaining backwards compatibility (except for the case where a `default` value should be a function, see the other pull request).

This is awesome because it let's me greatly simplify my classes. It let's me handle values in special way so I can avoid creating new objects and gain performance improvements as well as code structuring improvements so that I can work with existing functionality rather than having to re-write classes. 

Existing functionality in my classes keeps the same objects around for each property, and this new `withUpdate` functionality allows me to keep this. Otherwise I'd have to re-write my classes so that they work with more diffing and always new object references every time a prop is set (especially not desirable for games/graphics/animations, which is what I'm making).

I'm using it like follows (note I'm using it in both the current way and the new way, to show it is backwards compatible):

```js
import { props as skateProps } from 'skatejs'
import { Color } from 'three'
import XYZValues from '../../core/XYZValues'
import XYZNumberValues from '../../core/XYZNumberValues'
import XYZNonNegativeValues from '../../core/XYZNonNegativeValues'
import XYZStringValues from '../../core/XYZStringValues'
import XYZSizeModeValues from '../../core/XYZSizeModeValues'

// creates a prop definition using the existing `withUpdate` functionality.
function createGenericPropType(Type, override = {}) {
    return {
        attribute: { source: true, target: false }, // get the value from an attribute (but don't mirror it back)
        coerce: val => val instanceof Type ? val : new Type(val),
        default: new Type,
        deserialize: val => new Type(val),
        serialize: val => val.toString(),
        ...override,
    }
}

// creates a prop definition for use with certain of my classes, taking advantage of these new features
// so that we can cache/serialize/deserialize to/from existing objects without creating new ones.
// This open up other possibilities too...
function createXYZPropType(Type, override = {}) {
    return {
        attribute: { source: true, target: false }, // get the value from an attribute (but don't mirror it back)
        coerce(val, propName) { return val === this.element[propName] ? val : this.element[propName].from(val) },
        default(propName) { return this.element[propName] },
        deserialize(val, propName) { return this.element[propName].fromString(val) },
        serialize(val, propName) { this.element[propName].toString() },
        ...override,
    }
}

const props = Object.assign({}, skateProps, {
    THREE: {
        Color: createGenericPropType(Color, {
            default: () => new Color( Math.random(), Math.random(), Math.random() ),
            serialize: val => new Color( val ).getStyle(), // returns a CSS "rbg()" string
        }),
    },
    XYZValues: createXYZPropType(XYZValues),
    XYZNumberValues: createXYZPropType(XYZNumberValues),
    XYZNonNegativeValues: createXYZPropType(XYZNonNegativeValues),
    XYZStringValues: createXYZPropType(XYZStringValues),
    XYZSizeModeValues: createXYZPropType(XYZSizeModeValues),
})

export { props }
```

## Implementation

This was the simplest way to implement it to allow extra flexibility with backwards compatibility (except for `function` default values, see https://github.com/skatejs/skatejs/pull/1507).

## Tasks

* [ ] docs
* [ ] test

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -97,7 +97,9 @@ export function prop(definition: PropType | void): Function {
       configurable: true,
       get() {
         const val = this._props[name];
-        return val == null ? normalized.default.call(this, name) : val;
+        return val == null
+          ? (this._props[name] = normalized.default.call(this, name))
+          : val;
       },
       set(val) {
         const { attribute: { target }, serialize, coerce } = normalized;
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -97,19 +97,21 @@ export function prop(definition: PropType | void): Function {
       configurable: true,
       get() {
         const val = this._props[name];
-        return val == null ? normalized.default.call(this) : val;
+        return val == null ? normalized.default.call(this, name) : val;
       },
       set(val) {
-        const { attribute: { target }, serialize } = normalized;
+        const { attribute: { target }, serialize, coerce } = normalized;
         if (target) {
-          const serializedVal = serialize ? serialize(val) : val;
+          const serializedVal = serialize
+            ? serialize.call(this, val, name)
+            : val;
           if (serializedVal == null) {
             this.removeAttribute(target);
           } else {
             this.setAttribute(target, serializedVal);
           }
         }
-        this._props[name] = normalized.coerce(val);
+        this._props[name] = coerce.call(this, val, name);
         this.triggerUpdate();
       }
     });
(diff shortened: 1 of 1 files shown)


trusktr:
Now that `default` can be a function, I needed to push commit 24e50a9e61ed9f0a0000c70245efb626e2f7e3ca to ensure that it is called only once initially, so that successive gets don't re-generate a new default (f.e. when using a random default value).

F.e., if you have

```
  default: () => Math.random()
```

the expectation is that the default initial value is random.

treshugart:
This should now be possible in `@skatejs/element`.

Issue #1489, opened 2018-07-31
idea: Knowing which prop has changed without diffing.


trusktr:
## Feature request

It'd be `suuuuuper` sweet to be able to know which props have changed inside of `updated` without having to perform a diff.

For example, currently, one has to do this:

```js
updated(oldProps) {
  if (oldProps.foo !== this.foo) {
    // the foo prop changed
  }
}
```

The suggested API might look like this:

```js
updated(oldProps, oldState, changedProps) { // backwards-compatible signature

  // changedProps is a Set, so `.has()` is O(<n), faster than array `indexOf`.

  if (changedProps.has('foo')) {
    // the foo prop changed
  }
}
```

For values that different in equality, I could easily implement this in the `updating` method, then let `shouldUpdate` and `updated` use it.

_Buuuuut_, as usual, I have a case that doesn't fit in with the regular cases.....

### Use case(s)

It turns out, in my case, my class hierarchy has a getter/setter for the prop that is being observed. My `withUpdate` component observes changes to this value. When the value is set, it triggers an update.

However, when the value is get, it _always_ returns the same object reference.

For example:

```
element.size = [100, 100, 100] // triggers observation that my withUpdate component listens to
```

```js
element.size // returns the same object reference, f.e. { x: 100, y: 100, z: 100 }
```

which is an object who's reference never changes (for now at least).

So in my `withUpdate` component, my diff logic fails because:

```js
updated(oldProps) {
  if (oldProps.size !== this.size) {
    // we never reach this part
  }
}
```

It's not a common use case, but, I've ran into it.

If I can avoid re-writing lots of stuff in my library to accommodate this, that'd be awesome.

I think the implementation inside of `with-update.js` would actually be fairly small.

trusktr:
I forgot to note, for now I just don't perform any check, and just run the logic all the time. This won't be ideal in large applications because it involves re-creating new WebGL geometries even though nothing has changed.

I can probably cache the value further down in the logic and diff like that (f.e. observe `this.size.x` instead of just `this.size.y`, but I think it'd be nice to simply have a feature abstracted higher up that tells me explicitly which value changed with no diffing involved.

trusktr:
Because of performance. That's very _readable_, but it is _heavy_ and it is extra code (abstracted away behind lodash, take a look at all the stuff lodash is doing under the hood).

Direct notification is the fastest (this is why [Surplus is so fast](https://www.stefankrause.net/js-frameworks-benchmark7/table.html)).

[...]

PR #1510, opened 2018-08-19
Tell consumers which props were modified

trusktr:
Builds off of PR https://github.com/skatejs/skatejs/pull/1508. Related issue: https://github.com/skatejs/skatejs/issues/1489

* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

SkateJS already knows which props have changed, so we can drastically clean up end user code by simply exposing this information.

This drastically cleaned up some of my code. For example, I had code that I changed from

```js
    constructor(element) {
        Super(this).constructor(element)

        Private(this).lastSize = [null, null, null]
        Private(this).lastSizeMode = [null, null, null]
    },

    updated( oldProps, oldState ) {
        if (Private(this).sizeChanged(this.size, this.sizeMode)) {
            this._updateGeometryOnSizeChange(this.size)
        }
    },

    private: {
        lastSize: [null, null, null],
        lastSizeMode: [null, null, null],

        sizeChanged(newSize, newSizeMode) {
            let result = false

            if (
                this.lastSize[0] !== newSize.x ||
                this.lastSize[1] !== newSize.y ||
                this.lastSize[2] !== newSize.z
            ) {
                result = true
                this.lastSize[0] = newSize.x
                this.lastSize[1] = newSize.y
                this.lastSize[2] = newSize.z
            }

            if (
                this.lastSizeMode[0] !== newSizeMode.x ||
                this.lastSizeMode[1] !== newSizeMode.y ||
                this.lastSizeMode[2] !== newSizeMode.z
            ) {
                result = true
                this.lastSizeMode[0] = newSizeMode.x
                this.lastSizeMode[1] = newSizeMode.y
                this.lastSizeMode[2] = newSizeMode.z
            }

            return result
        },
    },
```

to

```js
    updated( oldProps, oldState, modifiedProps ) {
        if ( modifiedProps.has('size') || modifiedProps.has('sizeMode') ) {
            this._updateGeometryOnSizeChange(this.size)
        }
    },
```

This is highly valuable!

## Implementation

Implemented this way to be backwards compatible. Existing code continues to work as is.

## Tasks

* [ ] test
* [ ] docs

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -52,11 +52,13 @@ const defaultTypesMap = new Map();
 function defineProps(constructor) {
   if (constructor.hasOwnProperty('_propsNormalized')) return;
   const { props } = constructor;
+  constructor._propsList = [];
   keys(props).forEach(name => {
     let func = props[name] || props.any;
     if (defaultTypesMap.has(func)) func = defaultTypesMap.get(func);
     if (typeof func !== 'function') func = prop((func: any));
     func({ constructor }, name);
+    constructor._propsList.push(name);
   });
 }
 
@@ -110,6 +112,7 @@ export function prop(definition: PropType | void): Function {
           }
         }
         this._props[name] = normalized.coerce(val);
+        this._modifiedProps[name] = true;
         this.triggerUpdate();
       }
     });
@@ -134,6 +137,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
     _prevState: Object;
     _props: Object;
     _state: Object;
+    _modifiedProps: Object;
     _syncingAttributeToProperty: null | string;
     _syncingPropertyToAttribute: boolean;
     _updating: boolean;
@@ -153,6 +157,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
     _prevState = {};
     _props = {};
     _state = {};
+    _modifiedProps = {};
 
     static get observedAttributes(): Array<string> {
       // We have to define props here because observedAttributes are retrieved
@@ -217,6 +222,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
           const propertyValue = deserialize ? deserialize(newValue) : newValue;
           this._props[propertyName] =
             propertyValue == null ? defaultValue : propertyValue;
+          this._modifiedProps[propertyName] = true;
           this.triggerUpdate();
         }
       }
@@ -235,6 +241,9 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       if (super.connectedCallback) {
         super.connectedCallback();
       }
+      const propsList = this.constructor._propsList;
+      for (let i = 0, l = propsList.length; i < l; i += 1)
+        this._modifiedProps[propsList[i]] = true;
       this.triggerUpdate();
     }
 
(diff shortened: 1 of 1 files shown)


matthewp:
Yeah, I agree with @blikblum here. Is this not possible to build as a separate mixin?

packages/skatejs/src/with-update.js
```diff
@@ -138,6 +139,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
     _prevState: Object;
     _props: Object;
     _state: Object;
+    _modifiedProps: Set;
```
treshugart: Why not just use an object?

packages/skatejs/src/with-update.js
```diff
@@ -241,6 +245,9 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
       if (super.connectedCallback) {
         super.connectedCallback();
       }
+      keys(this.constructor.props).forEach(prop => {
```
treshugart: Why do all props need to be signalled as changing when connected?

trusktr: This would be effectively the same as firing a default `attributeChangedCallback` on connecting, so my thought here was that we want to react to all the values. Usually we need the values for something, so we need to use them all initially. In my cases for example, I need to set WebGL rendering properties/features based on the initial values.

There isn't a way to know if `updated` is being called for the firs time, and the `updated` method will generally always contain conditional checks to determine which props changed, so in order to fire logic for all props, this is why I thought to initially set all props modified.

If we don't set them all as "modified" initially, then the only option we have is to avoid the use of `modifiedProps` in the `updated` method and instead diff `oldProps` vs `this.props`, which defeats the purpose of having `modifiedProps`.

Is there maybe a better way to run initial logic for all props?

trusktr:
> Is this not possible to build as a separate mixin?

@matthewp  It doesn't seem easy to do this without duplicating much of the efforts already put into `withUpdate`.

> This will incurs in performance penalty (plus increased code size and memory consumption) for majority of cases / users that don't need this feature.

@blikblum Any thoughts on how we might improve it, maybe a way for classes to opt-in?

For big applications with lots of data/state, I'm fairly certain the cost of maintaining a `Set<string>` will be more efficient than repeatedly creating/destroying/diffing objects.

> Also breaks consistency between state/props (i would expect state have the same feature)

State's already completely different: it's basically a single prop whose value is always an object. Personally I never use it, props have all the reactivity I need. This is similar to Vue components (they don't have a `state` object, just props).

We could remove state, and effectively have the same thing by defining it with props:

```js
static props = {
  state: props.object,
}
```

@treshugart 

> Why not just use an object?

Hmmmm... do you mean, like have the keys on an object so they are constantly there, as booleans? Then

```js
updated(oldProps, oldState, changedProps) {
  if (changedProps.someProp) console.log('someProp changed!')
}
```

That's a good idea! Then compared to adding/removing string from a set the memory use doesn't fluctuate, and lookup is faster!

I'll update...

trusktr:
Alright, I've updated the PR. It seems like this would be cleaner.

The `connectedCallback` initially adds all the props to the _modifiedProps object (therefore updated logic can react to initial prop values).

Checking boolean values is about 5.5 times faster than checking for values in a Set according to this simple test in my Chrome:

```js
let o = {n:1}
let s = new Set([1,2,3,4])

var start = performance.now()
for (let i=0; i<5000000; i++) o.n
var end = performance.now()
console.log(end-start)

var start = performance.now()
for (let i=0; i<5000000; i++) s.has(4)
var end = performance.now()
console.log(end-start)
```

However, the `for..in` loop to clear reset the `_modifiedProps` object seems like it may be slower than clearing a Set.

trusktr:
I just did some perf tests.

- `for..in` takes 1/2 the time of `keys(o).forEach`.
  - iterating over 15 props with `for..in` 5,000,000 times took ~2300ms.
- `Object.keys(o).forEach` takes 3/5ths the time of `for..in`
  - iterating over 15 props with `Object.keys(o).forEach` 5,000,000 times took ~1550ms.
- `Object.keys(o)` paired with a plain `for (;;)` takes 1/31th the time of `Object.keys(o).forEach`
  - iterating over 15 props with the plain loop 5,000,000 times took ~50ms.

Updated the PR to use plain loops and to only ever call `Object.keys` one time.

packages/skatejs/src/with-update.js
```diff
@@ -133,11 +135,13 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
     static _attributeToPropertyMap: Object;
     static _observedAttributes: Array<string>;
     static _props: Object;
+    static _propsList: Array<string>;
```
trusktr: I made `_propsList` `static` instead of an instance variable so there's only one instance.

trusktr:
This implementation is fast now, and lightweight compared to diffing possibly much data across possibly many component instances.

It is currently branched from a branch of an unrelated feature, let me rebase it on master because the feature is standalone from my other pull requests.

trusktr:
Alright, I updated this PR to have `master` as base.

treshugart:
`@skatejs/element` already does this. As discussed this does it by using an object of old values. We've identified that this could, in theory, hinder 60fps performance but don't seem to have proven it yet. Since the discussion here is very long and hard to parse, we haven't proven a performance problem and it's already in a release (and it'd be a breaking change to change it to this), I'd prefer we:

1. Prove it's a problem.
2. Flesh it out in a separate issue keeping the discussion focused and terse.

PR #1513, opened 2018-08-20
Fix classList mutation methods

### LINKED ISSUE #1512: ClassList add and remove are broken
## Bug report

Add exibits the following behavior:
```js
const div = document.createElement('div')
div.classList.add('foo')
div.classList.contains('foo') // -> false
div.className // -> ""
```

Remove exibits the following behavior:
```js
const div = document.createElement('div')
div.className = 'foo'
div.classList.remove('foo')
div.classList.contains('foo') // -> true
div.className // -> "foo"
```

It seems that the className is never modified [here](https://github.com/skatejs/skatejs/blob/master/packages/ssr/register/ClassList.js#L3).

imaustink:
Skate SSR's implementation of classList does not make mutations to the className, this PR fixed the issues and adds tests.

* [x] Bug
* [ ] Feature

## Requirements

* [x] Wrote tests.

## Rationale
Fixes #1512

## Implementation
I didn't try anything else. I didn't feel the need to try and micro-optimize this implementation for obvious reasons.

## Open questions
What's going on with the deploy preview? It failed with `failed during stage 'building site': Deploy directory 'site/public' does not exist`. My changes should not have caused this.

## Other
All these questions make contributing painful.

### DIFF
--- a/packages/ssr/register/ClassList.js
+++ b/packages/ssr/register/ClassList.js
@@ -1,6 +1,8 @@
 // TODO performance test and update this from class accessors instead.
 function set(list) {
-  return new Set((list._elem.className || '').split(' '));
+  const classList =
+    (list._elem.className && list._elem.className.split(' ')) || [];
+  return new Set(classList);
 }
 
 class ClassList {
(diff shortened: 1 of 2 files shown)


trusktr:
LGTM

PR #1519, opened 2018-08-26
remove state? we already have all the functionality we need in props

trusktr:
* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

Do we _really_ need state?

`this.state` is

* just another prop, one that happens to be called `state`.
* it does not take advantage of the `props` API.
* redundant, we have everything we need using props.
* `state` is extra code that only duplicates a subset of the props functionality.

Contrasting to other libs:

* React uses `this.state` instead of props for reactivity
* Vue uses props instead of state for reactivity.
* Neither use both.

If someone really wants a `state` prop, they can easily add it using `props`:

```js
static props = {
  state: { /* ...  */ }
}

someMethod() {
  this.state = { ...this.state, foo: 'foo' }
}
```

The only thing missing is the `set` helper that merges a subset of props into the whole state so people can write `this.state = { foo: 'foo' }` instead of `this.state = { ...this.state, foo: 'foo' }`.

But, I don't think it's a big deal if we remove this. The premise is that users can define individual props, and set them directly, and don't really need to merge an object.

If we really want to provide this merging feature, it could be another mixin or utility function.

I think it is cleaner just to have `props`.

## Tasks

* [ ] tests
* [ ] docs

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -131,18 +131,16 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
     static _props: Object;
 
     _prevProps: Object;
-    _prevState: Object;
     _props: Object;
-    _state: Object;
     _syncingAttributeToProperty: null | string;
     _syncingPropertyToAttribute: boolean;
     _updating: boolean;
     _wasInitiallyRendered: boolean;
 
-    updated: ?(props: Object, state: Object) => void;
-    shouldUpdate: (props: Object, state: Object) => void;
+    updated: ?(props: Object) => void;
+    shouldUpdate: (props: Object) => void;
     triggerUpdate: () => void;
-    updating: ?(props: Object, state: Object) => void;
+    updating: ?(props: Object) => void;
 
     static _attributeToAttributeMap = {};
     static _attributeToPropertyMap = {};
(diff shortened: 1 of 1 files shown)


treshugart:
I think I really like this simplification of the API. Unfortunately, this is a breaking change, so it landing in master while v6 is in dev probably shouldn't happen. That said, I hope you don't mind that I've implemented in the v6-trey branch.

I should have this branch mostly stabilised in a couple of weeks and we can hash out the differences.

Cheers for the idea and PR!

trusktr:
👍 I wasn't aware of the `v6-trey` branch when I made this. But yeah, that's a good change! I also agree `this.props` is unnecessary, as we can just access `this.foo` instead of `this.props.foo`.

The only reason for it would be to do `this.props = {foo: 1, bar: 2}`, but we can easily just use `Object.assign(this, {foo: 1, bar: 2})` if we really want to.

Alternatively, it could be simplified to

```js
get props() { return this }
set props(props) {
  Object.assign(this, props)
}
```

if the shortcut is really useful. Is it useful?

treshugart:
> if the shortcut is really useful. Is it useful?

It's short enough that it's just an alias and given Skate is focused on minimalism, we should probably just leave it out.

PR #1526, opened 2018-09-03
Allow attr="false" for boolean attributes

trusktr:
* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

This makes it intuitive to do `some-attribute="false"` for boolean attributes.

## Implementation

Seems like the simplest way, but I haven't tested it yet.

### DIFF
--- a/packages/core/src/props.ts
+++ b/packages/core/src/props.ts
@@ -19,7 +19,7 @@ const array: PropType = {
 const boolean: PropType = {
   ...any,
   deserialize: (val): boolean => val != null,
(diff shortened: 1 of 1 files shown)


treshugart:
In react, `{false}` is a `boolean`, not a `string`.

>  it is less intuitive for attr="false" to be truthy

I remember being caught by this, but quickly corrected myself once I realised what I had done. I would prefer to be consistent than cater to ignorance.

PR #1528, opened 2018-09-03
swap object with array in props

trusktr:

* [ ] Bug
* [x] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

This is just a semantic change, as Array is a type of Object, but otherwise should behave the same. Not a big deal though.


treshugart:
Makes sense. Cheers!

Issue #1560, opened 2019-02-28
[bug] Every skate element gains property getters/setters of all other skate element.


trusktr:
## Bug report

_Delete this section if this is a feature request._

* Skate version: v5.2.4
* Affected browsers (and versions): all

### Current behaviour

After defining multiple classes with `withUpdate`, every instance of all classes will have the combination of all property getters/setters from every class in the application.

### Expected behaviour

Each element instance should only have the property getters/setters of the `props` defined by it's `class`, not those of any other class.

### more details

The reason this happens is because there's an `_observedAttributes` array on the `WithUpdate` class. This array is shared with all `class` constructor instances that inherit from `WithUpdate`.

The problem is similar to placing an array as a prototype property, and having all instances push/pop to/from it.

The solution is to create one array per constructor that inherits from `WithUpdate`. We can do that by placing the following line at the top of the `static get observedAttributes` getter:

```js
    static get observedAttributes(): Array<string> {
      if (!('_observedAttributes' in this)) this._observedAttributes = [] // THIS LINE
      defineProps(this, WithUpdate);
      return this._observedAttributes.concat(super.observedAttributes || []);
    }
```

It causes a new array to be created per class instance, instead of using the shared array defined by `static _observedAttributes = [];`.

trusktr:
Uh oh, I just noticed that this applies with `_attributeToAttributeMap` and `_attributeToPropertyMap` too! Each new subclass will clobber other subclass definitions if they have the same attribute names.

It doesn't look like these problems exist in the latest `master` any more.

trusktr:
I found the skate-5.2.5 commit and made a branch with the same name so that I could make a PR with the fixes: https://github.com/skatejs/skatejs/pull/1561

treshugart:
> It doesn't look like these problems exist in the latest master any more.

Can you confirm? If so can we close this?

PR #1561, opened 2019-03-03
make sure the static cache properties are created per-class

trusktr:
make sure the static cache properties are created per-class, instead of shared with all subclasses. Also remove unused properties

issue: https://github.com/skatejs/skatejs/issues/1560

* [x] Bug
* [ ] Feature

## Requirements

* [x] Wrote tests.

## Rationale

Without this change, any subclasses of a base class that uses` withUpdate` will share all `observedAttributes` of all other sub classes (and similar with _attributeToAttributeMap and _attributeToPropertyMap).

## Implementation

I placed the code in `get observedAttributes`, as that's the entry point, triggered by `customElements.define()`. The conditional check is required, because it is possible users and sub classes can also call `get observedAttributes`.

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -134,10 +134,7 @@ export const withUpdate = (Base: Class<any> = HTMLElement): Class<any> =>
     _prevState: Object;
     _props: Object;
     _state: Object;
-    _syncingAttributeToProperty: null | string;
-    _syncingPropertyToAttribute: boolean;
     _updating: boolean;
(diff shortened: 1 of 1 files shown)


treshugart:
Is this still an issue in skatejs/element?

trusktr:
Not a problem in the new `element`, just in v5.

treshugart:
Maybe let's focus on how you can use the new `element` package?

treshugart:
Closing as this isn't an issue in the new releases.

PR #1562, opened 2019-03-04
fix: it was impossible to remove `array` or `object` attributes when setting the JS prop to `null`.

trusktr:
* [x] Bug
* [ ] Feature

## Requirements

* [ ] Wrote tests.

## Rationale

It is impossible for `prop.serialize` to return null if it simply returns `JSON.stringify(val)`, because a string can not be coerced to `null`, therefore doing something like `this.someProp = null` does not remove the target attribute.

## Implementation

Checks if val is null, and if it is just returns null.

### DIFF
--- a/packages/skatejs/src/with-update.js
+++ b/packages/skatejs/src/with-update.js
@@ -276,7 +276,7 @@ const array: Function = prop({
     Array.isArray(val) ? val : empty(val) ? null : [val],
   default: Object.freeze([]),
   deserialize: parse,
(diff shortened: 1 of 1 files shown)


treshugart:
Is this still an issue in skatejs/element?

trusktr:
Seems like it is, here:

https://github.com/skatejs/skatejs/blob/359057e333e3cb6697156d736202ed69786bf21c/packages/element/src/props.ts#L18

PR #1575, opened 2020-02-21
Fix `@skatejs/ssr` in `jsdom` environment

NMinhNguyen:
`NodeList`s don't have a `map` method.
See https://developer.mozilla.org/en-US/docs/Web/API/NodeList

* [x] Bug
* [ ] Feature

## Requirements

* [x] Wrote tests.

## Rationale

_Why is this PR necessary?_

`NodeList`s don't have a `map` method.
See https://developer.mozilla.org/en-US/docs/Web/API/NodeList

I have existing tests using `undom` but I've had a lot of luck switching over to https://github.com/SimenB/jest-environment-jsdom-sixteen (see https://github.com/jsdom/jsdom/releases/tag/16.2.0).

## Implementation

I considered using Array spread or `Array.from` but went with this approach to avoid creating arrays. Happy to change the implementation for terseness though.

## Open questions

## Other

_Is there anything else we should know? Delete this section if you don't need it._

I've tested both this approach (as well as using Array spread) in my project via https://github.com/ds300/patch-package and everything works.

### DIFF
--- a/packages/ssr/index.js
+++ b/packages/ssr/index.js
@@ -17,8 +17,10 @@ function stringify(node) {
     return node.textContent;
   }
 
-  str += `<${node.localName}${(node.attributes || [])
-    .map(a => ` ${a.name}="${a.value}"`)
+  const { map } = Array.prototype;
+
+  str += `<${node.localName}${map
+    .call(node.attributes || [], a => ` ${a.name}="${a.value}"`)
     .join('')}>`;
 
   if (node.nodeName === 'BODY') {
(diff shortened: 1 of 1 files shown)


packages/ssr/index.js
```diff
@@ -17,22 +17,24 @@ function stringify(node) {
     return node.textContent;
   }
 
-  str += `<${node.localName}${(node.attributes || [])
-    .map(a => ` ${a.name}="${a.value}"`)
+  const { map } = Array.prototype;
```
treshugart, line 20: Should probably hoist this out of the function scope.

treshugart:
Left one small nit but otherwise lgtm.

treshugart:
Discussed in Twitter DM, but we'll just keep the `map` destructure inline for now.

